PR #17, opened 2019-06-04
Function to set DIO2 of SX126x chips to be used as RF antenna switch

beegee-tokyo:
Add function to set DIO2 of SX126x chips to be used as RF antenna switch.    
Based on example design of Semtech and used in many modules.    

This breaks the CAD function as DIO2 cannot be used as IRQ output.

Signed-off-by: Bernd Giesecke <bernd@giesecke.tk>

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -79,10 +79,6 @@ void Module::term() {
   _spi->end();
 }
 
-void Module::setDio2Func(bool enable) { 
-  _dio2RfSwitch = enable;
-}
-
 void Module::ATemptyBuffer() {
   while(ModuleSerial->available() > 0) {
     ModuleSerial->read();
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -79,6 +79,10 @@ void Module::term() {
   _spi->end();
 }
 
+void Module::setDio2Func(bool enable) { 
+  _dio2RfSwitch = enable;
+}
+
 void Module::ATemptyBuffer() {
   while(ModuleSerial->available() > 0) {
     ModuleSerial->read();
--- a/src/Module.h
+++ b/src/Module.h
@@ -297,6 +297,7 @@ class Module {
     int _rx;
     int _int0;
     int _int1;
+
     SPIClass* _spi;
     SPISettings _spiSettings;
 
--- a/src/Module.h
+++ b/src/Module.h
@@ -278,6 +278,20 @@ class Module {
     int getTx() const { return(_tx); }
 
     /*!
+      \brief Access method to get DIO2 RF switch flag.
+
+      \returns true if DIO2 is set as RF switch.
+    */
+    int getDio2Func() const { return(_dio2RfSwitch); }
+
+    /*!
+      \brief Access method to set/reset DIO2 RF switch flag.
+
+      \returns true if DIO2 is set as RF switch.
+    */
+    void setDio2Func(bool enable);
+
+    /*!
       \brief Access method to get the SPI interface.
 
       \returns SPI interface configured in the constructor.
@@ -297,7 +311,7 @@ class Module {
     int _rx;
     int _int0;
     int _int1;
-
+    bool _dio2RfSwitch = false;
     SPIClass* _spi;
     SPISettings _spiSettings;
 
--- a/src/Module.h
+++ b/src/Module.h
@@ -278,20 +278,6 @@ class Module {
     int getTx() const { return(_tx); }
 
     /*!
-      \brief Access method to get DIO2 RF switch flag.
-
-      \returns true if DIO2 is set as RF switch.
-    */
-    int getDio2Func() const { return(_dio2RfSwitch); }
-
-    /*!
-      \brief Access method to set/reset DIO2 RF switch flag.
-
-      \returns true if DIO2 is set as RF switch.
-    */
-    void setDio2Func(bool enable);
-
-    /*!
       \brief Access method to get the SPI interface.
 
       \returns SPI interface configured in the constructor.
@@ -311,7 +297,6 @@ class Module {
     int _rx;
     int _int0;
     int _int1;
-    bool _dio2RfSwitch = false;
     SPIClass* _spi;
     SPISettings _spiSettings;
 
--- a/src/modules/SX126x.cpp
+++ b/src/modules/SX126x.cpp
@@ -33,6 +33,11 @@ int16_t SX126x::begin(float bw, uint8_t sf, uint8_t cr, uint16_t syncWord, uint1
   }
 
   // configure publicly accessible settings
+  state = setDio2AsRfSwitch(false);
+  if(state != ERR_NONE) {
+    return(state);
+  }
+
   state = setSpreadingFactor(sf);
   if(state != ERR_NONE) {
     return(state);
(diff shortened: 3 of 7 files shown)


beegee-tokyo:
Tested on two modules
- eByte E22-900M which does not use DIO2 as RF antenna switch
- Insight ISP4520 which does use DIO2 as RF antenna switch

TODO
Get CAD function to work again (e.g. by using DIO3 if available).

src/Module.h
```diff
@@ -277,6 +277,20 @@ class Module {
...
+    */
+    int getDio2Func() const { return(_dio2RfSwitch); }
+
+    /*!
+      \brief Access method to set/reset DIO2 RF switch flag.
+
+      \returns true if DIO2 is set as RF switch.
+    */
+    void setDio2Func(bool enable);
+
```
jgromes: Module class should not contain module-specific code.

src/Module.cpp
```diff
@@ -79,6 +79,10 @@ void Module::term() {
   _spi->end();
 }
 
+void Module::setDio2Func(bool enable) { 
+  _dio2RfSwitch = enable;
+}
+
```
jgromes: Module class should not contain module-specific code.

src/Module.h
```diff
@@ -297,7 +311,7 @@ class Module {
     int _rx;
     int _int0;
     int _int1;
-
+    bool _dio2RfSwitch = false;
```
jgromes: Module class should not contain module-specific code.

src/modules/SX126x.cpp
```diff
@@ -1058,6 +1063,29 @@ int16_t SX126x::setFrequencyRaw(float freq) {
...
 
+int16_t SX126x::setDio2AsRfSwitch(bool enable) {
+  int dio2 = _mod->getInt1();
+  if (dio2 == -1)
+  {
+    // DIO2 is not defined, return error
+    _mod->setDio2Func(false);
+    return ERR_WRONG_MODEM;
+  }
+  uint8_t* data = new uint8_t[1];
```
jgromes: Memory leak. There's no reason to allocate dynamically, use 
```suggestion
  uint8_t data[1];
```

src/modules/SX126x.h
```diff
@@ -678,6 +678,13 @@ class SX126x: public PhysicalLayer {
     */
     float getSNR();
 
+    /*!
+      \brief Set DIO2 to function as RF switch (default in Semtech example designs).
+
+      \returns \ref status_codes 
+    */
+    int16_t setDio2AsRfSwitch(bool enable = false);
```
jgromes: Change default enable value to true. If user is calling this method, it is probably because he wants to enable the RF control.

jgromes:
Also, I don't think it't a good idea to use DIO3 for CAD timeout. DIO3 must be used on some modules for TCXO control.

If RF control on DIO2 is enabled, CAD should be disabled entirely.

src/Module.h
beegee-tokyo: So it should be in SX126x.h and SX126x.cpp ?

jgromes: Do we need these methods at all? We can keep the DIO2 state as a variable in SX126x class.

src/modules/SX126x.h
beegee-tokyo: Right.

beegee-tokyo:
Testing the changes right now. Function moved to SX126x.cpp, memory leak removed, simplified, removed unnecessary functions.

Question before I push the changes.

Do you want a new error value for scanChannel() if DIO2 is used for RF antenna switch?

jgromes:
Since that's quite a unique error, I think a new status code is a good idea. It should have the value -708.

beegee-tokyo:
```cpp
/*!
  \brief SX126x scan channel not possible because DIO2 is used as RF antenna switch.
*/
#define ERR_DIO2_UNAVAIL_CAD_FAILED           -708
```
and
```cpp
int16_t SX126x::scanChannel() {
  // check active modem
  if(getPacketType() != SX126X_PACKET_TYPE_LORA) {
    return(ERR_WRONG_MODEM);
  }

  if (_dio2RfSwitch) {
    // If DIO2 is used as RF switch this function does not work
    return(ERR_DIO2_UNAVAIL_CAD_FAILED);
  }
...
```

beegee-tokyo:
Should be ok now.

Another thing, I get a compile error on ESP8266 module using PlatformIO. I am not compiling for ESP8266 or including ESP8266.h from the lib (but it is included in RadioLib.h)!

```log
Compiling .pioenvs\LoraSender_4520\libeb3\RadioLib-DIO2\RadioLib.cpp.o
Compiling .pioenvs\LoraSender_4520\libeb3\RadioLib-DIO2\modules\CC1101.cpp.o
Compiling .pioenvs\LoraSender_4520\libeb3\RadioLib-DIO2\modules\ESP8266.cpp.o
Compiling .pioenvs\LoraSender_4520\libeb3\RadioLib-DIO2\modules\HC05.cpp.o
lib\RadioLib-DIO2\src\modules\ESP8266.cpp:183:10: error: prototype for 'uint16_t ESP8266::getNumBytes(uint32_t, size_t)' does not match any in class 'ESP8266'
uint16_t ESP8266::getNumBytes(uint32_t timeout, size_t minBytes) {
^~~~~~~
In file included from lib\RadioLib-DIO2\src\modules\ESP8266.cpp:1:0:
lib\RadioLib-DIO2\src\modules\ESP8266.h:55:12: error: candidate is: virtual size_t ESP8266::getNumBytes(uint32_t, size_t)
size_t getNumBytes(uint32_t timeout = 10000, size_t minBytes = 10);
^~~~~~~~~~~
*** [.pioenvs\LoraSender_4520\libeb3\RadioLib-DIO2\modules\ESP8266.cpp.o] Error 1
```
The cause is that in ESP8266.h the function is declared as
```cpp
    size_t getNumBytes(uint32_t timeout = 10000, size_t minBytes = 10);
```
but in ESP8266.cpp the function is set as
```cpp
uint16_t ESP8266::getNumBytes(uint32_t timeout, size_t minBytes) {
```
Will include it in this commit.

src/modules/SX126x.cpp
```diff
@@ -1058,8 +1063,25 @@ int16_t SX126x::setFrequencyRaw(float freq) {
...
+  int16_t state = SPIwriteCommand(SX126X_CMD_SET_DIO2_AS_RF_SWITCH_CTRL, data, 1);
+
+  if (state == ERR_NONE) {
+    _dio2RfSwitch = true;
+  }
+  return(state);
+}
+
 int16_t SX126x::config(uint8_t modem) {
   // set DIO2 as IRQ
```
jgromes: It would be better to remove this entire block. The purpose of config() metod is to configure stuff that is inaccessible via the public API. DIO2 setting is accessible now, so we should have a call `setDio2AsRfSwitch(false);` in the `begin()` and `beginFSK()` mtehods.

src/modules/SX126x.cpp
```diff
@@ -1058,8 +1063,25 @@ int16_t SX126x::setFrequencyRaw(float freq) {
...
+  if (enable) {
+  // set DIO2 as RF switch
+    data[0] = SX126X_DIO2_AS_RF_SWITCH;
+  } else {
+    data[0] = SX126X_DIO2_AS_IRQ;
+  }
+  int16_t state = SPIwriteCommand(SX126X_CMD_SET_DIO2_AS_RF_SWITCH_CTRL, data, 1);
+
+  if (state == ERR_NONE) {
+    _dio2RfSwitch = true;
```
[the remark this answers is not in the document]
beegee-tokyo: Good point. 
Of course.

jgromes:
All looks good now. Thanks!

PR #38, opened 2019-09-07
ESP32 - use HardwareSerial instead of SoftwareSerial

beegee-tokyo:
ESP32 has no working SoftwareSerial. With a simple #ifdef ESP32 a hardware serial will be used instead.    
Minimum changes without influence on existing installations or usage.


beegee-tokyo:
I know you are not targeting the ESP32. But your library is the best LoRa library I found. Would be a pitty if the ESP32 would be left out.
Unfortunately ESP32 has no SoftwareSerial support. The existing EspSoftwareSerial does not compile for me and is on top outdated.

Pablo2048:
The same thing is valid for SAMD platform (Arduino Zero), so +1 for this PR and use it like #if defined(ESP32) || defined(SAMD_SERIES) ...

beegee-tokyo:
@Pablo2048 
I cannot test on SAMD, I have no board with a SAMD, only one a Sparkfun LoRa board with an SAMD 21, is that the same?
Because that compiles without problems with this library.

Anyway, I somehow messed up this pull request, will retry.

PR #39, opened 2019-09-07
ESP32 & SAMD - use HardwareSerial instead of SoftwareSerial

beegee-tokyo:
ESP32 has no working SoftwareSerial. With a simple #ifdef ESP32 || SAMD a hardware serial will be used instead.    
Minimum changes without influence on existing installations or usage.

Hope this time I did a better job.

### DIFF
--- a/src/ISerial.cpp
+++ b/src/ISerial.cpp
@@ -9,7 +9,11 @@ void ISerial::begin(long speed) {
 }
 
 bool ISerial::listen() {
+#if defined ( ESP32 )  || defined (SAMD_SERIES)
+  return true;
+#else
   return(_mod->ModuleSerial->listen());
+#endif
 }
 
 void ISerial::end() {
@@ -17,15 +21,27 @@ void ISerial::end() {
 }
 
 bool ISerial::isListening() {
+#if defined( ESP32 ) || defined ( SAMD_SERIES )
+  return true;
+#else
   return(_mod->ModuleSerial->isListening());
+#endif
 }
 
 bool ISerial::stopListening() {
+#if defined( ESP32 ) || defined ( SAMD_SERIES )
+  return true;
+#else
   return(_mod->ModuleSerial->stopListening());
+#endif
 }
 
 bool ISerial::overflow() {
+#if defined( ESP32 ) || defined ( SAMD_SERIES )
+  return false;
+#else
   return(_mod->ModuleSerial->overflow());
+#endif
 }
 
 int ISerial::peek() {
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -1,13 +1,17 @@
 #include "Module.h"
 
-Module::Module(int rx, int tx) {
+Module::Module(int rx, int tx, HardwareSerial* useSer) {
   _cs = -1;
   _rx = rx;
   _tx = tx;
   _int0 = -1;
   _int1 = -1;
 
+#if defined(ESP32) || defined(SAMD_SERIES)
+    ModuleSerial = useSer;
+#else
   ModuleSerial = new SoftwareSerial(_rx, _tx);
+#endif
 }
 
 Module::Module(int cs, int int0, int int1, SPIClass& spi, SPISettings spiSettings) {
@@ -20,7 +24,7 @@ Module::Module(int cs, int int0, int int1, SPIClass& spi, SPISettings spiSetting
   _spiSettings = spiSettings;
 }
 
-Module::Module(int cs, int int0, int int1, int rx, int tx, SPIClass& spi, SPISettings spiSettings) {
+Module::Module(int cs, int int0, int int1, int rx, int tx, SPIClass& spi, SPISettings spiSettings, HardwareSerial* useSer) {
   _cs = cs;
   _rx = rx;
   _tx = tx;
@@ -29,7 +33,11 @@ Module::Module(int cs, int int0, int int1, int rx, int tx, SPIClass& spi, SPISet
   _spi = &spi;
   _spiSettings = spiSettings;
 
+#if defined(ESP32) || defined(SAMD_SERIES)
+  ModuleSerial = useSer;
+#else
   ModuleSerial = new SoftwareSerial(_rx, _tx);
+#endif
 }
 
 Module::Module(int cs, int int0, int int1, int int2, SPIClass& spi, SPISettings spiSettings) {
@@ -51,7 +59,11 @@ void Module::init(uint8_t interface, uint8_t gpio) {
       _spi->begin();
       break;
     case USE_UART:
+#if defined(ESP32)
+      ModuleSerial->begin(baudrate, SERIAL_8N1, _rx, _tx);
+#else
       ModuleSerial->begin(baudrate);
+#endif
       break;
     case USE_I2C:
       break;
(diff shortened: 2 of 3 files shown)


jgromes:
I'm not opposed to the idea, but we need to think about how to integrate it into the interface. You can't use addition argument to specify serial number, `Module(int, int, int)` conflicts with default SPI constructor.

Pablo2048:
Yes SAMD are SAMD21 based boards. Right now I've found SoftwareSerial library for Zero (I'm on Platformio and have to install it manually). Maybe the better idea is to use the same approach as with SPI? I mean use pointer to the HardwareSerial object instead of giving some number. What do You think?

beegee-tokyo:
Oh, I didn't check that because I have no module that is using serial. 
I like the idea of @Pablo2048 with a new constructor with the pointer to the Serial to be used. Would not disturb old installations but helps a lot with the ESP32.

Its getting late here, will try tomorrow.

jgromes:
@beegee-tokyo that's why we have CI ;) see https://travis-ci.org/jgromes/RadioLib/jobs/582051150

I agree with @Pablo2048's idea, we can just let the user create the HardwareSerial object, then just pass the pointer.

Just a thought - would this be compatible with AVR Arduino's `HardwareSerial`?

Pablo2048:
Well, I'm using this in SAMD sketch ```Stream *stream = &Serial2;``` so if we don't need to setup the speed, we can use Stream class which is same in whole Arduino platform...

beegee-tokyo:
Last comment before I fall asleep (it's past midnight here).
@Pablo2048 could you create a working solution based on your proposal? 

To be honest, I am selfish and I just want to get rid of the SoftwareSerial requirement for ESP32. Because Arduino-ESP32 framework doesn't have SoftwareSerial and the maintainers are very reluctant to add it because the ESP32 has 3 hardware serials and they don't see the need to add it (I agree with them).
Now of course I could just use a patched version of this fine and well maintained library in my projects, but I started to write learning [blogs](http://learn.circuit.rocks) and [tutorials](https://desire.giesecke.tk/) for local Filipino students and I cannot ask them to patch a library to get my lessons to work.

beegee-tokyo:
`Stream *stream = &Serial2;` can work, however, the user then has to do Serial2.begin() before he calls the module .init() function, because Stream has no begin.
That could break backwards compatibility.

I tried a similar approach with changed constructors for the modules that use Serial connection. But that would break compatibility for the users that implemented this library on SAMD before with a SoftwareSerial. But at least AVR users would have no problems. And for ESP32 it could still work in most cases, because you can assign RX and TX pins to nearly any GPIO, so it would just replace the SoftwareSerial with a HardwareSerial Serial1 on that pins.

I could not find a way to implement it without breaking some backward compatibility.

Lets see what Travis says :grin:

Btw. I am interested in this automatic checking with travis, but never found time to learn about it. I saw the .travis.yml here and it would match with my libraries as well, if I would know how to add the ESP32 environment to it. @jgromes could you help me how to add ESP32 to the tests?

jgromes:
Try passing `HardwareSerial` by pointer, that way it can be set to `NULL` or `nullptr` on AVR Arduino. on ESP32, you will have to call `Module(rx, tx, &Serial1);`

jgromes:
Also, could you please build these locally before pushing the changes? Just build the same example for ESP32 and Arduino Uno, then you don't have to wait for Travis to build everything, which might speed things up ;)

beegee-tokyo:
Ok, will try this.

beegee-tokyo:
Ok, it builds now locally for Uno, ATMega 2540, Leonardo, ESP8266 and ESP32.

However when I try to preset the `HardwareSerial& = nullptr` it throws a compiler error.
I left it on `HardwareSerial& useSer = Serial` which doesn't hurt because it will be ignored for anything than ESP32 and SAMD

Lets see what Travis complains now :sleeping:

jgromes:
`HardwareSerial&` is not a pointer, it's a reference to `HardwareSerial` object - you can't set reference to null. You should use `HardwareSerial*` instead.

beegee-tokyo:
Ok, Changed it. Example compiles locally without problems for  Uno, ATMega 2540, Leonardo, ESP8266, SAMD21 and ESP32.
Maybe a test with a real device should be made, but I have no such hardware.

jgromes:
Thanks, I will test it out later today (or tomorrow).

beegee-tokyo:
Here it is running on two ESP32, one connected to a SX1268, the other one to a SX127x.

Which of the modules you support have serial connection? I have some modules laying around, maybe I can test the serial connection on an ESP32.

Pablo2048:
Sorry guys for late reply - I do have SAMD21 board with some EBYTE modules so I give it a try also... Thank you anyway...

beegee-tokyo:
Got it to work with an ESP32 connected to a HC-05. HW serial works fine on the ESP32.
However, one coding error
```ModuleSerial->begin(baudrate, _rx, _tx);```
must be changed to 
```ModuleSerial->begin(baudrate, SERIAL_8N1, _rx, _tx);```
Will extend the pull request with this change.
xHC05

jgromes:
I can confirm the serial connection is working on ESP32 + ESP8266 (though there are some issues in my code related to dynamic memory allocation - I will investigate this further).

Thanks for this PR, I will merge and add ESP32 section to the Travis build - release to follow shortly.

mmrein:
I guess this should also work for STM32 based boards. I can try with STM32F103 aka "BluePill" board (as soon as I learn how to manage GitHub repos with arduino libs and stuff like that).

PS.: Found this lib on friday and was stuck with SoftwareSerial. Luckily got back on monday :) And BTW you made me to sign up for GitHub, good job guys.

jgromes:
Thanks, please let me know if it works (please open an issue if it doesn't). Out of curiosity, which module are you planning to use?

There's a few ways to get the library, I recommend one of these two:
1. Download through Arduino IDE library manager - it will receive updates whenever a new release is created. I'm about to release a new version with ESP32 support from this PR, it will take the library manager about an hour to pick up that update.
2. Clone the repository into Arduino libraries folder. Bit more complicated, but it lets you contribute and get updates more often.

beegee-tokyo:
@mmrein 
I don't think it will work for the STM32 right out of the box. There will be more #if defined be needed to suppress the use of SoftwareSerial. I did this only for ESP32 and SAMD, the STM32 would be treated as any other AVR board.

But it should be not too difficult to add the STM32 to it. I just cannot do any changes.

If you look into the pull-requests changed files you can see where I added processor specific changes, there you will have to add the STM32.

mmrein:
@jgromes 
I have RFM95 and SX1261 based modules, just wanted to try some FSK modulation with Semtech Radios.

@beegee-tokyo 
Yes, I saw that and that's exactly what I wanted to try. Give me few days to get familiar with RadioLib and other things, hopefully it would work.

Thanks for the tips, guys.

mmrein:
OK, I've cloned the repo, added "|| defined (ARDUINO_ARCH_STM32)" after "defined(SAMD_SERIES)" and built successfully. 

Now I'm having problems with bootloader / drivers, so I can't upload now. Will check on that tomorrow.

PR #44, opened 2019-09-16
Add setSyncBits for SX126x

mmrein:
Add setSyncBits for possibility to set the sync word length in bits

### DIFF
--- a/src/modules/SX126x.cpp
+++ b/src/modules/SX126x.cpp
@@ -733,7 +733,9 @@ int16_t SX126x::setSyncBits(uint8_t *syncWord, uint8_t bitsLen) {
   }
 
   uint8_t bytesLen = bitsLen / 8;
-  if ((bitsLen % 8) != 0) bytesLen++;
+  if ((bitsLen % 8) != 0) {
+    bytesLen++;
+  }
   
   // write sync word
   int16_t state = writeRegister(SX126X_REG_SYNC_WORD_0, syncWord, bytesLen);
--- a/src/modules/SX126x.cpp
+++ b/src/modules/SX126x.cpp
@@ -721,6 +721,33 @@ int16_t SX126x::setSyncWord(uint8_t* syncWord, uint8_t len) {
   return(state);
 }
 
+int16_t SX126x::setSyncBits(uint8_t *syncWord, uint8_t bitsLen) { 
+  // check active modem
+  if(getPacketType() != SX126X_PACKET_TYPE_GFSK) {
+    return(ERR_WRONG_MODEM);
+  }
+
+  // check sync word Length
+  if(bitsLen > 0x40) {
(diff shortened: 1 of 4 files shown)


src/modules/SX126x.cpp
```diff
@@ -721,6 +721,33 @@ int16_t SX126x::setSyncWord(uint8_t* syncWord, uint8_t len) {
...
+    return(ERR_WRONG_MODEM);
+  }
+
+  // check sync word Length
+  if(bitsLen > 0x40) {
+    return(ERR_INVALID_SYNC_WORD);
+  }
+
+  uint8_t bytesLen = bitsLen / 8;
+  if ((bitsLen % 8) != 0) bytesLen++;
```
jgromes: Could you add the braces. as explained in [code style guideline](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md#code-style-guidelines)?

src/modules/SX126x.h
```diff
@@ -604,6 +604,17 @@ class SX126x: public PhysicalLayer {
     */
     int16_t setSyncWord(uint8_t* syncWord, uint8_t len);
 
+    /*!
+      \brief Sets FSK sync word in the form of array of up to 8 bytes.
```
jgromes: We should probably stress that setting a sync word here will override whatever was set by `setSyncWord`. Also, a note about ignoring LSB would be nice.

src/modules/SX126x.h
```diff
@@ -604,6 +604,17 @@ class SX126x: public PhysicalLayer {
...
+
+      \param syncWord FSK sync word to be set.
+
+      \param len FSK sync word length in bits.
+
+      \returns \ref status_codes
+    */
+    int16_t setSyncBits(uint8_t *sync, uint8_t bitsLen);
```
jgromes: The name of the method should also be added to keywords.txt (please use true tab there), and it would be nice to have it in the SX126x_Settings example.

src/modules/SX126x.cpp
mmrein: Sorry, I forgot about this one. Will repair ASAP.

src/modules/SX126x.h
mmrein: I'm not sure if I do understand well, but overriding whatever was set by `setSyncWord `would be expected IMO. Should be mentined though. 

And yes, I was thinking about the way to describe the bits to be used, but I didn't find the right words so far. I will try to formulate that.

src/modules/SX126x.h
mmrein: Ok, I will try to add that. Now I'm just not sure how to "update" Pull request (GitHub newbie)..

jgromes:
Merged, thanks!

Issue #54, opened 2019-11-05
The length byte is optional and should be written/read only if the module is in variable-length mode


phretor:
**Is your feature request related to a problem? Please describe.**
Pure enhancement.

**Describe the solution you'd like**
As for the `filter` byte, the `length` byte should be optional, as stated in the specs: it's written only if the application protocol is in variable-length mode. Also, is there any reason why the `filter` reg value is not cached on the MCU memory? It would save one SPI read, and in high throughput transmissions it may make a difference.

I'm talking about CC1101 or RFM69, and similar modules.

jgromes:
Variable length mode is the only supported mode for all FSK/LoRa radio modules in RadioLib, because it provides the easiest integration into the high-level API - that's why length byte is always written/read. Therefore, changes in #55 would only make sense if modes other than variable length were available to the user.

I don't plan on adding fixed or unlimited length mode in the near future, but I will accept PRs for that purpose - provided that everything integrates nicely into the high-level API. There was an effort to add support for unlimited packets for SX127x in LoRaLib (https://github.com/jgromes/LoRaLib/pull/66), but ultimately it went nowhere.

> is there any reason why the filter reg value is not cached on the MCU memory?

Most register values aren't cached - it's the old size vs. speed dilema.

phretor:
Thanks for considering this.

I wish I had found about RadioLib before, at least before I decided that it was time to for RadioHead to remove the packet-structure assumptions (i.e., header with to/from addresses) they have in there. For this reason I started porting RadioHead code into my [RadioHAL](https://github.com/trendmicro/RadioHAL). I needed a clean, think HAL for [RFQuack](https://github.com/trendmicro/RFQuack). At this point, looking at the radio abstraction alone, the only difference in the approachs of RadioHAL and RadioLib, is the user access to variable/fixed length mode. But of course RadioLib has way many more radios supported. So, I'm considering using RadioLib instead of RadioHAL, but I'm still in the phase where I want to get used to it.

> I don't plan on adding fixed or unlimited length mode in the near future, but I will accept PRs for that purpose

Do the changes in #55 qualify as "acceptable" for your standards?

jgromes:
I will have to take a look at how will fixed length packets affect PhysicalLayer.

There's a simple rule when adding new features - if there's a real world use for it, it will likely be added. What's the use case for fixed length packets?

phretor:
> What's the use case for fixed length packets?

[...]

PR #55, opened 2019-11-06
Using defined debug port and not Serial + CC1101 length mode config options

phretor:


phretor:
Proposal on how to address #54 in the CC1101 module. I can implement and send more pull request for other modules if this approach is acceptable.

src/modules/CC1101.cpp
```diff
@@ -477,7 +480,11 @@ uint8_t CC1101::getLQI() {
 
 size_t CC1101::getPacketLength(bool update) {
   if(!_packetLengthQueried && update) {
-    _packetLength = _mod->SPIreadRegister(CC1101_REG_FIFO);
+
+    uint8_t format = SPIgetRegValue(CC1101_REG_PKTCTRL0, 1, 0);
```
jgromes: Should be able to use `_packetLengthConfig` variable here instead of SPI read.

src/modules/CC1101.cpp
```diff
@@ -497,6 +504,38 @@ int16_t CC1101::config() {
   return(state);
 }
 
+int16_t CC1101::fixedPacketLengthMode(uint8_t len) {
```
jgromes: Neither one of the new methods updates `_packetLengthConfig`.

phretor:
Thanks for the thorough review. I'll handle the comments. Should I just make a new pull request or merge new changes with this one?

jgromes:
New PR for the new changes in other radio modules seems more appropriate, thanks.

phretor:
Changeset 112110d should address your requests.

src/modules/CC1101.cpp
```diff
@@ -480,11 +480,12 @@ uint8_t CC1101::getLQI() {
 
 size_t CC1101::getPacketLength(bool update) {
   if(!_packetLengthQueried && update) {
-
-    uint8_t format = SPIgetRegValue(CC1101_REG_PKTCTRL0, 1, 0);
-    if (format == CC1101_LENGTH_CONFIG_VARIABLE) {
+    if (_packetLengthConfig == CC1101_LENGTH_CONFIG_VARIABLE) {
+      _packetLength = _mod->SPIreadRegister(CC1101_REG_FIFO);
+    } else {
       _packetLength = _mod->SPIreadRegister(CC1101_REG_FIFO);
```
jgromes: I think we should read the length from CC1101_REG_PKTLEN for fixed packets, right?

phretor: Yeah, sorry. The upcoming one should do. I've started integrating the same changes in other modules, too. Will push in the following hours. For some of them I have hardware that I can test.

phretor:
Commit [a2e64f1](https://github.com/jgromes/RadioLib/pull/55/commits/a2e64f1645a9f6630e7ef2b9ec9abc7e6c335349) should do it.

jgromes:
Looks godd to me - merged, thanks!

PR #56, opened 2019-11-08
Variable/fixed packet length mode config for RF69/SX1231

phretor:
Related idea: is it possible to just call `RF69::begin` within `SX1X1231::begin` and just add the `if (_chipRevision == SX1231_CHIP_REVISION_2_A)` block before returning?

### DIFF
--- a/src/modules/RF69.cpp
+++ b/src/modules/RF69.cpp
@@ -4,6 +4,7 @@ RF69::RF69(Module* module) : PhysicalLayer(RF69_CRYSTAL_FREQ, RF69_DIV_EXPONENT,
   _mod = module;
   _tempOffset = 0;
   _packetLengthQueried = false;
+  _packetLengthConfig = RF69_PACKET_FORMAT_VARIABLE;
 }
 
 int16_t RF69::begin(float freq, float br, float rxBw, float freqDev, int8_t power) {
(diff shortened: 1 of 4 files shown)


src/modules/RF69.cpp
```diff
@@ -250,6 +261,11 @@ int16_t RF69::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
   // set packet length
   _mod->SPIwriteRegister(RF69_REG_FIFO, len);
 
+  // optionally write packet length
+  if (_packetLengthConfig == RF69_PACKET_FORMAT_VARIABLE) {
+    SPIwriteRegister(RF69_REG_FIFO, len);
```
jgromes: Call `_mod->SPIwriteRegister()`

phretor: OK. BTW, why some modules have `_mod` and some don't?

phretor:
Changeset [39a4614](https://github.com/jgromes/RadioLib/pull/56/commits/39a4614a775f0f81fd00cafb4af4b50c8eeaf314) should address the latest request.

jgromes:
> is it possible to just call `RF69::begin` within `SX1X1231::begin`

Unforutnately it's not that easy - SX1231 has different values in the version register. `RF69::begin` would have to be generalized, a bit like  `SX127x::begin`

jgromes:
Looks good, thanks!

PR #61, opened 2019-11-11
Implemented TX PA Clamping, datasheet section 15.2

BarryPSmith:
Small change. Pored through the datasheet while trying to debug my TX power issues, this wasn't the cause but it's recommended by the manufacturer so why not?


jgromes:
Same as #57 - could you please take a look if this is the case for SX1268 as well?

BarryPSmith:
Yep, it was there in the 1268 datasheet. This code is now called for both.

src/modules/SX126x.h
```diff
@@ -751,6 +752,10 @@ class SX126x: public PhysicalLayer {
 
     int16_t setFrequencyRaw(float freq);
 
+    /*!
+      \brief Fixes overly eager PA clamping on SX1262 / SX1268, as described in section 15.2 of the datasheet
+    */
+    int16_t fixPaClamping();
```
jgromes: I don't think this method needs to be public - according to the datasheet, the register should be only set during initialization, right?

jgromes:
You were right, there seems to be a minor conflict (inconsistent brackets in return).

BarryPSmith:
Ah I somehow didn't notice that all the existing return statements use brackets. I've updated this one to not use brackets so it can be merged, future work will use the existing convention of return(...)

jgromes:
Ah, sorry, I didn't realize `SX126x::fixPaClamping` was called from derived classes - move it to `protected` in that case.

BarryPSmith:
Yeah I should have picked that up too. I clearly need more tea. Strange, I usually check it compiles before committing. Probably distracted by the conflicts and the squashing and the force commits.

jgromes:
No worries, that's why the Travis script is there ;)

Merged, once again, thank you very much for this - when I was initially writing the SX126x driver, the datasheet left a lot to be desired, mainly in the PA/power configuration section. To be honest, there's still way more undocumented registers and settings than I would have liked, but at least they added some explanation in places.

jgromes:
@BarryPSmith hey, I just noticed the datasheet says the PA clamping should be fixed "after a Power On Reset, or a wake-up from cold Start" - I think we should move the calls to `fixPaClamping()` to `SX126x::reset()`. Though the datasheet doesn't specify at which stage of initialization should this happen (i.e. before or after sending SPI commands).

BarryPSmith:
Hmmm.

More of a library use detail than hardware, I think: After reset, the module should be back to its default configuration. Which means that any setup functions that are not publicly accessible must be called by the library after a reset.

Maybe `reset()` should just call `begin(...)`? It would require ensuring that all necessary parameters are saved in the MCU memory. Perhaps they already are? You could spend a bunch of effort figuring out exactly what users expect, or just do that and document it.

Right now I think the user has to call `begin(...)` after `reset()`. And `begin(...)` performs a reset, so as it stands I can't really imagine in what situation the user would actually want to call `reset` themselves. As I say, the only real use I can think of is as a wrapper for `begin`, but then the word `reset` doesn't seem entirely appropriate - perhaps `reinitialise` would be better if you decide to go that route?

---------
As for when to call `fixPaClamping` - I expect that it doesn't matter exactly when its called, just ensure that the unit is able to receive commands (i.e. finished its reset sequence) and that it's called before the unit tries to transmit.

jgromes:
You're right, there's probably no use case for calling `reset()` on its own, since the radio will lose all configuration - so just calling `begin()` again in the case of some weird error should be enough.

PR #62, opened 2019-11-12
Added fixed-length example for CC1101

phretor:


jgromes:
I'm not sure we need this - just amending CC1101_Settings with note about fixed and variable packets is probably enough. 

The examples are supposed to demonstrate the basic library API. I would argue that fixed-length packets are advanced feature, so it's probably not necessary to include additional example.

PR #63, opened 2019-11-12
Implemented promiscious mode for CC1101 and RF69

phretor:
Promiscuous mode is needed in sniffing scenarios, similarly to variable vs. fixed packet mode. I also made `setSyncWord` uniform, retaining the original signature for compatibility.

I'm not aware of a way to disable sync word detection alone in CC1101. The only way I know also disables preamble detection altogether. Therefore, `disableSyncWordFiltering` also disables preamble filtering.

If you know a better way to address this, I'll commit some changes.


src/modules/CC1101.cpp
```diff
@@ -441,11 +443,35 @@ int16_t CC1101::setOutputPower(int8_t power) {
   return(SPIsetRegValue(CC1101_REG_PATABLE, powerRaw));
 }
 
-int16_t CC1101::setSyncWord(uint8_t syncH, uint8_t syncL) {
-  // set sync word
-  int16_t state = SPIsetRegValue(CC1101_REG_SYNC1, syncH);
-  state |= SPIsetRegValue(CC1101_REG_SYNC0, syncL);
-  return(state);
+int16_t CC1101::setSyncWord(uint8_t* syncWord, uint8_t len, uint8_t maxErrBits) {
+  if(maxErrBits > 1) {
```
jgromes: Add sync word length check

src/modules/CC1101.cpp
```diff
@@ -547,6 +573,76 @@ int16_t CC1101::variablePacketLengthMode(uint8_t maxLen) {
...
+    if (_syncWordLength == 1) {
+      // in 16 bit sync word, expect at least 15 bits
+      return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_15_16, 2, 0));
+    } else if (_syncWordLength == 2) {
+      // in 32 bits sync word (16 + 16), expect 30 of 32 to match
+      return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_30_32, 2, 0));
+    }
+  }
+
+  return(ERR_NONE);
```
jgromes: Change to ERR_UNKNOWN, since there should be no way to reach this return statement.

src/modules/CC1101.cpp
```diff
@@ -547,6 +573,76 @@ int16_t CC1101::variablePacketLengthMode(uint8_t maxLen) {
...
+int16_t CC1101::enableSyncWordFiltering(uint8_t maxErrBits) {
+  if (maxErrBits > 1) {
+    return(ERR_INVALID_SYNC_WORD);
+  }
+
+  if (maxErrBits == 0) {
+    if (_syncWordLength == 1) {
+      // in 16 bit sync word, expect all 16 bits
+      return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_16_16, 2, 0));
+    } else if (_syncWordLength == 2) {
```
jgromes: Use just else (to let compiler and/or static code analysis tools let know these are the only options).

src/modules/CC1101.cpp
```diff
@@ -547,6 +573,76 @@ int16_t CC1101::variablePacketLengthMode(uint8_t maxLen) {
...
+      // there's no 32 of 32 case, so we resort to 30 of 32 bits required
+      return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_30_32, 2, 0));
+    }
+  }
+
+  if (maxErrBits == 1) {
+    if (_syncWordLength == 1) {
+      // in 16 bit sync word, expect at least 15 bits
+      return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_15_16, 2, 0));
+    } else if (_syncWordLength == 2) {
```
jgromes: Same as previous comment

src/modules/CC1101.cpp
```diff
@@ -547,6 +573,76 @@ int16_t CC1101::variablePacketLengthMode(uint8_t maxLen) {
...
+  }
+
+  return(ERR_NONE);
+}
+
+int16_t CC1101::disableSyncWordFiltering() {
+  return(SPIsetRegValue(CC1101_REG_MDMCFG2, CC1101_SYNC_MODE_NONE, 2, 0));
+}
+
+int16_t CC1101::enableCrcFiltering() {
```
jgromes: Use just one method -  `setCRCFiltering(bool)`

src/modules/CC1101.cpp
```diff
@@ -547,6 +573,76 @@ int16_t CC1101::variablePacketLengthMode(uint8_t maxLen) {
...
+  return(SPIsetRegValue(CC1101_REG_PKTCTRL0, CC1101_CRC_ON, 2, 2));
+}
+
+int16_t CC1101::disableCrcFiltering() {
+  return(SPIsetRegValue(CC1101_REG_PKTCTRL0, CC1101_CRC_OFF, 2, 2));
+}
+
+int16_t CC1101::promiscuousMode(bool promiscuous) {
```
jgromes: Change name to `setPromiscuousMode(bool)` - fits the API style better.

src/modules/RF69.cpp
```diff
@@ -87,8 +92,8 @@ int16_t RF69::begin(float freq, float br, float rxBw, float freqDev, int8_t powe
   }
 
   // default sync word values 0x2D01 is the same as the default in LowPowerLab RFM69 library
-  uint8_t syncWord[] = {0x2D, 0x01};
-  state = setSyncWord(syncWord, 2);
+  uint8_t syncWord[] = RF69_DEFAULT_SYNC_WORD;
```
jgromes: I'd rather keep the default 0x2D01 sync word, as per the comment.

examples/CC1101/CC1101_Receive_Fixed/CC1101_Receive_Fixed.ino
```diff
@@ -0,0 +1,98 @@
+/*
```
jgromes: Remove this example as per my comment in #62

src/modules/RF69.cpp
```diff
@@ -636,6 +642,71 @@ int16_t RF69::variablePacketLengthMode(uint8_t maxLen) {
...
+  // disable sync word detection and generation
+  state = _mod->SPIsetRegValue(RF69_REG_SYNC_CONFIG, RF69_SYNC_OFF | RF69_FIFO_FILL_CONDITION, 7, 6);
+  if (state != ERR_NONE) {
+    return(state);
+  }
+
+  return(state);
+}
+
+int16_t RF69::enableCrcFiltering() {
```
jgromes: Same as the comment for CC1101 - just one method `setCRCFiltering(bool)` is enough here.

src/modules/RF69.cpp
```diff
@@ -636,6 +642,71 @@ int16_t RF69::variablePacketLengthMode(uint8_t maxLen) {
...
+  return(_mod->SPIsetRegValue(RF69_REG_PACKET_CONFIG_1, RF69_CRC_ON), 4, 4);
+}
+
+int16_t RF69::disableCrcFiltering() {
+  return(_mod->SPIsetRegValue(RF69_REG_PACKET_CONFIG_1, RF69_CRC_OFF), 4, 4);
+}
+
+int16_t RF69::promiscuousMode(bool promiscuous) {
```
jgromes: Change name to `setPromiscuousMode(bool)`

src/modules/RF69.h
```diff
@@ -724,6 +728,45 @@ class RF69: public PhysicalLayer {
...
     int16_t variablePacketLengthMode(uint8_t maxLen = RF69_MAX_PACKET_LENGTH);
 
+     /*!
+      \brief Enable sync word filtering and generation.
+
+      \param numBits Sync word length in bits.
+
+      \returns \ref status_codes
+    */
+    int16_t enableSyncWordFiltering(uint8_t numBits = 16);
```
jgromes: The parameter should be `maxErrBits`, not the sync word length, right?

jgromes:
> I'm not aware of a way to disable sync word detection alone in CC1101

Neither am I, sorry. Then again, I don't use that radio much - it was added on request.

Also, please ignore the Travis errors - it's because ESP32 builds fail since Espressif's website returns timeouts at the moment.

phretor:
Sorry, not ready yet.

src/modules/RF69.cpp
phretor: I don't get it, isn't `RF69_DEFAULT_SYNC_WORD` set to `{0x2D, 0x01}`?

jgromes: Ah, my mistake - I got confused with the changed default sync word in CC1101. There's no reason CC1101 should be kept compatible with LowPowerLab library, so it's fine as it is.

PR #64, opened 2019-11-12
Removed memory allocation and copying from SX126X

BarryPSmith:
Extends command interface to allow multi-byte commands.

This allows functions that need to prepend some parameters onto input data to do so without allocating memory and performing a memcpy.

Benefits:
 - SX126x code can now be used without linking in operator new
 - Allows transmit and receive to be called when not much memory is available: previously these functions could allocate as much as 257 bytes when dealing with large messages.
 - Improved performance by removing memcpy


jgromes:
Looks good, was this tested on hardware?

BarryPSmith:
Yes, this is tested.

Aside: I noticed that my pull requests might conflict with each other. If this happens let me know and I can rebase the remaining ones and resolve the conflicts..

jgromes:
Merged, thanks.

There's no conflict in this one, though there's a minor one in #61

PR #66, opened 2019-11-13
Revisioned promiscious-mode implementation for CC1101 and RF69 #63 

phretor:
Promiscuous mode is needed in sniffing scenarios, similarly to variable vs. fixed packet mode. I also made `setSyncWord` uniform, retaining the original signature for compatibility.

I'm not aware of a way to disable sync word detection alone in CC1101. The only way I know also disables preamble detection altogether. Therefore, `disableSyncWordFiltering` also disables preamble filtering.

If you know a better way to address this, I'll commit some changes.


jgromes:
Merged, release most likely later this week.

Thank you very much for this contribution, I really appreciate it.

PR #71, opened 2019-11-18
CC1101 Variable Preamble Length

Guglio95:


jgromes:
Thanks for the contribution, few remarks:
1. Could you add preambleLength parameter to the `CC1101::begin()` method, set to 4 by default? Other FSK radio modules have this parameter.
2. The doxygen comment should mention in the /param section what is the unit of preamble length - bytes, bits, something else?

Other than that, everything looks fine.

jgromes:
The Travis error is due to ESP8266 breaking SoftwareSerial again, so I think we can merge this, thanks!

Guglio95:
Yep! I had the same problem but tought I had an obsolete esp8266 framework :)
Seems like you already fixed it!

PR #77, opened 2019-11-30
[CC1101] The synchronization word is a two-byte value

Guglio95:
Hi Jan,
I think there's a small inaccuracy since [CC1101 Datasheet](https://www.ti.com/lit/ds/symlink/cc1101.pdf) at page 38 states: "The synchronization word is a two-byte value"


jgromes:
You're right - I didn't read the part that says

>  A one-byte sync word can be emulated by setting the SYNC1 value to the preamble pattern.'

Merged, thanks!

Issue #80, opened 2019-12-01
SX126x do not check CRC


BarryPSmith:
**Describe the bug**
The CRC of incomming packets is not checked. Garbled packets are passed through readData(...) with no indication that the data is bad.

I don't know whether or not Tx packets have a CRC at all, because I only have these units to test with. I am certain that no CRC check is done because at range I was getting corrupted packets with no software errors to indicate they weren't correct (until they failed to process downstream).

I hope to have a PR for this in a short while.

**To Reproduce**
Two sketches, one running receive and one running transmit. The receive sketch is supposed to check CRCs with the call to `lora.setCrc(1)`. The transmit sketch alternatively turns CRCs on and off. I would expect the packets without CRCs to fail. However, all packets are received fine.

(Note: These use 425MHz because I was having problems getting my units to Tx on 433. Change as appropriate for regulatory conditions)

**EDIT** After a bunch of testing, I realised that one needs to be in implicit header mode for a brute force demonstration like this. Changing `sx126x.h` so that the default parameter in `sx126x::setPacketParams` to `SX126X_LORA_HEADER_IMPLICIT` is necessary to make this demonstration work.

Transmit sketch:
```
#include <RadioLib.h>

#define LORA_CHECK(a) do {\
      auto res = a; \
      if (res != ERR_NONE) \
      { \
        Serial.print("Error performing " #a ": "); \
        Serial.println(res); \
        while(1) {} \
      } else { \
        Serial.println(#a " successful."); \
      }} while(0)

#define DIO1 2
#define BUSY 4
#define SELECT 9

#define FREQ 425.0
#define SF 7
#define BW 62.5
#define CR 5

//Note: sx126x.h must be edited so that the radio is in implicit header mode for this demonstration.
Module mod(SELECT, DIO1, BUSY);
SX1262 lora = &mod;

void setup() {
  Serial.begin(38400);
  delay(10);
  Serial.println("starting...");
  //Note: Not using DIO3 as TCXO power.
  if (lora.begin(FREQ, BW, SF, CR, SX126X_SYNC_WORD_PRIVATE, 10, 140, 12, 0))
    LORA_CHECK(lora.begin(FREQ, BW, SF, CR, SX126X_SYNC_WORD_PRIVATE, 10, 140, 12, 0));
  LORA_CHECK(lora.setOutputPower(10));
  LORA_CHECK(lora.setCurrentLimit(140));
  LORA_CHECK(lora.setDio2AsRfSwitch());
[...]

PR #81, opened 2019-12-01
Enabled SX126x CRC check IRQs on receive

BarryPSmith:
It seems that IRQs don't show up in `getIrqStatus()` if they're not globally enabled (set 1 in the first parameter of `setDioIrqParams`). As a result, CRC errors weren't being detected in `readData` because the IRQs weren't enabled in `startReceive`.

This addresses issue #80. The issue report details how to test it. It was an easy fix, but surprisingly painful to test (not difficult, just need to get everything lined up).

The datasheet is misleading in this respect: It never explains the purpose of the first parameter of `setDioIrqParams`, and gives bad advice in `14.3` step `7`.

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -444,7 +444,7 @@ int16_t SX126x::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
 
 int16_t SX126x::startReceive(uint32_t timeout) {
   // set DIO mapping
-  int16_t state = setDioIrqParams(SX126X_IRQ_RX_DONE | SX126X_IRQ_TIMEOUT, SX126X_IRQ_RX_DONE);
+  int16_t state = setDioIrqParams(SX126X_IRQ_RX_DONE | SX126X_IRQ_TIMEOUT | SX126X_IRQ_CRC_ERR | SX126X_IRQ_HEADER_ERR, SX126X_IRQ_RX_DONE);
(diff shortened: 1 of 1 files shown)


jgromes:
Yet another mistake in SX126x docs? Why am I not surprised anymore ...

Though to be fair, I probably should have been more suspicious - I don't recall ever seeing CRC error with SX126x.

jgromes:
Merged, thanks!

Btw, with all the bugs you're finding, I think we ought to release a bugfix patch (2.0.1) - unless you're planning on reporting more ;)

BarryPSmith:
Unrelated to this PR, but if you're thinking of making a new release: I've got a patch that eliminates much of the floating point arithmetic from the SX126x module if you want to include that.

It's quite extensive so I need to do some decent testing before submitting it though (and actually make sure it's a worthwhile thing).

The reason for the patch is the atmega doesn't have hardware floating point, so FP takes alot of code space. For instance, getTimeOnAir is 1kB in one of my programs, and setBandwidth is even larger (the BW can be configured remotely so I can't rely on a compile time optimisation to eliminate the FP).

jgromes:
This is an interesting proposal - could you open a new PR for that?

I'm not entirely sure if it will be merged, the goal is not to have minimal footprint, rather to be easy to use. That's why there's so many floats in the public API. I guess that could be kept if we rewrite the internals so that the only place with floats is the public API methods.

PR #83, opened 2019-12-04
Removed some sx126x floating point logic to save code space

BarryPSmith:
Removed floating point logic from `sx126x::setBandwidth`, `sx126x::getTimeOnAir` and `sx126x::transmit` to save code space. These were flagged as some of the larger functions in my project.

This maintains the same public function signatures, and doesn't significantly impact readability (in my opinion).

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -176,11 +176,11 @@ int16_t SX126x::transmit(uint8_t* data, size_t len, uint8_t addr) {
   uint8_t modem = getPacketType();
   if(modem == SX126X_PACKET_TYPE_LORA) {
     // calculate timeout (150% of expected time-on-air)
-    timeout = (float)getTimeOnAir(len) * 1.5;
+    timeout = (getTimeOnAir(len) * 3) / 2;
 
   } else if(modem == SX126X_PACKET_TYPE_GFSK) {
     // calculate timeout (500% of expected time-on-air)
-    timeout = (float)getTimeOnAir(len) * 5.0;
+    timeout = getTimeOnAir(len) * 5;
 
   } else {
     return(ERR_UNKNOWN);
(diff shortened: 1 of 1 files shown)


BarryPSmith:
I had a larger changeset that added a second interface of all public functions, which accepted integer functions. For instance `setFrequenct(float freq, bool calibrate)` now had an integer counterpart `setFrequency_i(uint32_t freq_Hz, bool calibrate)`, and so on for all functions. The idea was that the user could choose to use only those functions, and avoid pulling in floating point.

However, it significantly impacted code readability and maintainability. Then I found out that GCC can't optimise away virtual functions, so all of the floating point functions are still pulled in. I feel a bit dumb about the amount of effort I put into that, but it's learning at least.

BarryPSmith:
This was tested by:
 - enabling god mode.
 - commenting out the lines of code that actually spoke the hardware.
 - creating a sketch that called setBandwidth(...) with all allowed parameters, and checking that the correct code path was followed with some Serial.println(...) [These printlns have since been removed]
 - Adjusting internal variables and calling getTimeOnAir with all possible bandswidths and spreading factors, and packet lengths ranging from 0 to 255 bytes in 15 byte intervals. Calculated values were checked against the value calculated be the old code. All agreed to within a small fraction of a percent.

I know the primary intent of the library is usability rather than small footprint. But if it can have a smaller footprint with the same public signature and without a significant cost in maintainability, then that makes it more useable because sketches will upload faster! Well, that's how I justify spending my days doing this anyway.

jgromes:
Even though the amount of saved Flash is not that dramatic (roughly 300 bytes when Travis is building SX126x_Receive and SX126x_Transmit on Arduino Uno), I still think this is worth it. Perhaps something similar could be done with other methods that use floating point arithmetic - other classes could be adapted too, not just SX126x.

There's a few things I'd like to address before merging (mostly formatting). It will probably be faster if I do it and push to your branch - let me know if you're OK with that.

BarryPSmith:
No problem if you want to make commits. I assume it's already set up since I made the pull request?

I see different things on code space depending on how I use it. If setBandwidth is called with compile time constants, then I think the compiler can get rid of all of the branches. I guess this is the most common case, but I wanted to make it customisable in my project, and it ended up being quite large once I was loading the BW from EEPROM.

I can't remember why I did getTimeOnAir and transmit. Probably just thought it'd be easy to pull that part from the full-integer version.

As I say, I managed to write functions that can duplicate the current logic of everything in SX126x without floating point, but decided to only make a PR for these changes because the code got a fair bit harder to read in places, and there weren't very good gains (because as far as I can tell, it's not until you eliminate FP entirely that things get decent, and even then it's only 2-3 kilobytes).

I ended up playing games to avoid overflow without losing precision. For instance, setFrequencyRaw had this line in it:
``` c
uint32_t frf = (((((freq_Hz * 4) / 125) * 128) / 125) * 32)
```
kinda pretty in a twisted way, but it just looks... difficult to work with.

jgromes:
I use unmodified examples to compare Flash usage between versions - the assumption is that most people will start from those.

`uint32_t frf = (((((freq_Hz * 4) / 125) * 128) / 125) * 32)` does look weird :P Perhaps fixed point variables could be used to get a compromise in readability while avoiding floats completely? Just an idea though.

jgromes:
Merged, thanks!

BarryPSmith:
Hmmm. Hadn't thought about fixed point. A quick test of the arduino IDE suggests that they don't have GCC configured to accept fractions (or it's running an old enough compiler that it's not there): `short _Fract fraction;` (or `fract fraction;` after `#include <stdfix.h>`) refuses to compile. Without native compiler support, it doesn't seem user friendly. To be honest though I didn't know that fixed point was even being considered for C inclusion, so perhaps I'm wrong on this.

If you want to see what a full integer implementation could like like, check out:
https://github.com/BarryPSmith/RadioLib/tree/sx126x_integer/src/modules/SX126x

Specifically, SX126x.h & .cpp and SX1262.h & cpp (when I realised I wasn't going to make a pull request for it, I didn't do the modules I wasn't using).

Some of the function signatures are a bit odd - when choosing between ease of use/readability and performance/code size, I probably veered too much on performance (e.g. setBandwidth_i could take a 32 bit integer to accept frequency in Hz, but I chose a 16 bit integer with bandwidth in hecto-hertz: hHz).

It's not tested much. I'm using it in my project (slightly modified, and using godmode to call the protected _i functions) and it seems to work, but I only use a small subset of the library. Also in order to get GCC to not pull in the floating point stuff I had to change the class declarations to not derive from PhysicalLayer. That was what convinced me it wasn't worthwhile to make a PR.

(In the end it seems I'm going to have to switch to a different [non-arduino supporting] processor at some point anyway, so this was a case of premature optimisation. Oh well, it's all learning.)

PR #84, opened 2019-12-04
Added sx126x receive duty cycle interface

BarryPSmith:
Added startReceiveDutyCycle with manually configurable duty cycle (basically raw interface to the device).

Added startReceiveDutyCycleAuto that configures duty cycle for current modem parameters (ideal duty cycle parameters are determined by symbol duration and sender preamble length).


src/modules/SX126x/SX126x.cpp
```diff
@@ -443,7 +443,91 @@ int16_t SX126x::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
...
+    return(ERR_INVALID_SLEEP_PERIOD);
+  }
+  sleepPeriod_us -= transitionTime;
+
+  // divide by 15.625
+  uint32_t rxPeriodRaw = (rxPeriod_us * 8) / 125;
+  uint32_t sleepPeriodRaw = (sleepPeriod_us * 8) / 125;
+
+  // 24 bit limit:
+  if ((rxPeriodRaw & 0xFF000000) || (sleepPeriodRaw & 0xFF000000)) {
```
jgromes: This is a bit too generic - caller doesn't know if the error here is caused by `rxPeriod` or `sleepPeriod`. It might be a better idea to check them independently and throw specific errors (e.g. ERR_SLEEP_PERIOD_TOO_LONG and ERR_RX_PERIOD_TOO_LONG).

src/modules/SX126x/SX126x.cpp
```diff
@@ -443,7 +443,91 @@ int16_t SX126x::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
...
+  return(state);
+}
+
+int16_t SX126x::startReceiveDutyCycle(uint32_t rxPeriod_us, uint32_t sleepPeriod_us)
+{
+  // datasheet claims time to go to sleep is ~500us, same to wake up.
+  // compensate for that 1ms + tcxo delay:
+  uint32_t transitionTime = _tcxoDelay_us + 1000;
+  if (sleepPeriod_us < transitionTime + 16) {
+    return(ERR_INVALID_SLEEP_PERIOD);
```
jgromes: ERR_SLEEP_PERIOD_TOO_SHORT might be a better name for the error (since it can be too short here, or too long and overflow).

src/modules/SX126x/SX126x.cpp
```diff
@@ -443,7 +443,91 @@ int16_t SX126x::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
...
+  if ((rxPeriodRaw & 0xFF000000) || (sleepPeriodRaw & 0xFF000000)) {
+    return(ERR_OVERFLOW);
+  }
+
+  int16_t state = startReceiveCommon();
+  if (state != ERR_NONE) {
+    return(state);
+  }
+
+  byte data[6] = {
```
jgromes: Please use the exact type, uint8_t instead of byte.

src/modules/SX126x/SX126x.cpp
```diff
@@ -443,7 +443,91 @@ int16_t SX126x::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
...
+  // in this case, we don't catch minSymbols before going to sleep,
+  // so we must be awake for at least that long before the sender stops transmitting.
+  uint16_t sleepSymbols = senderPreambleLength - 2 * minSymbols;
+
+  // if we're not to sleep at all, just use the standard startReceive.
+  if (2 * minSymbols > senderPreambleLength) {
+    return(startReceive());
+  }
+
+  uint32_t symbolLength_us = ((uint32_t)(10 * 1000) << _sf) / (10 * _bwKhz);
```
jgromes: I'm not entirely sure this is correct, since the symbol duration equation should be: Ts = (2^SF)/BW

src/modules/SX126x/SX126x.h
```diff
@@ -499,6 +499,36 @@ class SX126x: public PhysicalLayer {
...
+    /*!
+      \brief Interrupt-driven receive method where the device mostly sleeps and periodically wakes to listen.
+
+      \param rxPeriod_us The duration the receiver will be in Rx mode, in microseconds.
+
+      \param sleepPeriod_us The duration the receiver will not be in Rx mode, in microseconds.
+
+      \returns \ref status_codes
+
+      Note that this function assumes the unit will take 500us to change state. See datasheet [SECTION].
```
jgromes: Please include datasheet section and version when quoting datasheets.

jgromes:
Thanks, I'm assuming this was tested with hardware, right?

src/modules/SX126x/SX126x.cpp
BarryPSmith: Would it be OK to just have `ERR_INVALID_SLEEP_PERIOD` and `ERR_INVALID_RX_PERIOD`? In most other radiolib functions there aren't distinct errors for too large and too small.

jgromes: That's true actually, let's not compilcate things. ERR_INVALID_SLEEP_PERIOD and ERR_INVALID_RX_PERIOD it is.

src/modules/SX126x/SX126x.cpp
BarryPSmith: The 1000 is because symbolLength is in us and bw is in kHz:
Ts = (2^SF)/Bw.
Ts * 1E6 = (2^SF)/Bw * 1E6
Ts(us) = (2^SF)/(Bw(kHz) * 1000) * 1E6
Ts(us) = (2^SF)/(Bw(kHz)) * 1E3
Ts(us) = (1E3 * 2^SF) / BwKHz

Then it multiplies _bwKhz by 10 because it might be as low as 7.8. It was supposed to be cast to integer before the division, looks like I missed that. Anyway, multiply the numerator and denominator by the same number and the equation is unchanged, so that's why there's the 10 in the top too. I guess that could be left out and it can just work in FP.

src/modules/SX126x/SX126x.h
BarryPSmith: ah yes. That was the intention, hence the placeholder. Done.

src/modules/SX126x/SX126x.cpp
BarryPSmith: As below: How about just having the too long have the same error (but error changes depending on variable), for consistency with the rest of the library?

BarryPSmith:
Yep, it's tested. Just two units talking to each other on my desk running modified examples, with `startReceive` replaced with `startReceiveDutyCycleAuto`.

That's how I got the default value of 3 in the setReceiveDutyCycleAuto for minSymbols.

I'm not going to get a chance to implement the changes until tomorrow; when I do that I'll hook a unit up to an oscilliscope and check that the timing is correct (should be easy to see whether it's in Rx or sleep based on the current draw).

As an aside: I just noticed the datasheet calls this 'listen' mode. `startListen` is a nicer name than `startReceiveDutyCycle`, but I don't like it because it's not clear how it's different from `startReceive`. Thoughts?

src/modules/SX126x/SX126x.cpp
jgromes: Checks out, thanks. I was confused as to how did the 10*1000 got there.

jgromes:
> I'm not going to get a chance to implement the changes until tomorrow

No rush, take your time.

>  I just noticed the datasheet calls this 'listen' mode

I see what you mean. Strictly speaking, there's nothing in the library called "listen" mode. But to me, listen mode sounds more like the way `startReceive` works - put the module into receive mode and just listen for packets without any interruptions. Probably doesn't help that `startReceive` is referred to as "listen" mode in the examples :P

BarryPSmith:
Requested changes are implemented.

I tested it with a scope. Works more or less as intended.

The sleep and wake times of 500us each in the datasheet aren't quite right (looks more like 400us + 300us), but in my opinion leaving the documented & datasheet backed 500+500us is better than putting in my experimentally determined values. Net result: the unit will sleep for slightly shorter than requested. This means slightly higher battery consumption when working with a quick duty cycle, but chance of missing a message shouldn't be compromised. Behaviour in radiolib is documented, so if the user wants to get their scope out and figure out the exact details of everything, they can still do that.

BarryPSmith:
Seb on the Semtech forum just posted that the receiver requires 8 symbols to reliably latch a preamble, I propose that this should be the default value of minSymbols in setRxDutyCycleAuto(...)

This makes it a useless function with the any preamble length less than of 17 symbols, but I guess if you want good power savings there has to be a compromise.

jgromes:
Agree - though I think the duty cycle methods will be used mainly by poeple who know what they're doing. We can add a note about the minimum required preamble length to the doxygen comments.

jgromes:
Merged, thanks!

PR #88, opened 2019-12-12
[CC1101] Output power updated according to chosen frequency

Guglio95:
Hi Jan,
I think output power register should be updated after changing frequency, otherwise the radio may be left in an inconsistent state since different frequencies have different values for the PA_TABLE.

For example if you call `setFrequency(433.0)` the PA_TABLE reg is still set to the default value for the 868Mhz (which is `0x50`) but the default value for the 434Mhz is different (`0x60`).

What do you think ?


jgromes:
You're right, this really is the case. Merged, thanks!

PR #90, opened 2019-12-13
[CC1101] Small fixes.

Guglio95:
Hi Jan,
I had some issues changing the sync word on my CC1101; 
I don't know why but setting it in burst mode was just applied to the SYNC1 register, SYNC0 was not affected. I put two [SPIsetRegValue](https://github.com/jgromes/RadioLib/compare/master...Guglio95:master#diff-bfbdc416744bf6d51bce6780aa1e9cbaR477). 
Then I fixed a wrong SYNC_MODE reg assignation.
Moreover i added a [reset on init](https://github.com/jgromes/RadioLib/compare/master...Guglio95:master#diff-bfbdc416744bf6d51bce6780aa1e9cbaR660) since the radio may still be configured by previous RadioLib instances.
Lastly I enabled the CRC check only if CRC validation is actually enabled.

Let me know.

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -720,8 +720,8 @@ int16_t CC1101::setPacketMode(uint8_t mode, uint8_t len) {
     return(ERR_PACKET_TOO_LONG);
   }
 
-  // set to fixed packet length
-  int16_t state = _mod->SPIsetRegValue(CC1101_REG_PKTCTRL0, mode, 7, 7);
+  // set PKTCTRL0.LENGTH_CONFIG
+  int16_t state = _mod->SPIsetRegValue(CC1101_REG_PKTCTRL0, mode, 1, 0);
   if (state != ERR_NONE) {
     return(state);
   }
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -477,7 +477,7 @@ int16_t CC1101::setSyncWord(uint8_t* syncWord, uint8_t len, uint8_t maxErrBits) 
   state = SPIsetRegValue(CC1101_REG_SYNC1, syncWord[0]);
   state |= SPIsetRegValue(CC1101_REG_SYNC0, syncWord[1]);
 
-  return(ERR_NONE);
+  return (state);
 }
 
 int16_t CC1101::setSyncWord(uint8_t syncH, uint8_t syncL, uint8_t maxErrBits) {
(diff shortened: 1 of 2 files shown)


Guglio95:
Oops I just spot an issue:
[Here](https://github.com/jgromes/RadioLib/pull/90/commits/af9409b941e9419dec5d5f7e45dc961e51c6744a#diff-bfbdc416744bf6d51bce6780aa1e9cbaL456) the len parameter refers to the buffer size (I suppose).

[Here](https://github.com/jgromes/RadioLib/pull/90/commits/af9409b941e9419dec5d5f7e45dc961e51c6744a#diff-bfbdc416744bf6d51bce6780aa1e9cbaR593) the _syncWordLength refers to the fact that the " 2 bytes sync word" may be repeated obtaining a 4 bytes sync word....

Fun fact, how to handle this?

src/modules/CC1101/CC1101.cpp
```diff
@@ -474,7 +474,8 @@ int16_t CC1101::setSyncWord(uint8_t* syncWord, uint8_t len, uint8_t maxErrBits)
   }
 
   // set sync word register
-  _mod->SPIwriteRegisterBurst(CC1101_REG_SYNC1, syncWord, len);
+  state = SPIsetRegValue(CC1101_REG_SYNC1, syncWord[0]);
+  state |= SPIsetRegValue(CC1101_REG_SYNC0, syncWord[1]);
```
jgromes: Coudl you double check the byte order is correct? CC1101 datasheet says CC1101_REG_SYNC1 is the high byte.

src/modules/CC1101/CC1101.cpp
```diff
@@ -474,7 +474,8 @@ int16_t CC1101::setSyncWord(uint8_t* syncWord, uint8_t len, uint8_t maxErrBits)
...
 
   return(ERR_NONE);
```
jgromes: We should return `state` here.

jgromes:
Do we need to support the case when sync word is repeated? I would argue not, but I don't really use CC1101. I only added it because it can do RTTY ;)

Guglio95:
Hi Jan,
You are right, we may remove it! I’m not using the repeated sync word 😏

I’m off the grid, will fix it on Monday 👍

Guglio95:
Hi Jan,

It should be ok now :+1: 
I found a wrong reg assignation (56cdbbd) 

Let me know.

jgromes:
Merged, thanks!

Issue #92, opened 2019-12-17
[CC1101] setModulation method.


Guglio95:
Hi Jan,

**Is your feature request related to a problem? Please describe.**
I'm using the CC1101 radio with OOK modulation, in RadioLib it's listed and supported only as a FSK Modem.

**Describe the solution you'd like**
I've already implemented the _basic_ [code](https://github.com/trendmicro/RFQuack/blob/RadioLib/src/radio/RFQCC1101.h#L19) to set it in OOK mode. If you are interested I'll send a pull request to RadioLib. (My implementation is not complete yet, it misses a setOutputPower fix).

**Describe alternatives you've considered**
If so, do you have some tips/preferences? How should I call the method? `setModulation(enum)`

Let me know.

jgromes:
There's a similar feature implemented in the SX127x via `int16_t SX127x::setOOK(bool enableOOK)` - I think something similar would be appropriate for the CC1101 driver.

Guglio95:
Super, I'll send a PR!

_PS: I fixed the link_

PR #93, opened 2019-12-20
[CC1101] Added setOOK(bool enableOOK)

Guglio95:
Refers to https://github.com/jgromes/RadioLib/issues/92


jgromes:
Looks good, thanks. Were you able to test this, does it work as intended?

Guglio95:
I've tested `setOOK(true)` and works as expected. 

Subsequently I've tried `setOOK(false)` after `setOOK(true)` and the radio is transmitting back in FSK; I've no way to test the accuracy of the rollback ( same datarate etc..)

jgromes:
Merged, thanks!

PR #96, opened 2019-12-22
Reorganized morsecode symbols array and updated the MorseClient::writ…

IU5HKU:
The idea was taken from the CubeSat way of doing morsecode, we gained 510bytes and extended the morsecode to the whole ascii table.

### DIFF
--- a/examples/Morse/Morse_Transmit/Morse_Transmit.ino
+++ b/examples/Morse/Morse_Transmit/Morse_Transmit.ino
@@ -104,7 +104,7 @@ void loop() {
 
   // floating point number
   // NOTE: When using println(), the transmission will be
-  //       terminated with cross signal (.-.-.).
+  //       terminated with end-of-work signal (...-.-).
   float f = -3.1415;
   morse.println(f, 3);
 
--- a/src/protocols/Morse/Morse.cpp
+++ b/src/protocols/Morse/Morse.cpp
@@ -4,68 +4,79 @@
 /*!
   \cond RADIOLIB_DOXYGEN_HIDDEN
 */
-struct Morse_t {
-  char c;                                             // ASCII character
-  char m[7];                                          // Morse code representation
-};
-/*!
-  \endcond
-*/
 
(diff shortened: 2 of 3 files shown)


jgromes:
Thanks, I will push some formatting fixes to this branch if that's OK.

I did notice there's a Morse codes in the array for underscore, that doesn't seem to be a part of the ITU standard (https://www.itu.int/dms_pubrec/itu-r/rec/m/R-REC-M.1677-1-200910-I!!PDF-E.pdf), so I'll remove that.

Also, there are sone ITU Morse codes that do not have direct ASCII equivalent, like the starting signal or end of work. I think I'll append those at the end of the lookup table.

IU5HKU:
Hi Jan, sure, you can modify the source code as you want, for special
purposes chars I suggest to use the empty places inside the array, the
zeroed ones, in this way you have two benefits, save space and maintain the
low overall occupied space and have a char to rapresent the special morse
code inside the string you want to transmit, as example > or <...
Have a nice day and a great Xmas holiday!
Marco

Il lun 23 dic 2019, 09:58 Jan Gromeš <notifications@github.com> ha scritto:

jgromes:
Tested and working on Arduino Uno + SX1278, decoded using fldigi.

Merged, thanks - and Merry Christmas!

IU5HKU:
Thank you so much Jan, proud to take part (minimum part) to your
interesting project. just a question,i've just see that two weeks ago you
dropped the sx1278 support for the Fossasat LoRa Receiver....is it not
compatible with the one used onboard the sat? I've a pair of sx1278 in the
drawer that i was thinking to use for receive, but if you told me that they
don't work, well, i'll keep them in the drawer and i'll go for new
radiomodules (sx126x works better?). If you wanna give me other tasks,
well, let me know, i'm here.
Have a good and happy xmas holidays!
 Marco

Il giorno lun 23 dic 2019 alle ore 14:56 Jan Gromeš <
notifications@github.com> ha scritto:

Issue #86, opened 2019-12-11
[SX126x] Implicit mode support


naggie:
**Is your feature request related to a problem? Please describe.**
It's not currently possible to put the modem in Implicit header mode which is useful for low latency applications where the payload size is always known.

**Describe the solution you'd like**
I think a `SX126x::setHeaderType)` would work, though I'm not familiar with the library to know this is the best interface.

This would supply `SX126X_LORA_HEADER_EXPLICIT` or `SX126X_LORA_HEADER_IMPLICIT` to `setPacketParams()` where appropriate.

jgromes:
As far as I can tell, this shouldn't break compatibility with existing code, since all the parameters required for implicit header (payload length, coding rate and CRC presence) should by configured by the default values of `begin()`.

I think the interface should be similar to the way fixed/variable packet length mode is handled, i.e. use two public methods:

```
int16_t SX126x::implicitHeader();
int16_t SX126x::explicitHeader();
```

both of which call common protected `int16_t SX126x::setHeaderType()`.

Also, SX127x has implicit header mode too - perhaps the same interface should be added to that one as well.

naggie:
OK -- I assume calling one of those methods should set a variable which all calls to `setPacketParams` would be called with (whenever called by something else) rather than actually programming the radio at time of call.

Alternatively setPacketParams could use the member variable instead of taking a parameter (which is currently not used). What would you prefer?

jgromes:
The issue with this approach is that if the user only wants to change current header mode, he would then have to call some method which calls `setPacketParams` - there is at least one corner case when user could call `explicitHeader` (after having previously used implicit mode), then call some reception method (e.g. `receive`). There would be no call to `setPacketParams`, hence the header mode would remain implicit, even though user thinks explicit mode is being used.

Since there's no documented way to read the current configuration of SX126x, I think the safest approach is to call `setPacketParams`, despite the increased SPI traffic. `_preambleLength` and `_crcType` should be known at the time of call to `implicitHeader` or `explicitHeader`.

Also, length parameter must be added to `startReceive` and `startReceive` common, to set packet length in implicit header mode.

jgromes:
Closed due to inactivity - @naggie feel free to reopen later or submit a PR.

naggie:
Ok -- I will submit a PR when I return to work.

naggie:
[...]

PR #103, opened 2020-01-09
[SX126x] implicit mode

naggie:
Not complete yet. See https://github.com/jgromes/RadioLib/issues/86 for implementation discussion.

- [x] Declarations/interface
- [x] SetHeaderType
- [x] setImplicitHeader/setExplicitHeader
- [x] check getOnAirTime takes lack of header into account
- [x] make sure all calls to setPacketParams use cached _headerType and _implicitLen (new vars)
- [x] only set these vars in setHeaderMode, do not use setPacketParams. _Maybe remove setHeaderMode as it's not doing much since discussion_
- [x] receive methods should call setPacketParams in implicit mode only
- [x] decide what to do with existing specified len on receive()
- [x] test implementation on hardware

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -117,6 +117,8 @@ clearGdo2Action	KEYWORD2
 setTCXO	KEYWORD2
 setDio2AsRfSwitch	KEYWORD2
 getTimeOnAir	KEYWORD2
+implicitHeader
+explicitHeader
 setSyncBits	KEYWORD2
 setWhitening	KEYWORD2
 startReceiveDutyCycle	KEYWORD2
--- a/keywords.txt
+++ b/keywords.txt
@@ -117,8 +117,8 @@ clearGdo2Action	KEYWORD2
 setTCXO	KEYWORD2
 setDio2AsRfSwitch	KEYWORD2
 getTimeOnAir	KEYWORD2
-implicitHeader
-explicitHeader
+implicitHeader	KEYWORD2
+explicitHeader	KEYWORD2
 setSyncBits	KEYWORD2
 setWhitening	KEYWORD2
 startReceiveDutyCycle	KEYWORD2
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -1170,8 +1170,8 @@ uint32_t SX126x::getTimeOnAir(size_t len) {
   }
 }
 
-int16_t implicitHeader() {
-    return setHeaderType(SX126X_LORA_HEADER_IMPLICIT);
+int16_t implicitHeader(size_t len) {
+    return setHeaderType(SX126X_LORA_HEADER_IMPLICIT, len);
 }
 
 int16_t explicitHeader() {
@@ -1354,9 +1354,9 @@ int16_t SX126x::setPacketMode(uint8_t mode, uint8_t len) {
   return(state);
 }
 
-int16_t SX126x::setHeaderType(uint8_t headerType) {
+int16_t SX126x::setHeaderType(uint8_t headerType, size_t len) {
   // set requested packet mode
-  int16_t state = setPacketParams(_preambleLength, _crcType, 0xFF, headerType);
+  int16_t state = setPacketParams(_preambleLength, _crcType, len, headerType);
 
   if(state != ERR_NONE) {
     return(state);
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -743,7 +743,7 @@ int16_t SX126x::setPreambleLength(uint16_t preambleLength) {
   uint8_t modem = getPacketType();
   if(modem == SX126X_PACKET_TYPE_LORA) {
     _preambleLength = preambleLength;
-    return(setPacketParams(_preambleLength, _crcType, 0xFF, _headerType));
+    return(setPacketParams(_preambleLength, _crcType, _implicitLen, _headerType));
   } else if(modem == SX126X_PACKET_TYPE_GFSK) {
     _preambleLengthFSK = preambleLength;
     return(setPacketParamsFSK(_preambleLengthFSK, _crcTypeFSK, _syncWordLength, _addrComp, _whitening, _packetType));
@@ -1048,7 +1048,7 @@ int16_t SX126x::setCRC(uint8_t len, uint16_t initial, uint16_t polynomial, bool 
       _crcType = SX126X_LORA_CRC_OFF;
     }
 
-    return(setPacketParams(_preambleLength, _crcType, 0xFF, _headerType));
+    return(setPacketParams(_preambleLength, _crcType, _implicitLen, _headerType));
   }
 
   return(ERR_UNKNOWN);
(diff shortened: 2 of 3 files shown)


src/modules/SX126x/SX126x.cpp
```diff
@@ -1345,6 +1354,19 @@ int16_t SX126x::setPacketMode(uint8_t mode, uint8_t len) {
   return(state);
 }
 
+int16_t SX126x::setHeaderType(uint8_t headerType) {
+  // set requested packet mode
+  int16_t state = setPacketParams(_preambleLength, _crcType, 0xFF, headerType);
```
naggie: I hope using 0xFF here is the right thing to do, as it will presumably be overridden before TX/RX operations

src/modules/SX126x/SX126x.cpp
```diff
@@ -598,6 +608,14 @@ int16_t SX126x::startReceiveCommon() {
   // clear interrupt flags
   state = clearIrqStatus();
 
+  // set implicit mode and expected len if applicable
+  if (_headerType == SX126X_LORA_HEADER_IMPLICIT && getPacketType() == SX126X_PACKET_TYPE_LORA) {
```
naggie: `getPacketType` is an extra SPI expense but I think it's necessary. Is it?

src/modules/SX126x/SX126x.h
```diff
@@ -810,7 +826,7 @@ class SX126x: public PhysicalLayer {
     int16_t setTxParams(uint8_t power, uint8_t rampTime = SX126X_PA_RAMP_200U);
     int16_t setModulationParams(uint8_t sf, uint8_t bw, uint8_t cr, uint8_t ldro = 0xFF);
     int16_t setModulationParamsFSK(uint32_t br, uint8_t pulseShape, uint8_t rxBw, uint32_t freqDev);
-    int16_t setPacketParams(uint16_t preambleLength, uint8_t crcType, uint8_t payloadLength = 0xFF, uint8_t headerType = SX126X_LORA_HEADER_EXPLICIT, uint8_t invertIQ = SX126X_LORA_IQ_STANDARD);
+    int16_t setPacketParams(uint16_t preambleLength, uint8_t crcType, uint8_t payloadLength, uint8_t headerType, uint8_t invertIQ = SX126X_LORA_IQ_STANDARD);
```
naggie: I made these parameters required -- I think they should be to ensure _headerType and _implicitLen is always considered.

jgromes:
Ran into a possible issue while testing this, transmit methods will override whatever implicit length was set in `implicitHeader()` (there's a call to setPacketParams). Not really sure what to do about it, we could restore the original payload length but that would only solve blocking transmit, non-blocking will always override implicit payload length in setPacketParams.

src/modules/SX126x/SX126x.cpp
```diff
@@ -288,6 +290,14 @@ int16_t SX126x::receive(uint8_t* data, size_t len) {
...
     timeout = (uint32_t)(symbolLength * 100.0 * 1000.0);
 
+      // set implicit mode and expected len if applicable
+      if (_headerType == SX126X_LORA_HEADER_IMPLICIT) {
+          state = setPacketParams(_preambleLength, _crcType, _implicitLen, _headerType);
+          if(state != ERR_NONE) {
+            return(state);
+          }
+      }
+
```
jgromes: This is not necessary, there's a call to `startReceive` later, which in turn will call `startReceiveCommon` that already calls setPacketParams

src/modules/SX126x/SX126x.cpp
```diff
@@ -1169,6 +1187,16 @@ uint32_t SX126x::getTimeOnAir(size_t len) {
   }
 }
 
+void SX126x::implicitHeader(size_t len) {
```
jgromes: Please add both of these to keywords.txt using true tab.

jgromes: Also, both of these methods should check active modem - implicit/explicit header only applies to LoRa.

naggie: I believe the active modem should be checked at RX/TX -- it could change after calling these methods. Though, if we're OK with that risk it might be simpler.

src/modules/SX126x/SX126x.cpp
naggie: Quite right, missed that -- thanks.

src/modules/SX126x/SX126x.cpp
jgromes: You're right, it is necessary in this case.

src/modules/SX126x/SX126x.cpp
jgromes: Modem is already checked at Rx/Tx, this is purely for safety reasons - if the user attempts to change LoRa-only setting while FSK modem is active, it's a sign they're doing something wrong. I also still think both of these should call setPacketParams to send the changed header type into the radio itself - that's probably the safest option.

keywords.txt
```diff
@@ -117,6 +117,8 @@ clearGdo2Action	KEYWORD2
 setTCXO	KEYWORD2
 setDio2AsRfSwitch	KEYWORD2
 getTimeOnAir	KEYWORD2
+implicitHeader
+explicitHeader
```
jgromes: The new keywords seem to be missing the type specifier, i.e.

impicitHeader<tab>KEYWORD2

naggie:
> Ran into a possible issue while testing this, transmit methods will override whatever implicit length was set in `implicitHeader()` (there's a call to setPacketParams). Not really sure what to do about it, we could restore the original payload length but that would only solve blocking transmit, non-blocking will always override implicit payload length in setPacketParams.
> 
> EDIT: realized there's also a "reverse" of the above effect in receive methods - if you specify length that is different than payload length set in implicitHeader, you can still receive the data, albeit incomplete/with garbage at the end.

I can think of 3 options:

1. Always requiring a length for TX/RX at point of use (instead of before with `implicitHeader()`) would prevent surprises and is clean, but break the  compatibility with the module interface unless we change it but that would implicate all radios...
2. Requiring a matching length if set and implicit mode is activated (safe, easy but a bit odd)
3. Restoring length as you say but atomicity is an issue with interrupt receive.

jgromes:
I can think of 4th one - not worrying about it :P

At the core, implicit mode requires payload length to be known in advance, which means that it must be known to the top level Arduino sketch at the time of calling implicitHeader/transmit/receive. I would also consider it more advanced than the default explicit mode, so I think we can leave it up to the user.

naggie:
> I can think of 4th one - not worrying about it :P
> 
> At the core, implicit mode requires payload length to be known in advance, which means that it must be known to the top level Arduino sketch at the time of calling implicitHeader/transmit/receive. I would also consider it more advanced than the default explicit mode, so I think we can leave it up to the user.

Even better! I've just tested the code, I think it's ready. Thanks for taking the time to discuss and review.

naggie:
oops, just force pushed by mistake, sorry. Turns out I fixed the keywords but did not push

jgromes:
No worries, I think this about wraps things up here ;)

Thank you very much for the contribution, I really appreciate it! Will release this as 3.1.0.

naggie:
No problem, glad to help!

PR #106, opened 2020-01-28
[SX126x] Parameter to use LDO in begin()

naggie:
See https://github.com/jgromes/RadioLib/commit/4231edee706759d29ef31f201e4110c7e2cf00f9
for context.

This allows compatibility with https://www.rfsolutions.co.uk/radio-modules-c10/dip-lora-transceiver-module-20km-868mhz-p807 (failure mode resulted in unreliable RX with excessive current)

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -4,7 +4,7 @@ SX126x::SX126x(Module* mod) : PhysicalLayer(SX126X_CRYSTAL_FREQ, SX126X_DIV_EXPO
   _mod = mod;
 }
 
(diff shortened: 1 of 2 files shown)


jgromes:
Looks, good, thanks for the PR. Will release a patch 3.2.1 later today.

naggie:
Wow that was fast. Thanks @jgromes !

naggie:
Actually I realised I didn't add the parameter to the 1262, 1268, 1261 begin -- sorry. Now adding for another PR>

jgromes:
Sionce it's a common setting acrros all SX126x modules, it should be probably added to the base SX126x class.

naggie:
I agree it should be in the base class -- I just meant that as the subclasses define their own `begin` the parameters need to be passed over. I've done a PR to that affect https://github.com/jgromes/RadioLib/pull/107

PR #107, opened 2020-01-28
[SX126x] Pass useRegulatorLDO to SX1262/SX1261/SX1268 subclasses

naggie:


jgromes:
Looking even better now, thanks ;) I'll add a note to all the examples about default regulator selection and release the patch after that.

naggie:
Ok. As for the regulator selection it's looking like a batch defect with the inductor in the modems I have as the cause.

jgromes:
That's what I was wondering - it seems like the module you linked to is actually using DC-DC too.

naggie:
It turns out the "defect" was that the SMPS was simply not implemented in the Lambda module. The LDO has to be used instead.

PR #108, opened 2020-01-29
Prevent spurious resets on some boards

naggie:
My receiver was failing to receive after a random amount of time (2 - 60
seconds). I discovered some power supply issues (DC-DC converter
related) that turned out to be another cause of the same problem but
only on some boards. I also noted that reinitialising the radio temporarily fixed the problem.

The reset procedure for most of the boards that RadioLib can drive
changes the pin mode of the reset line to an input after reset,
effectively tri-stating the output. I had seen this but dismissed it
after checking that the SX126x has a pullup on NRST meaning this was not
an issue.

The receiver I have produced uses a level converter to translate the 5v0
signals to 3v3. The level converters are not themselves pulled up or
down, which means when a pin is connected in a high-impedance input
state it will float around possibly randomly.

This can cause spurious resets on my board, and possibly others. I
remembered the reset procedure when I realised I could reproduce the
problem by rubbing the board on my shirt, probably causing some ESD to
trigger a change on the reset line.

This PR simply removes the lines that change the pinmode to input after
reset leaving it as an output which is hard-driven and the safest way. I
assume that the current behaviour was chosen to decrease the chance of a
conflict if used incorrectly.

Note that the NSS line is already safe (always an output).

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -95,7 +95,6 @@ void RF69::reset() {
   Module::digitalWrite(_mod->getRst(), HIGH);
   delayMicroseconds(100);
   Module::digitalWrite(_mod->getRst(), LOW);
-  Module::pinMode(_mod->getRst(), INPUT);
   delay(10);
 }
 
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -155,7 +155,6 @@ int16_t SX126x::reset(bool verify) {
   Module::digitalWrite(_mod->getRst(), LOW);
   delayMicroseconds(150);
   Module::digitalWrite(_mod->getRst(), HIGH);
-  Module::pinMode(_mod->getRst(), INPUT);
 
   // return immediately when verification is disabled
   if(!verify) {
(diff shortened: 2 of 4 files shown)


jgromes:
The current behavior was chosen on purpose, because SX127x datasheet suggests the reset line should be set to high impedance state after the reset (and beacuse I've also seen this as common practice on previous projects):

Screenshot_3

This looks like a very hardware-dependent issue, then again, there doesn't seem to be any harm in driving the pin.

jgromes:
Out of curiosity - what sort of level shifters are you using?

naggie:
I see -- that makes sense. Perhaps my board should have a pull-up before the level converter to emulate this behaviour. Thanks for merging.

I'm using a pair of TI LVC8T245 transceivers.

w0067814:
It would be good to wrap the GPIO definitions such as the reset line in a #DEFINE or similar which determines whether the lines are Open-Drain or Push-Pull.

When power may be removed from a device that is connected to a host with Push-Pull GPIO, it is necessary to ensure that all GPIO lines are driven LOW when power removed to ensure that the device is not back powered through the GPIO lines, and thus eliminate leakage current. With Open-Drain connections it is not necessary to drive the host's GPIO lines low when the device is powered off providing they are pulled up to the device's supply rail and not the host's.

My personal preference is to drive critical lines such as reset from a stiff voltage source (Push-Pull) as they are less susceptible to induced noise from EMI events such a ESD discharge when compared to high impedance Open-Drain lines. A compromise is to make the pull-up resistor very strong however the leakage current implications must be considered along with how long the line shall sit in the LOW state.

jgromes:
@w0067814 I'm not sure I entirely undestand how open-drain/push-pull could be implemented here - could you elaborate (or better yet, open a new issue/PR).

PR #111, opened 2020-02-08
LoRa: Add ability for receiver to check if CRC is present in payload

mcgurk:
setCRC() has no effect when used with receiver in LoRa explicit header mode (SX1276/77/78/79 datasheet page 30). This modification adds crcOnPayload()-function, which can be used to detect if CRC is present in payload. With this it is possible to verify that CRC is really used between sender and receiver.

SX1276/77/78/79 datasheet page 112:
RegHopChannel (0x1C), bit 6: CrcOnPayload (read)
CRC Information extracted from the received packet header (Explicit header mode only)
0 = Header indicates CRC off
1 = Header indicates CRC on

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -90,6 +90,7 @@ setDataShaping	KEYWORD2
 setOOK	KEYWORD2
 setDataShapingOOK	KEYWORD2
 setCRC	KEYWORD2
+crcOnPayload	KEYWORD2
 variablePacketLengthMode	KEYWORD2
 fixedPacketLengthMode	KEYWORD2
 setCrcFiltering KEYWORD2
--- a/src/modules/SX127x/SX1278.cpp
+++ b/src/modules/SX127x/SX1278.cpp
@@ -439,6 +439,10 @@ int16_t SX1278::setCRC(bool enableCRC) {
   }
 }
 
+uint8_t SX1278::crcOnPayload() {
+  return(_mod->SPIgetRegValue(SX127X_REG_HOP_CHANNEL, 6, 6) >> 6);
+}
+
 int16_t SX1278::setBandwidthRaw(uint8_t newBandwidth) {
   // set mode to standby
   int16_t state = SX127x::standby();
--- a/src/modules/SX127x/SX1278.h
+++ b/src/modules/SX127x/SX1278.h
@@ -259,6 +259,13 @@ class SX1278: public SX127x {
     */
     int16_t setCRC(bool enableCRC);
 
+	  /*!
+      \brief Check if payload includes CRC and thus CRC is enabled in explicit header mode
+
+      \returns 1 if payload includes CRC and CRC is enabled, otherwise 0
+    */
+	  uint8_t crcOnPayload();
+
 #ifndef RADIOLIB_GODMODE
   protected:
 #endif
--- a/src/modules/SX127x/SX1278.h
+++ b/src/modules/SX127x/SX1278.h
@@ -259,12 +259,12 @@ class SX1278: public SX127x {
     */
     int16_t setCRC(bool enableCRC);
 
-	  /*!
+    /*!
       \brief Check if payload includes CRC and thus CRC is enabled in explicit header mode
 
       \returns 1 if payload includes CRC and CRC is enabled, otherwise 0
     */
-	  uint8_t crcOnPayload();
+    uint8_t crcOnPayload();
 
 #ifndef RADIOLIB_GODMODE
   protected:


jgromes:
What's the rationale behind exposing `crcOnPayload()` method to the user? If anything, this should be handled internally, inside `SX127x::readData()`.

Also, I'm not sure this is even necessary, since the datasheet on page 30 says:
> If the bit CrcOnPayload is at ‘0’, it means there was no CRC on the payload and thus the IRQ Flag PayloadCrcError will not be trigged even if the payload has errors

mcgurk:
I have case where I want to be sure that data verified with CRC. Without crcOnPayload I have no way to check in receiver if CRC is used.

jgromes:
I'd argue you do, since you have to be the one to call `setCRC()` - until you do that, CRC remains in the default state, i.e. enabled. It also stands to reason that you have the ability to configure both transmitter and receiver, so you just configure both of them the same way.

mcgurk:
Receiver cannot be configured to require CRC. It is true, that it is more likely CRC is enabled. Indication if CRC is used would be still nice feature, if packets come from many different sources.

jgromes:
If you need this feature for debugging purposes, you can enable godmode and access the register directly via `lora._mod->SPIgetRegValue(SX127X_REG_HOP_CHANNEL, 6, 6)`, so the public `crcOnPayload()` wrapper becomes redundant.

Other than that, I think it's the user's responsibility to make sure the network is consistent. I simply don't see the need for this, sorry. It could be argued that a similar methods could be added for all the other configurations - and there's a lot of them, so I'd rather avoid it.

mcgurk:
Thanks for the God mode -hint. Great consept. Problem solved:
```
#define RADIOLIB_GODMODE
#define CRCONPAYLOAD() _mod->SPIgetRegValue(SX127X_REG_HOP_CHANNEL, 6, 6)
#include <RadioLib.h>
```
Now I can use `if(lora.CRCONPAYLOAD()) ...`.

jgromes:
Glad you have it working.

If there's future interest by other people to add this feature, I will consider it.

amirna2:
@jgromes Hello,
I know this PR is closed, but I am running into a situation where corrupt data is still passing through RadioLib SX127x implementation and into my Arduino application.
The use case is the following:
One SX126x transmitter (sends 59 bytes of data periodically)
One SX127x receiver (an Arduino Zero compatible board) receives the data.
Both radios are set to enable CRC

In 99.5% of the cases there are no problems. Packets are sent and arrive as they were sent.
In about half the cases, the same messages are rejected by readData with a CRC error (i.e error code -7).
But the other half is let through, resulting in corrupt packages being processed.

I noticed that if I check for `_mod->SPIgetRegValue(SX127X_REG_HOP_CHANNEL, 6, 6)`, the value return is 64 on the good packets, but 0 on the bad ones.
So I have a workaround for `readRadio()` to return a CRC error if the return value of SX127X_REG_HOP_CHANNEL is 0.
I think it would be useful to have that check.

What do you think?
Amir.

jgromes:
I think what's happening is that the packet header gets corrupted - that's why the bit in SX127X_REG_HOP_CHANNEL is not set even though it should be (since the packet does contain a CRC).

Checking just for SX127X_REG_HOP_CHANNEL == 0 isn't enough though, as it will always return CRC error when CRC is disabled (and possibly also when implicit header gets implemented). However, we could only check this when CRC is supposed to be enabled. If that check fails, then there's clearly come discrepency between the expected configuration and what was actually received.

Perhaps it would be best to implement the same check for the other information in the header (coding rate and length) and discard the packets that don't match.

amirna2:
You are right the CRC on payload only makes sense if the CRC is enabled in the first place.
And I definitely did run into issues when mixing up both type of Radio when the other parameters were not the same. Specifically CR and SyncWord must be the same. In fact the radio won't talk to each other if these were different.

This is my workaround for now and it only works because we explicitly enable CRC when setting up the radio.
The same setup is used for all radios as this is done through the ClusterDuckProtocol library I am working on.

```
    // check integrity CRC
    int16_t regValue = _mod->SPIgetRegValue(SX127X_REG_IRQ_FLAGS, 5, 5);
    int16_t crcOnPayload = _mod->SPIgetRegValue(SX127X_REG_HOP_CHANNEL, 6, 6);

    if(regValue == SX127X_CLEAR_IRQ_FLAG_PAYLOAD_CRC_ERROR || crcOnPayload == 0) {
      // clear interrupt flags
      clearIRQFlags();
      return(ERR_CRC_MISMATCH);
    }
```

PR #113, opened 2020-02-16
Carrier sense as sync word qualifier

Guglio95:
Hi @jgromes 
I'm sending you some tested changes:

-  Had to increase tolerance in `setRxBandwidth()` since some bandwidth values (coming from datasheet) were not accepted. It should be ok but, please, check it out.
   **Example**:   `[(26*1000000) / (8*( 4 + 0 ) * 2 ^ 2 )]  -  203 * 1000 = 125`
     _man = 0, exp = 2  corresponds to 203 Khz (datasheet page 35)._
     The above diff is 125 so that value won't be accepted with 0.1 tolerance.
- Some minor fixes. (imho)
- Added "carrier sense" as sync word qualifier; defaults to false to achieve back compatibility.

Thanks, 
Let me know.

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -357,14 +357,14 @@ int16_t CC1101::setRxBandwidth(float rxBw) {
   for(int8_t e = 3; e >= 0; e--) {
     for(int8_t m = 3; m >= 0; m --) {
       float point = (CC1101_CRYSTAL_FREQ * 1000000.0)/(8 * (m + 4) * ((uint32_t)1 << e));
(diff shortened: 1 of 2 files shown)


Guglio95:
I'm using the CC1101 with packets longer than the queue (64 bytes). 
I've patched `startTransmit()` and `readData()` methods to send/read packets longer than the queue; 
Will they be useful to RadioLib? Do you mind if I add them to this PR?

jgromes:
Thanks, all of this looks good.

Regarding the new handling for longer packets - I'll accept it if it doesn't break compatibility with PhysicalLayer. Perhaps better to open a new PR for those.

PR #117, opened 2020-02-24
[SX127x] Set OOK parameter before setting the bitrate to avoid reading undefined variable

4m1g0:
On SX127x when calling beginFSK, the [setOOK](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX127x/SX127x.cpp#L107) method is called after [setBitRate](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX127x/SX127x.cpp#L78) but inside setBitRate the _ook variable is used (before it is set): https://github.com/jgromes/RadioLib/blob/master/src/modules/SX127x/SX127x.cpp#L660

This results in ERR_INVALID_BIT_RATE deppending on what that section of memory has at the time it is called.

Calling setOOK before setBitRate solve the problem and I think it does not cause any other issue.

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -74,6 +74,10 @@ int16_t SX127x::beginFSK(uint8_t chipVersion, float br, float freqDev, float rxB
     RADIOLIB_ASSERT(state);
   }
 
+  // enable/disable OOK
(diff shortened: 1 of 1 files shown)


jgromes:
Looks good, thanks! I'll release this as 3.3.1 shortly.

PR #119, opened 2020-02-26
nRF24 small improvements.

Guglio95:
Hi Jan, 

I'm using your nRF24 driver to scan over frequencies, I made some small changes:

- `isCarrierDetected()`, to detect whatever the carrier is detected.
- I'm attaching notes on modified lines.

Please, let me know.

### DIFF
--- a/src/modules/nRF24/nRF24.cpp
+++ b/src/modules/nRF24/nRF24.cpp
@@ -230,14 +230,9 @@ int16_t nRF24::setFrequency(int16_t freq) {
     return(ERR_INVALID_FREQUENCY);
   }
 
-  // set mode to standby
-  int16_t state = standby();
-  RADIOLIB_ASSERT(state);
-
   // set frequency
   uint8_t freqRaw = freq - 2400;
-  state = _mod->SPIsetRegValue(NRF24_REG_RF_CH, freqRaw, 6, 0);
-  return(state);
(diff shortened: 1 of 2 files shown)


src/modules/nRF24/nRF24.cpp
```diff
@@ -230,14 +230,9 @@ int16_t nRF24::setFrequency(int16_t freq) {
     return(ERR_INVALID_FREQUENCY);
   }
 
-  // set mode to standby
-  int16_t state = standby();
```
Guglio95: Seems that there's no need to put modem in standby to change frequency; 
I've tested it and double checked on other nRF24 drivers.

src/modules/nRF24/nRF24.cpp
```diff
@@ -438,6 +438,13 @@ size_t nRF24::getPacketLength(bool update) {
 }
 
 int16_t nRF24::setCrcFiltering(bool crcOn) {
+  // Auto Ack needs to be disabled in order to disable CRC.
```
Guglio95: Enable CRC is forced high if one of the bits in EN_AA is high (datasheet, page 54)

jgromes:
Looks like there's an ambiguity in `setAutoAck(false)` on Arduino Due: https://travis-ci.org/jgromes/RadioLib/jobs/655399163

jgromes:
Checks out now, thanks! I didn't want to change it in master since that would cause a conflict.

PR #141, opened 2020-05-01
Set the min/max ranges per the Semtech SX1278/SX1276 datasheet

geeksville:
Hi.  I love ya'll's lib ;-).  I just switched our [www.meshtastic.org](www.meshtastic.org) project over from our fairly diverged originally RadioHead drivers to your nice framework.

I'm sending in a few PRs you might want.  This one is because I noticed that the high frequency limit for the RFM95 was too tight for legal US frequency bands.  So I changed the limits based on the Sematech datasheet values for both the RFM95 and RFM96.


jgromes:
Thanks, it's always nice to see the library being used in projects :)

The original values are from HopeRF's version of the SX1278 datasheet - though it is quite confused in places and often contradicts itself, so I wouldn't be surprised if this was one of the mistakes.

PR #142, opened 2020-05-01
when probing for sx127xs don't wait so long between attempts.

geeksville:
Hi,

Is there a reason the delay between probe attempts was so long?  I can't find anything in the datasheet that would suggest this requirement.  So I've changed it to 10ms per probe, so that if the device is missing we only stall the boot for about 100ms.


jgromes:
There was no specific reason for the 1s wait - in retrospect that is quite a lot, seeing as there's 10 attempts in total.

I'll change it in other similar drivers as well.

PR #143, opened 2020-05-01
Don't teardown interfaces (SPI) that are possibly shared 

geeksville:
In my application I have multiple devices on a single SPI bus (which are not radios), but Module.term() tears down the SPI controller.  Also, it tears down hardware serial.

Is there a reason for not just leaving the SPI controller alone when we fail to find our chip?

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -14,7 +14,7 @@ int16_t SX127x::begin(uint8_t chipVersion, uint8_t syncWord, uint8_t currentLimi
   // try to find the SX127x chip
   if(!SX127x::findChip(chipVersion)) {
     RADIOLIB_DEBUG_PRINTLN(F("No SX127x found!"));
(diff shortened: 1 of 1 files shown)


jgromes:
It didn't occur to me at the time, but you're right - there's no reason the radio module would be the only thing on the SPI bus (though that seems to be the most common configuration).

It would probably be better to change this in `Module::term()`, similarly to how `_initInterface` flag is checked in `Module::init()` - the assumption is that if it's the library that initialized the SPI bus, then it should also be the library to de-initialize it. So if `_initInterface` is set, then we should `end()` the SPI bus controller.

jgromes:
@geeksville I slightly modified `Module::term` method so that now it won't end interfaces that were not created by the library - this should resolve your issue, as long as you use one of the `Module` constructors that contain argument for `SPIClass`.

geeksville:
Sounds good.  Keep rocking it!

(Sent from a phone - please ignore typos)

PR #144, opened 2020-05-01
make Radiolib work on NRF52 arduino.  Two changes:

geeksville:
Hi,

I was getting build errors on my Adafruit Arduino NRF52 based project because SoftwareSerial is not defined on that platform (and no SoftwareSerial.h).  So I marked it as RADIOLIB_SOFTWARE_SERIAL_UNSUPPORTED.  Also, their default build doesn't have a hardware serial port either.  So I changed RADIOLIB_HARDWARE_SERIAL_PORT so that platforms could leave it as NULL in the constructor call to Module.

With this change my project now builds and runs on ESP32 and NRF52.

---
* Adafruit NRF52 arduino doesn't have a software serial port.

* Make RADIOLIB_HARDWARE_SERIAL_PORT optional and if not set assume NULL.
Note: this changes the type to a ptr from an object.

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -52,7 +52,7 @@
   #define RADIOLIB_INTERRUPT_STATUS                   RADIOLIB_PIN_STATUS
   #define RADIOLIB_NC                                 (0xFF)
   #define RADIOLIB_SOFTWARE_SERIAL_UNSUPPORTED
-  #define RADIOLIB_HARDWARE_SERIAL_PORT               Serial1
+  #define RADIOLIB_HARDWARE_SERIAL_PORT               &Serial1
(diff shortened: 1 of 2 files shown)


jgromes:
That's odd - it builds fine for me using Adafruit Feather nRF52832. Which board are you using?

Also, what was the issue on ESP32?

geeksville:
nrf52840-dk.  When back at my computer later I'll check if perhaps they
have some difference there.

jgromes:
Closing as I can't replicate the reported issues using Adafruit nRF52 or ESP32 cores (and because the recent changes to BuildOpt caused a bit of conflict).

Feel free to open an issue/PR if there's a way to replicate this.

PR #150, opened 2020-05-31
[SX127x] add getTempRaw method

gozu42:
this is pretty much just a RadioLib-ed version of the example code from the datasheet page 126++.

since this is quite the ritual dance with a bunch of register read/writes, it seemed better to include
it RL side instead of going with thin accessor functions like modemStatus.

SX127X_TEMP_MONITOR_ON/OFF logic seems to have been inverted, see datasheet page 89.
since these defines were unused within RL afaict i didnt make that part a separate commit.
this might bite anyone who does use the defines through godmode.

partially for that reason the code is more verbose then it really needs to be when dealing
with the bit-operations, to avoid having "assumptions about the defines coded in".

i considered adding more to the description in the .h, but was not sure how to properly phrase it.
possibly worth adding would be:
- this changing the opmode (so dont call it while actively rx/tx)
- it restoring the modem to original opmode
- blocking for 200us

200us is more than the example code uses, but depending on where in the datasheet you look, they use
anything from 140-160us, and 200us is still "very short" and "works well here".
this could probably be made configurable or a parameter.

adding an "async mode" would also be possible, but would require a way to "stash" the old OpMode
and doesnt seem worth it considering 200us is ... still very short.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -126,6 +126,7 @@ setRSSIConfig	KEYWORD2
 setEncoding	KEYWORD2
 getIRQFlags	KEYWORD2
 getModemStatus	KEYWORD2
+getTempRaw	KEYWORD2
 
 # RF69-specific
 setAESKey	KEYWORD2
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -934,6 +934,58 @@ uint8_t SX127x::getModemStatus() {
   return(_mod->SPIreadRegister(SX127X_REG_MODEM_STAT));
 }
 
+int8_t SX127x::getTempRaw() {
+  int8_t temp = 0;
+  uint8_t previousOpMode;
+  uint8_t ival;
+
+  // save current Op Mode
+  previousOpMode = _mod->SPIgetRegValue(SX127X_REG_OP_MODE);
+
+  // check if we need to step out of LoRa mode first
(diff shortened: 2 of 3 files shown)


jgromes:
> partially for that reason the code is more verbose then it really needs to be when dealing
with the bit-operations, to avoid having "assumptions about the defines coded in".

To be honest, I'd prefer it to just set the bit using `SPIsetRegValue()` with the bit positions specified - the swapped values for SX127X_TEMP_MONITOR_ON/OFF might well be a typo on my end. Besides, godmode is by no means meant to be relied on to bypass the public API. If there's a feature missing from it, I would encourage people to open an issue or PR rather than relying on godmode to patch it up.

200us waiting is no big deal, definitely not worth adding async methods for it.

gozu42:
> To be honest, I'd prefer it to just set the bit using `SPIsetRegValue()` with the bit positions specified - 

i was tied about that since i dont like giving the bit positions as "raw numbers" in the SPIsetRegValue() call, but i didnt want to add something like SX127X_TEMP_MONITOR_ONOFF_MSB and _LSB defines either.
your code, your call, changed.


> 200us waiting is no big deal, definitely not worth adding async methods for it.

after reading the implementation of SPIsetRegValue() it is more obvious how 200us of wait really doesnt matter if it is in the middle of 4-6 OP_MODE writes.

jgromes:
Merged, thanks!

I'll extend the doxygen comments, but that's just a bit of documentation ;)

PR #163, opened 2020-07-14
Tweaks and fixes for CC1101 and nRF24

phretor:
Sorry, these have been sitting on our working copy for a while.


src/modules/CC1101/CC1101.cpp
```diff
@@ -351,7 +414,12 @@ int16_t CC1101::setBitRate(float br) {
   // set bit rate value
   int16_t state = SPIsetRegValue(CC1101_REG_MDMCFG4, e, 3, 0);
   state |= SPIsetRegValue(CC1101_REG_MDMCFG3, m);
-  return(state);
+
+  if (state == ERR_NONE) {
+    _br = br;
```
jgromes, line 419: CI is complaining `_br` doesn't exist - I guess it's not declared in the header.

src/modules/CC1101/CC1101.h
```diff
@@ -598,7 +599,7 @@ class CC1101: public PhysicalLayer {
 
       \param func ISR to call.
 
-      \param dir Signal change direction. Defaults to FALLING.
+      \param dir Signal change direction. Defaults to RISING.
```
jgromes, line 602: The `dir` parameter still defaults to `FALLING`

src/modules/CC1101/CC1101.cpp
```diff
@@ -228,27 +223,56 @@ int16_t CC1101::startTransmit(uint8_t* data, size_t len, uint8_t addr) {
...
   RADIOLIB_ASSERT(state);
 
+  // data put on FIFO.
+  uint8_t dataSent = 0;
+
   // optionally write packet length
   if (_packetLengthConfig == CC1101_LENGTH_CONFIG_VARIABLE) {
+
+    // enforce variable len limit.
+    if (len > 254) {
```
jgromes, line 233: The magic number 254 should probably be replaced with CC1101_MAX_PACKET_LENGTH - 1.

jgromes:
I just noticed that the while loop that was added in `CC1101::startTransmit()` will turn it into blocking method - it will wait until FIFO is sufficiently empty. Because CC1101 can transmit at very low speeds (down to 25 bps), I'd say that's a major issue.

I think it would be better to keep `startTransmit` limited to the internal FIFO size, and keep all blocking functionality (including feeding FIFO "on the fly") to the `transmit` method, which is meant to be blocking.

jgromes:
@phretor @Guglio95 sorry guys, I can't merge this as-is. I don't think it's a good idea to make `startTransmit()` essentially into a blocking method (because it has to wait for an external component to finish).

rnbokade:
Is this going to be opened and worked upon soon? Looking forward to using longer packets on RFM69HCW. If anything I'll try myself some modifications but not sure will be able to produce good quality code though.

jgromes:
As per my previous comment, I don't think it's a good idea to make `startTransmit` wait for some part of the packet to get transmitted, since it could cause confusion. At the moment, `startTransmit` will fill the FIFO and start transmitting the packet, but then it will return immediately, so the whole time the module is transmitting it's under user control (and not the library). Merging this as is would change that behavior, since `startTransmit` would have to fill up FIFO, start transmitting and then wait until the FIFO is sufficiently empty to fill in the rest of the packet. The problem is that this is what `transmit` does - it takes care of the entire operation and only returns once packet is transmitted.

IMO the best solution is to enforce FIFO size limit when using `startTransmit`, (possibly provide user method(s) to fill in FIFO on his own) and implement longer packets within `transmit`.

Unforunately I don't have the time to work on this at the moment, but I will accept PR that implements this.

rnbokade:
Can we not do spi transactions within interrupt? I think there is something like SPIusingInterrupt(interrupt number) function in spi library which may help transferring during interrupt. So the FIFO reloading function can be initiates when the interrupt for fifo_threshold is triggered. I don't have a clear idea about it but may work. Or maybe there is also some atomic_macro structure that may help although I am not sure.
Having different packet length capacities on different transmit functions might confuse users.

jgromes:
The problem isn't the SPI transaction, but rather that inside `startTransmit` we would have to wait until module transmits part of the packet. Packets larger than FIFO are handled by filling the FIFO while the module is transmitting.

The idea behind `startTransmit` was to have a quick function that will prepare everything for transmission but doesn't wait until it's finished since that can take a really long time.

I agree that different maximum packet lengths for different transmit functions would be confusing, but that can be avoided by providing public methods to fill in FIFO during transmission. In other words, it would be the user code, not the `startTransmit` method that's doing the FIFO filling. Then `startTransmit` wouldn't have to wait for anything.

rnbokade:
I have a rough idea on how to get the transmit on the fly working without blocking. Idk how correct i am on this but please let me know.
startTransmit() => 
1. check if less than max buffer size of the given module, if true, use direct logic,else do the second step onwards
2.  print (packet size is long so using on the fly method)
3. take a volatile variable with the scope that can be accessed globally and make its value = length to be transmitted let's say "translen" .
4. set the interrupt pin to interrupt on threshold. set this threshold to max_buffer_length/2
5. Attach this pin to a funtion say "OnFlyITransISR" (need to create this new method)
6. Add Max_buffer_length equivalent of bytes to the FIFO and then update the value of  tanslen =translen- Max_buffer_length
7. Start transmission.
OnFlyTransISR() =>
1. check if (translen<Max_buffer_length/2). 
2. If it is less then, add the complete remaining string to fifo, set translen to 0 and detach the interrupt. Put intterupt back to transmit complete mode and do the normal complete string printing if required.
3. If it is not load max_buffer_length/2 next bytes and update the translen variable and leave everything as it is.

jgromes:
> 5. Attach this pin to a funtion say "OnFlyITransISR" (need to create this new method)

You can't attach class member method to an interrupt - the function you're attaching has to be typed as `void (*)()` while the OnFlyITransISR would be something like `void CC1101::(*)()`. The ISR would have to be a plain function in the global scope. However, then it would have no way to tell which instance of the CC1101 class is transmitting - RadioLib is object-oriented, so there very well can be multiple different instances of a module running, or even being created and destroyed during runtime. Though I agree to usual use-case is a single module created on startup and never destroyed.

The only way to do this is similar to how all the *_Interrupt examples are handled, which is to have the ISR in the user sketch (i.e. in global scope) and feed the FIFO from there.

PR #167, opened 2020-07-17
[SX127x] Dio1Action precondition checks were inverted

halfbakery:
The precondition in the clear/set Dio1Action functions were inverted, it rendered these functions useless. :-)


jgromes:
Copy-paste bugs strike again - this is most likely in other modules optional interrupt pins too, I'll fix those as well.

Merged, thanks!

naggie:
Thanks @halfbakery !

PR #239, opened 2021-02-02
New platform support for SparkFun LoRa Thing Plus - expLoRaBLE

lewispg228:
Dear Jan Gromes,
Thank you so much for your work on this RadioLib Arduino Library. It has been wonderful developing with a new project using the apollo3 and sx1262! We are about to release our new product, and we will most definitely be directing users to your library in our hookup guide. In an effort to make using your library even easier with our new hardware, we have made the following changes. 

Please let us know what you think and thanks for considering our pull request!
Pete Lewis
SparkFun Electronics

Added platform support for new SparkFun Hardware, the SparkFun LoRa Thing Plus - expLoRaBLE. This module uses the apollo3 and sx1262.

-Note, also adjusted the SX126x.cpp file. I added a check for apollo3 archtecture to ensure the SPITransfer works more consistently using the 1 ms delay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ SX127x, RFM9x, SX126x, RF69, SX1231, CC1101, nRF24L01, RFM2x, Si443x and SX128x
   * [__Curie__](https://github.com/arduino/ArduinoCore-arc32) - Arduino 101
 
 * __SparkFun__
-  * [__Apollo3__](https://github.com/sparkfun/Arduino_Apollo3) - Sparkfun Artemis Redboard
+  * [__Apollo3__](https://github.com/sparkfun/Arduino_Apollo3) - Sparkfun Artemis Redboard, SparkFun LoRa Thing Plus - expLoRaBLE
 
 * __ST Microelectronics__
   * [__STM32__ (official core)](https://github.com/stm32duino/Arduino_Core_STM32) - STM32 Nucleo, Discovery, Maple, BluePill, BlackPill etc.
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -312,6 +312,43 @@
     #define RADIOLIB_PROGMEM                            PROGMEM
     #define RADIOLIB_PROGMEM_READ_BYTE(addr)            pgm_read_byte(addr)
 
+  #elif defined(ARDUINO_LoRa_THING_PLUS_expLoRaBLE)
+    // SparkFun LoRa Thing Plus - expLoRaBLE
+    #define RADIOLIB_PLATFORM                           "SparkFun LoRa Thing Plus expLoRaBLE"
+    #define RADIOLIB_PIN_TYPE                           pin_size_t
+    #define RADIOLIB_PIN_MODE                           Arduino_PinMode
+    #define RADIOLIB_PIN_STATUS                         PinStatus
+    #define RADIOLIB_INTERRUPT_STATUS                   RADIOLIB_PIN_STATUS
+    #define RADIOLIB_DIGITAL_PIN_TO_INTERRUPT(p)        digitalPinToInterrupt(p)
+    #define RADIOLIB_NC                                 (0xFF)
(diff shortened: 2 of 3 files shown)


lewispg228:
FYI, I have verified that this works with the following setup:

Arduino IDE 1.8.13
SparkFun Boards package for apollo3 version 2.0.5 (https://github.com/sparkfun/Arduino_Apollo3)

Note, I verified using two boards and successfully transmitted and received "hello world". Using the transmit and receive examples with the following changes:

    SX1262 radio = new Module(D36, D40, D44, D39);

    int state = radio.begin(915.0, 250.0, 7, 5, 0x34, 20, 10, 0, false);

These were the only modifications necessary to get them to work with the expLoRaBLE.

Cheers!

jgromes:
Thanks for the PR, but I'm not entirely sure it's necessary to add the expLoRaBLE board as a platform.

"Platform" in the context of this library means basically "MCU architecture", and since expLoRaBLE is using Apollo3, it should already be supported as-is, right?

danJTDC:
Hi @jgromes, thank you for your quick reply. I've been working closely with Pete on the expLoRaBLE project. I understand what you mean by Apollo3 is already a part of RadioLib, however, I believe it is not updated to support the latest Apollo 3 Arduino core ( v2.x.x based on Mbed OS). 
If we submit a PR updating the Apollo3 platform, would it be possible to include the expLoRaBLE as a subcategory? The only major difference would be that the expLoRaBLE uses SPI1 for LoRa by default since it is built into the chip itself.

jgromes:
@danJTDC Interesting - the Apollo3 support has been added quite a while ago. I don't usually actively monitor changes in supported platforms, because any backwards-incompatible changes become apparent in failing CI jobs when new RadioLib version is released.

If there were major changes in Apollo3 Arduino core that are being used in RadioLib build options (such as datatypes), then it's probably better to update the Apollo3 support anyway, and while we're at it, change default SPI to SPI1.

On a side note, I wasn't able to find any information on expLoRaBLE on Sparkfun website - is that some new, as of yet unreelased project?

santaimpersonator:
@jgromes Yes, the new board [https://www.sparkfun.com/products/17506] was just released this afternoon.

jgromes:
@santaimpersonator thanks for the info, it looks pretty cool!

@danJTDC @lewispg228 I had a look at the 2.x Apollo3 Arduino core, I think it's best to update RadioLib platform to use that instead of creating a separate platform for expLoRaBLE (since that would also require adding the board to CI testing matrix).

To use expLoRaBLE, you'll just need to add `SPI1` as an argument to the `Module` constructor:

```
SX1262 radio = new Module(D36, D40, D44, D39, SPI1);
```

lewispg228:
Hi @jgromes ,
Thanks @santaimpersonator and @danJTDC  for chiming in here! 

Jan, your idea does seem like the best way to go. However, I was just giving it a try and came across this compile error. It seems that it can't find a matching constructor function. 

I tried adding in the SPISettings like so, and it then compiled okay.

```
SPISettings spiSettings = SPISettings(2000000, MSBFIRST, SPI_MODE0);
SX1262 radio = new Module(D36, D40, D44, D39, SPI1, spiSettings); 
```



Any ideas why this is not finding the constructor with spiSettings omitted? I would think that it is not absolutely necessary, and if you omitted it, then it would just use the defaults as defined [here](https://github.com/jgromes/RadioLib/blob/master/src/Module.h#L113).

Let us know and thanks again!
Pete

image

jgromes:
The constructor there is one of the generic constructor intended for RadioShield - the constructor that was actually being used was [this one](https://github.com/jgromes/RadioLib/blob/59c44d388371869d9b875e76e38ee6ee57293758/src/Module.h#L91). Not sure why it didn't specify default SPISettings - though looking at it, other Module contructors don't specify it either, I might have to think about this.

Anyway it's fixed now, so I think we can close this PR. Let me know if there are any issues with the platform, as I can't really test it.

PR #243, opened 2021-02-04
Added GetRSSIInst command for SX126x

angrypig7:
Hi,

While trying to implement the LBT(Listen-Before-Talk) feature on SX1262 chips, I've added the GetRSSIInst command (Opcode=0x15) for SX126x chips.
Since this command only works during the reception, in order to implement LBT for LoRa transmission the user will have to manually switch to FSK, set to receive mode, run the GetRSSIInst command, and switch back to LoRa.
I've tested it with STM32F103 MCU, but I expect other platforms to have no problem.

### DIFF
--- a/src/modules/SX126x/SX126x.h
+++ b/src/modules/SX126x/SX126x.h
@@ -807,9 +807,9 @@ class SX126x: public PhysicalLayer {
    uint32_t getTimeOnAir(size_t len);
 
    /*!
-     \brief Get instantaneous RSSI value in FSK mode. Can be used for LBT implementation, checking for interference before transmitting.
+     \brief Get instantaneous RSSI value during recption of the packet. Should switch to FSK receive mode for LBT implementation.
 
-     \returns Instantaneous RSSI value in (-2)*dBm
+     \returns Instantaneous RSSI value in (-2)*dBm.
    */
    uint8_t getRSSIInst();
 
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks for the PR!

Maybe it would be better to return the value in dBm as to have an actual physical quantity, rather than an arbitrary -2*dBm "unit".

src/modules/SX126x/SX126x.h
```diff
@@ -806,6 +806,13 @@ class SX126x: public PhysicalLayer {
    */
    uint32_t getTimeOnAir(size_t len);
 
+   /*!
+     \brief Get instantaneous RSSI value during recption of the packet. Should switch to FSK receive mode for LBT implementation.
+
+     \returns Instantaneous RSSI value in (-2)*dBm.
+   */
+   uint8_t getRSSIInst();
```
jgromes, line 814: This should be `float` too ;)

angrypig7: Yeah, I got mails from Github saying some checks have failed and was wondering if editing directly to a PR was the right way to do it.
I should get more familiar with Github :)

angrypig7:
Yeah, I was hoping to minimize the resource returning the value in uint8_t type, but processing the *unit* would be better I guess.
The units are now in dBm.

jgromes:
Merged, thanks!

BTW, to implement LBT for LoRa, RSSI might not be enough, since LoRa signal can have negative SNR (i.e. it's below the noise floor) and still be received. CAD (Channel Activity Detection) is more useful for this.

angrypig7:
In regions like South Korea, the LBT feature based on center frequency is mandatory for radio certification. I believe Japan is too.
Thanks for noting me, I guess I'll use CAD too :)

PR #261, opened 2021-03-03
[SX1278] Added method to control InvertIQ

marcinczeczko:
Hi,
I came across the issue of missing methods to control Invert IQ signals on SX1278.  The register address is already present in SX127x.h module (0x33). Here is the implementation of `invertIQ` method for SX1278

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -131,6 +131,7 @@ setRfSwitchPins	KEYWORD2
 forceLDRO	KEYWORD2
 autoLDRO	KEYWORD2
 getChipVersion	KEYWORD2
+invertIQ    KEYWORD2
 
 # RF69-specific
 setAESKey	KEYWORD2
--- a/src/modules/SX127x/SX1278.cpp
+++ b/src/modules/SX127x/SX1278.cpp
@@ -538,6 +538,22 @@ int16_t SX1278::setHeaderType(uint8_t headerType, size_t len) {
   return(state);
 }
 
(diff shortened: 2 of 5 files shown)


jgromes:
Thanks, is the register address and its values the same on SX1278 and SX1272? If so, then this should be moved into the base SX127x class.

marcinczeczko:
@jgromes 1272/73 have the same register but have two bits to separately control IQ on RX and TX paths, while 1276-79 have only one bit as I implemented.
Should I add corresponding methods to 1272-73 ? It would be `invertIQRX` and `invertIQTX`

jgromes:
I just checked my version of SX1276/77/78/79 datasheet (Rev. 7 May 2020) and that one also has invert IQ for Rx and Tx separately. 
Also, both have a register InvertIQ2 which apparently should be modified if inverted IQ is used.

marcinczeczko:
Ahh, I didn't notice I read datashet from March 2015, while for 1272 I just downloaded the latest one. That's why I see this difference. Ok, I'll update it accordingly - meaning:
- On invertIQ(true), both InvertIQ RX & TX = 1, and InvertIQ2 to 0x19
- On invertIQ(false), both InvertIQ RX & TX = 0, and InvertIQ2 to 0x1D

Thanks for pointing it out.

marcinczeczko:
@jgromes I made changes as mentioned in the previous comment

jgromes:
Looking good, thanks for the contribution!

PR #282, opened 2021-04-15
Feature/implement afc control

yeckel:
With this patch, a AFC works pretty well and could be fully user-controlled. There is however a **glitch** calling ```SX127x::receiveDirect()``` ends always with an error -16. Setting SX127X_RX(0b00000101) *doesn't* work.

It has to be set to ```0b00000101``` in order to start RX, but it's not stored even after a maximal timeout! When setting just ```0b00000100``` (RSRx) the receiver is (obviously) not started but the SPI Write succeeds.  

I would suggest returning always true from ```int16_t SX127x::setMode(uint8_t mode)``` but it's an API change and not sure of possible implications.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -104,10 +104,14 @@ setPreambleLength	KEYWORD2
 setGain	KEYWORD2
 getFrequencyError	KEYWORD2
 getRSSI	KEYWORD2
+getAFCError	KEYWORD2
 getSNR	KEYWORD2
 getDataRate	KEYWORD2
 setBitRate	KEYWORD2
 setRxBandwidth	KEYWORD2
+setAFCBandwidth	KEYWORD2
+setAFC		KEYWORD2
+setAFCAGCTrigger	KEYWORD2
 setFrequencyDeviation	KEYWORD2
 setNodeAddress	KEYWORD2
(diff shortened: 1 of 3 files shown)


jgromes (commented):
Thanks, I left a couple comments. Also, could you add the new setAFC methods to keywords.txt? Please use true tab (not spaces) in that file, thanks.

The code scan error can be ignored, it's fixed it in the meantime.

jgromes:
Regarding the issue with -16  error - I don't think it's a good idea to always return OK error code from `setMode()`, because this "stuck bit" always happens in FSK mode (it seems to be working with LoRa) and only for some modem settings. I'll try to figure out how to only ignore that bit while still checking all others.

jgromes:
Merged, thanks! What frequency offset values are you getting with AFC enabled?

yeckel:
Do you mean with boards I have? Up to +/- 3.5kHz, I've added an option into gui to save it as a config option. So an easy calibration could be done. After that calibration the offset is about 100Hz.

core-c:
I can confirm AFC works well.
I'm dealing with about the same frequency offsets (+/- 3.5 KHz).

PR #290, opened 2021-04-24
Add assign a class method as an interrupt handler

sergimn:
The intention is to be able to execute code inside of a class with access to the instance data when a packet is received.

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -2,6 +2,9 @@
 #define _RADIOLIB_MODULE_H
 
 #include "TypeDef.h"
+#include <functional>
+#include <FunctionalInterrupt.h>
+
 
 #include <SPI.h>
 #ifndef RADIOLIB_SOFTWARE_SERIAL_UNSUPPORTED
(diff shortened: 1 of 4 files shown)


src/Module.h
```diff
@@ -2,6 +2,9 @@
 #define _RADIOLIB_MODULE_H
 
 #include "TypeDef.h"
+#include <functional>
+#include <FunctionalInterrupt.h>
```
sergimn, line 6: I'm not sure why this is needed but the signature of `attachInterrupt` without this is `void attachInterrupt(uint8_t, void (*)(), int)`. Is it supposed to be like that or is it an error of mine of not having the appropriate compiling includedirs?

sergimn:
If you want to try the current implemtation, try with the following line:
 `  radio->setDio0Action(std::bind(&MyClass::onReceive, this));`

It works as expected

src/Module.h
jgromes: `<functional>` is not supported on AVR Arduino - see this failing run: https://github.com/jgromes/RadioLib/pull/290/checks?check_run_id=2428779662

Looking into AVR Arduino `attachInterrupt` implementation, it is hard-coded to only accept ISRs without `void*` arguments. So this will never work on AVR Arduinos like Uno, which are unfortuantely still the most common ones.

jgromes:
Closed due to incompatibility with some RadioLib platforms (at least AVR).

Issue #89, opened 2019-12-12
SX1262: Cannot Calibrate image or transmit


naggie:
Hi,

I've been trying to successfully transmit a packet using the [SX126x_Transmit](https://github.com/jgromes/RadioLib/blob/master/examples/SX126x/SX126x_Transmit/SX126x_Transmit.ino) example. I have BUSY and DIO1 connected. I've tried both v2.0.1 and master (e61bd58).

The modem I'm using is this: https://www.rfsolutions.co.uk/radio-modules-c10/dip-lora-transceiver-module-20km-868mhz-p807 . It has a regular XTAL, not a TCXO so I don't need to set up DIO3 to setup a TCXO.

# Image calibration
However, whilst the [modem](https://www.rfsolutions.co.uk/radio-modules-c10/dip-lora-transceiver-module-20km-868mhz-p807) successfully initialises many of the parameters, it gets stuck on SetFrequency(868.0) as part of SX1262::Begin() -- specifically when trying to do an Image calibration -- the error status returned is ERR_SPI_CMD_TIMEOUT which corresponds to waiting for the BUSY line. I get the 5 second timeout before failing. I've checked the busy line, it appears to work as expected for the other SPI commands.

# Transmitting
By skipping image calibration (by setting the second arg to SetFrequency to false) I was able to attempt to transmit. When doing this, the code hits the same timeout error but looking at the code it's waiting for int0/DIO1 for the modem to say it's started transmitting.

I am out of ideas as to what the problem is. I've tried using a slower SPI speed, as well as checked connections many times over. I'd greatly appreciate any suggestions.

Thanks for your time.

jgromes:
Since your module doesn't have TCXO, did you set the TCXO reference voltage to 0 in the begin method? It's set to 1.6V by default, which will enable TCXO control. Upload the changed code, then unplug the power and wait for a few seconds to make sure everything is completely off. Then plug the power back in.

As far as I know, SX126x doesn't have any documented way to disable TCXO control once it's been enabled, due to the lack of documentation for SPI registers and the fact that the corresponding SPI command can only enable the control, not disable it. Reset or power cycling to make sure all registers have the default value seems to be the only way to do it.

naggie:
II literally just tried that before reading your reply after noticing the default of 1.6. --- it worked! Thank you so much!

Callan

PR #341, opened 2021-08-01
Modify (SX1262.h, SX1262.cpp): remove tcxoVoltage from begin, because…

RJSDevel:
remove tcxoVoltage from begin, because it isn't support in SX1262 and LLCC68, override setTCXO by the same cause

### DIFF
--- a/src/modules/SX126x/SX1262.cpp
+++ b/src/modules/SX126x/SX1262.cpp
@@ -5,9 +5,9 @@ SX1262::SX1262(Module* mod) : SX126x(mod) {
 
 }
 
-int16_t SX1262::begin(float freq, float bw, uint8_t sf, uint8_t cr, uint8_t syncWord, int8_t power, uint16_t preambleLength, float tcxoVoltage, bool useRegulatorLDO) {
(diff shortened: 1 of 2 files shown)


jgromes:
How did you arrive at the conclusion that TCXO isn't supported by SX1262 and LLCC68? Both SX1262 and LLCC68 datasheets say otherwise, see for example chapter 4.1.4 4 TCXO Control Block in LLCC68 datasheet.

Your particular LLCC68 module might be using XTAL instead of TCXO, but that's not a reason to drop TCXO support for every LLCC68 module.

RJSDevel:
https://github.com/jgromes/RadioLib/issues/89#issuecomment-565008040

RJSDevel:
If it supports TCXO , then add any information to docs that it sometimes doesn't support it, because I've spent about week to find that

jgromes:
> it sometimes doesn't support it

Unforunately it's a bit more complicated - all SX126x transceivers (LLCC68 included) support TCXO. It's up to whoever is building the module using the transceiver whether he decides to use TCXO or not. By default, RadioLib assumes TCXO to be used, because from what I've seen, modules with TCXO are more common. 

If I were to set the default to XTAL being used, the same issues would appear on TCXO-equipped modules. I can add even more information to the already quite extensive documentation, but ultimately it's up to the user to read the docs, and then configure RadioLib to match the hardware - same thing would apply if you were to for example select an incorrect SPI bus, not much I can do to stop you from doing that.

Issue #357, opened 2021-09-02
CC1101: RX Interrupt triggered when Transmitting


studiofuga:
**Describe the bug**
It seems that when transmitting a packet using `transmit()`, an IRQ is asserted.
If the receiver is used with IRQ, a spurious read state is triggered.

**To Reproduce**

The foollowing code is built on the `receiver_interrupt` example.

Just added a periodic transmission.

```
// include the library
#include <RadioLib.h>

// CC1101 has the following connections:
// CS pin:    10
// GDO0 pin:  2
// RST pin:   unused
// GDO2 pin:  3 (optional)
CC1101 radio = new Module(10,2, RADIOLIB_NC,3);

// flag to indicate that a packet was received
volatile bool receivedFlag = false;

// disable interrupt when it's not needed
volatile bool enableInterrupt = true;

unsigned long lastSend = 0;

bool enableSend = true;

// this function is called when a complete packet
// is received by the module
// IMPORTANT: this function MUST be 'void' type
//            and MUST NOT have any arguments!
void setFlag(void) {
    // check if the interrupt is enabled
    if(!enableInterrupt) {
        return;
    }

    // we got a packet, set the flag
    receivedFlag = true;
}

void setup() {
    Serial.begin(38400);

    // initialize CC1101 with default settings
    Serial.print(F("[CC1101] Initializing ... "));
    int state = radio.begin(868.3, 38.4, 21.0, 200.0, 10, 32);
    if (state == ERR_NONE) {
        Serial.println(F("success!"));
    } else {
        Serial.print(F("failed, code "));
        Serial.println(state);
        while (true);
    }

    // set the function that will be called
    // when new packet is received
    radio.setGdo0Action(setFlag);

    // start listening for packets
    Serial.print(F("[CC1101] Starting to listen ... "));
    state = radio.startReceive();
    if (state == ERR_NONE) {
        Serial.println(F("success!"));
    } else {
        Serial.print(F("failed, code "));
        Serial.println(state);
        while (true);
    }

    // if needed, 'listen' mode can be disabled by calling
    // any of the following methods:
    //
    // radio.standby()
    // radio.sleep()
    // radio.transmit();
    // radio.receive();
    // radio.readData();

    lastSend = millis();
    enableSend = true;
}

void loop() {
    unsigned long now = millis();

    if (enableSend && now - lastSend > 5000) {
        // Send something
            Serial.println(F("Sending..."));
[...]

PR #358, opened 2021-09-02
Fix issue #357.

studiofuga:
Fix issue #357 as described in the issue report.
The fix simply change the interrupt handler as long as the transmit is completed, then restoring the original handler before returning.


jgromes:
Resolved by separating GDOx interrupts (as described in #357).

Issue #399, opened 2021-11-03
[SX1278] Can't receive long packets in FSK mode


notsleep196:
Hello!

We are trying to use your lib to receive long packets (more than 64 bytes) in FSK mode . Yes, we know that SX127x have only 64 bytes of FIFO in FSK mode, but it can receive long packets with some special workflow (paragraph 2.1.13.5 in datasheet).
Unfortunately Radiolib does not support such workflow and packets more than 64 bytes in FSK for SX127x. We've found several incompatibilities described below.

Goal of this issue is to describe these incompatibilities and discuss possible solutions. Maybe we'll implement some changes at application side, other at Radiolib with your assistance. In worst scenario we'll fork Radiolib with our rough patches.

General scheme to work with long packets is to use FIFOLevel interrupt, and start reading FIFO until it's fully filled:
 - Start reading bytes from the FIFO when FifoEmpty is cleared or FifoThreshold becomes set.
 - Suspend reading from the FIFO if FifoEmpty fires before all bytes of the message have been read.
 - Continue to step 1 until PayloadReady or CrcOk fires
 - Read all remaining bytes from the FIFO either in Rx or Sleep/Standby mode
 - FIFOLevel interrupt is maped to DIO1 by default, so we can use setDio1Action() function to connect FIFOThreshold handler with this interrupt.

Example of FIFOThress handler:
```cpp
void IRAM_ATTR onFIFOThresh() {
    std::vector<uint8_t> dataVector;
    while ((radio->getIRQFlags() >> 14) & 0b01) == 0) {  //check FIFOEmpty irq flag
        uint8_t buffer[16];  //FIFO_THRESHOLD + 1 bytes
        int state = norbyTransceiver.readData(buffer, 16);
        if (state == ERR_NONE) {
            dataVector.insert(dataVector.end(), buffer, buffer + 16);
        }
    }
}
```
And it doesn't work properly now. We've found several issues here:

1. readData() method puts module to standby mode, so it can't receive anything after FIFOLevel interrupt was triggered on DIO1 and onFIFOThresh() irq handler called. So we can receive FIFO_THRESHOLD of bytes only.

2. Strange part with length variable for FSK mode. I think clearFIFO() function will never be called in FSK mode, because in string 542 in SX127x.cpp we can see
```cpp
length = getPacketLength();
```
and
```cpp
size_t packetLength = getPacketLength();
if(packetLength > length) {
    clearFIFO(packetLength - length);
}
```
[...]

Issue #201, opened 2020-11-12
Stream implementation


jgromes:
As per discussion in #163 with @phretor and @rnbokade, it could be possible to extend maximum packet length beyond the FIFO size limitation by providing user with methods to feed the FIFO "on-the-fly", essentially creating a data stream instead of packetized transfer. This could be implemented for any module with interrupts on FIFO level, currently:

* CC1101
* Si443x/RFM2x
* SX127x/RFM9x in FSK mode
* RF69

Transmitter (method names are placeholders):

1. In Arduino sketch, user creates an ISR, e.g.

```
void fifoAdd(void) {
  radio.fifoAdd(txBuffPtr, &remLength);
}
```

2. The ISR is passed as callback to e.g. `radio.setFifoEmptyAction(fifoAdd);`, much like `setFlag`/`setDio0Action`.
3. If the packet is longer than FIFO size, `startTransmit()` enables interrupt for some FIFO level. If not, interrupt is not enabled.
4. `startTransmit` begins transmission just like the current implementation and returns immediately after setting mode to Tx.
5. `fifoAdd` ISR now gets called every time FIFO level gets low enough, progressively filling FIFO with new bytes to be sent.
6. When `remLength` gets to zero, FIFO level interrupt is disabled and transmission is considered finished.

Receiver structure is the same, the only difference is that interrupt must be triggered on a different FIFO fill level (close to full), and the method called from ISR is extracting bytes from (e.g. `fifoGet(uint8_t* buffPtr, size_t* remLength)`).

rnbokade:
Thanks for looking into this. I will also be working side by side on implementation. BTW, I think RF69 will also be compatible with this approach, please confirm.

jgromes:
Finally got around to this, Stream is now implemented for SX127x. Have to say, watching the waterfall is quite funny when sending 512 byte packets using a 4800 bps transmission.

![Screenshot_27](https://user-images.githubusercontent.com/3407192/178151967-8865fcf7-0183-4ad4-9903-2104e17be321.png)

Will add support for the other modules mentioned above in the coming days.

PR #403, opened 2021-11-12
SX127X FSK Mode: transmit up to 256 bytes

justephens:
The datasheet describes an "on-the-fly" method of refilling the FIFO buffer when transmitting in FSK mode, which allows for FSK packets up to 256 bytes long. I made small tweaks to the `SX127X::transmit` method to handle this case.

I am not familiar `yield()`, but this implementation will capture control and just sit in a loop until all of a message has been pushed into the FIFO, which I suspect might not be fantastic.

### DIFF
--- a/examples/SX127x/SX127x_FSK_FIFO_Refill/SX127x_FSK_FIFO_Refill.ino
+++ b/examples/SX127x/SX127x_FSK_FIFO_Refill/SX127x_FSK_FIFO_Refill.ino
@@ -1,0 +1,83 @@
+/*
+   RadioLib SX127x FSK FIFO on-the-fly Refilling Example
+
+   This example shows how to use refill the FIFO buffer in SX127x chips, to allow for transmitting
+   FSK/OOK packets longer than 256 bytes.
+
+   For default module settings, see the wiki page
+   https://github.com/jgromes/RadioLib/wiki/Default-configuration#sx127xrfm9x---fsk-modem
+
+   For full API reference, see the GitHub Pages
+   https://jgromes.github.io/RadioLib/
+*/
+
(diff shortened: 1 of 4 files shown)


jgromes:
This was done previously for CC1101 in https://github.com/jgromes/RadioLib/pull/279, in a very similar way (waiting inside `startTransmit` for FIFO to be sufficiently empty to add more bytes).

> this implementation will capture control and just sit in a loop until all of a message has been pushed into the FIFO, which I suspect might not be fantastic.

You're absolutely right: that's the general issue I see with this approach. It basically turns the non-blocking `startTransmit` method into a blocking one, since it has to wait until some data is transmitted. For low datarates, that can take quite some time.

However, a more generic approach to handling packets larger than FIFO is being discussed in https://github.com/jgromes/RadioLib/issues/399 and (primarily) https://github.com/jgromes/RadioLib/issues/201

justephens:
Thanks for pointing me to those issues (which I probably should have searched for myself, oops). I have a break next week and will try my hand at implementing the feature in a proper way then.

Any guidance on how to implement the feature in way friendly to multiple chips?

jgromes:
Take a look at other methods that are shared accross the modules, for example `transmit()` method. It has a pure virtual method in `PhysicalLayer` class, which is then implement in all classes derived from `PhysicalLayer`: `SX127x`, `CC1101` etc., so then the `transmit` method for any module may be called by calling its `PhysicalLayer` interface.

jgromes:
Since there was no progress on this for quite some time, and Stream mode is now supported for SX127x modules, I'm closing this PR - sorry about that @justephens.

PR #435, opened 2021-12-28
Fix: Si4432 interrupt examples

### LINKED ISSUE #434: Examples for Si4432 module are crashing on ESP8266
**Describe the bug**
When running [Si4432x_Receive_Interrupt.ino](https://github.com/jgromes/RadioLib/blob/master/examples/Si443x/Si443x_Receive_Interrupt/Si443x_Receive_Interrupt.ino) or [Si4432x_Transmit_Interrupt.ino](https://github.com/jgromes/RadioLib/blob/master/examples/Si443x/Si443x_Transmit_Interrupt/Si443x_Transmit_Interrupt.ino) on ESP8266 mcu, it crashes on runtime.

Error : `ISR not in IRAM!`

**To Reproduce**
1. Load the ESP8266 board with any of the above mentioned Si4432 examples
2.  Open the serial monitor
3. You shold see an error similar to the screenshot below. 

**Expected behavior**
ESP8266 mcu should not crash when running the example.

**Screenshots**
image

**Possible Solution**
Since ESP8266 requires the ISR functions to be stored in RAM, the ICACHE_RAM_ATTR macro could be added before the interrupt function.
Obs: I'm not shure and I don't have one to test it, but I think the same is valid for the ESP32 mcu.

Something like:
```
...
#ifdef ESP8266
  ICACHE_RAM_ATTR
#endif
void setFlag(void) {
...
```

**Additional info (please complete):**
 - MCU: ESP8266
 - Wireless module type Si4432
 - Arduino IDE version 1.8.16
 - Library version 5.0.0

Arthurk12:
Added ICACHE_RAM_ATTR macro to Si4432 interrupt examples when compiling for ESP8266.

Fixes #434

### DIFF
--- a/examples/Si443x/Si443x_Receive_Interrupt/Si443x_Receive_Interrupt.ino
+++ b/examples/Si443x/Si443x_Receive_Interrupt/Si443x_Receive_Interrupt.ino
@@ -76,6 +76,9 @@ volatile bool enableInterrupt = true;
 // is received by the module
 // IMPORTANT: this function MUST be 'void' type
 //            and MUST NOT have any arguments!
+#ifdef ESP8266
+  ICACHE_RAM_ATTR
+#endif
 void setFlag(void) {
   // check if the interrupt is enabled
   if(!enableInterrupt) {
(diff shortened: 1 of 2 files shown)


jgromes:
Fixed for all examples separately.

PR #458, opened 2022-01-31
[Si443x] Fixes getPacketLength

Arthurk12:
Fixes getPacketLength function to read the correct register when on fixed packet length mode.


jgromes:
Merged, thanks!

PR #481, opened 2022-02-23
Add method to support SX128x Frequency Error of last received LoRa or…

JasonRJ:
… Ranging packet.  Updated SX128x receive examples.

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -1112,7 +1112,7 @@ float SX128x::getSNR() {
   }
 }
 
-double SX128x::getFrequencyError() {
+float SX128x::getFrequencyError() {
     // check active modem
     uint8_t modem = getPacketType();
     if (!((modem == RADIOLIB_SX128X_PACKET_TYPE_LORA) || (modem == RADIOLIB_SX128X_PACKET_TYPE_RANGING))) {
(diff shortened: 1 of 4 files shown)


src/modules/SX128x/SX128x.cpp
```diff
@@ -1064,6 +1112,39 @@ float SX128x::getSNR() {
   }
 }
 
+double SX128x::getFrequencyError() {
```
jgromes, line 1115: `float` should be used instead of `double` - there's no need for double-precision, and platforms with single-precision FPU might not be able to take advatnage of it.

jgromes:
Thank you for the contribution! I left one comment.

Was this tested with hardware? And if so, what was the performance?

src/modules/SX128x/SX128x.cpp
JasonRJ: OK changing method and calculations from double to float.  I used the Semtech frequency error example code from mbed and they used double for the calculation.  I just continued using double.  I also used setGainControl as an example of reading the SX128x registers. I am not sure if the  RADIOLIB_ASSERT(state) is required after each read?

The method has been tested on an ESP32 SPI connected to an Ebyte E28 2G4M27S transceiver based on the SX1280.  Using a second E28 to cycle through -200kHz of receiver frequency to +250kHz and the results of the Frequency Error method were reasonably accurate with the changing transmit frequency.

No performance testing was done other than receiving four packets a second during the frequency sweep test.

JasonRJ:
The method has been tested on an ESP32 SPI connected to an Ebyte E28 2G4M27S transceiver based on the SX1280.  Using a second E28 to cycle through -200kHz of receiver frequency to +250kHz and the results of the Frequency Error method were reasonably accurate with the changing transmit frequency.

No performance testing was done other than receiving four packets a second during the frequency sweep test.

jgromes:
> No performance testing was done other than receiving four packets a second during the frequency sweep test.

The important part is that it went through some form of functional testing. Seeing as it has, I merged the PR, thanks for the contribution!

PR #485, opened 2022-02-28
Introduce asynchronous reception and transmission for CC1101

obones:
This allows using the asynchronous mode from the CC110, should a project need it for "raw" transmissions. This is most useful in projects that do timing and packet analysis by themselves, such as RFLink32 for instance.

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -174,6 +174,14 @@ int16_t CC1101::standby() {
 }
 
 int16_t CC1101::transmitDirect(uint32_t frf) {
+  return transmitDirect(true, frf);
+}
+
+int16_t CC1101::transmitDirectAsync(uint32_t frf) {
+  return transmitDirect(false, frf);
+}
+
+int16_t CC1101::transmitDirect(bool sync, uint32_t frf) {
   // set RF switch (if present)
   _mod->setRfSwitchState(LOW, HIGH);
 
@@ -187,7 +195,7 @@ int16_t CC1101::transmitDirect(uint32_t frf) {
   }
 
   // activate direct mode
-  int16_t state = directMode();
+  int16_t state = directMode(sync);
   RADIOLIB_ASSERT(state);
 
   // start transmitting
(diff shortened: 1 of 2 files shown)


jgromes:
Thank you for the contribution!

One point though: this introduces some code duplication. Overall, I believe the only meaningful difference between the current `directMode` and the new `asyncDirectMode` is one value in `RADIOLIB_CC1101_REG_PKTCTRL0`.

I think it would be better to add a parameter to `directMode(bool async = false)`, so that it defaults to the current synchronous direct mode (since that's supported on other modules as well).

The public API methods `transmitDirectAsync`/`receiveDirectAsync` should remain.

obones:
> I believe the only meaningful difference between the current `directMode` and the new `asyncDirectMode` is one value in `RADIOLIB_CC1101_REG_PKTCTRL0`.

Well, it's not just this one, there is also a different mapping for the GD0 pin: `CC1101_REG_IOCFG0` is set to `CC1101_GDOX_SERIAL_DATA_ASYNC`

But I agree that there is quite a bit of code duplication, most noticeably in `transmitDirectAsync`. Wouldn't it be better if I was to change my PR to get this final result:

image

I believe this would remove all code duplication.

jgromes:
Unfortunately we can't change the signatures of `transmitDirect`/`receiveDirect`, because they implement pure virtual methods from `PhysicalLayer`.

obones:
Yes, I know that, but I'm not suggesting to change those, I'm just adding a new "overload" with an additional parameter called `sync` that gets called by the `override` ones. Here is an example with `receiveDirect`:

```c
int16_t CC1101::receiveDirect() {
  return receiveDirect(true);
}

int16_t CC1101::receiveDirectAsync() {
  return receiveDirect(false);
}

int16_t CC1101::receiveDirect(bool sync) {
  // set RF switch (if present)
  _mod->setRfSwitchState(HIGH, LOW);

  // activate direct mode
  int16_t state = directMode(sync);
  RADIOLIB_ASSERT(state);

  // start receiving
  SPIsendCommand(CC1101_CMD_RX);
  return(ERR_NONE);
}
```

This compiles and runs just fine, `receiveDirect()` is left as is in the header file, while `receiveDirect(bool)` is added to the `protected` section at the bottom, near `directMode()`

jgromes:
Ah, I see - in that case it's OK.

Was this tested with hardware?

obones:
> Was this tested with hardware?

Yes, I have it running here just fine, as used by my yet to be published work on the [RFLink32](https://github.com/cpainchaud/RFLink32) project

jgromes:
Seems the CI failed on `ERR_NONE` being undefined. This changed in the 5.0.0 release, now it's `RADIOLIB_ERR_NONE`.

Please double check you're using the latest library version.

obones:
I have pushed the discussed changes, I believe this is now just fine.

jgromes:
Looking good now - merged, thanks for the contribution!

PR #499, opened 2022-03-24
[CC1101] Data shaping and encoding params in right order

loopj:
For some reason, the `RADIOLIB_SHAPING_NONE` was being passed to `setEncoding` and `RADIOLIB_ENCODING_NRZ` was being passed to `setDataShaping`. Flipped these to be passed to the correct functions.

Note: These are both defined as `0x00` so this should have no impact on functionality, just makes initialization less confusing.


jgromes:
Probably a typo, but thanks for fixing it!

PR #500, opened 2022-03-26
[RF69] Set RX bandwidth correctly for OOK mode

loopj:
The formula for calculating RX bandwidth on the RF69 differs depending on if you are in FSK mode or OOK mode (see datasheet section 3.4.6). 

This PR fixes RX bandwidth selection by using the correct exponent value when in OOK mode.

This bug was present prior to #498 being merged, but #498 makes it much easier to fix.

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -464,7 +464,7 @@ int16_t RF69::setRxBandwidth(float rxBw) {
   // calculate exponent and mantissa values for receiver bandwidth
   for(int8_t e = 7; e >= 0; e--) {
     for(int8_t m = 2; m >= 0; m--) {
-      float point = (RADIOLIB_RF69_CRYSTAL_FREQ * 1000000.0)/(((4 * m) + 16) * ((uint32_t)1 << (e + 2)));
+      float point = (RADIOLIB_RF69_CRYSTAL_FREQ * 1000000.0)/(((4 * m) + 16) * ((uint32_t)1 << (e + (_ook ? 3 : 2))));
       if(fabs(rxBw - (point / 1000.0)) <= 0.1) {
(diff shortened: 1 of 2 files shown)


loopj:
I'm actually not sure if this approach is complete, since I imagine most people are configuring rxBw by calling `RF69::begin`, and there is no (current) way to enable OOK mode from `begin`.

This means if someone calls
```
rf69.begin(.., someRxBw,...)
```
then calls
```
rf69.setOOK(true);
```

The rx bandwidth will not be set correctly.

There are a couple of approaches I can see to making this work more reliably.

1. Add an optional `enableOOK` param to `RF69::begin` (like we have in SX127x)
2. Call `RF69::setRxBandwidth(_rxBw)` when `RF69::setOOK` is called.

I prefer option #1, but would love to hear opinions before I update this PR.

loopj:
Went ahead and added an optional `enableOOK` param to `RF69::begin`, since I think this is the cleanest solution and there is precedence in other modules (eg SX127x).

jgromes:
> Went ahead and added an optional enableOOK param to RF69::begin

I think that solution is fine, though strictly speaking it's still incomplete. If one were to change OOK enable/disable during runtime, the Rx bandwidth would remain misconfigured.

I think that we should reconfigure Rx bandwidth whenever `setOOK` is called, and add a note to the Doxygen docs of `setOOK`, to let the user know that this action will take place. Alternatively, we could just add a note to `setOOK` that `setRxBandwidth()` must be called when switching OOK on or off. I would prefer the first option, since it seems like an action we can take without needing interaction from the user.

loopj:
Agreed. I've gone ahead and updated my PR to reflect that, and also removed the enableOOK param I previously added to `begin`. Thanks for the feedback.

jgromes:

PR #501, opened 2022-03-26
[RF69] Persist bit synchronization settings

loopj:
This PR persists `disableContinuousModeBitSync` or `enableContinuousModeBitSync` preferences so that any subsequent calls to `receiveDirect` respect the previously set bit sync preference. 

This is particularly important when building a protocol that uses continuous mode without bit sync, since we are interacting with the radio using `PhysicalLayer` methods and cannot always call `disableContinuousModeBitSync` after calling `receiveDirect`.

This PR defaults to bit sync "enabled" when calling `receiveDirect` to maintain the current behavior.


jgromes:
Looking good, thank you for the contribution!

Issue #524, opened 2022-05-22
[SX1276] Request to add setCrcFiltering(bool crcOn = true)


matthias-bs:
For my application, I have to disable CRC filtering in FSK packet mode.

As a workaround, I changed SX127x.cpp as follows:
In `SX127x::configFSK()`:
`state = _mod->SPIsetRegValue(RADIOLIB_SX127X_REG_PACKET_CONFIG_1,[...] | RADIOLIB_SX127X_CRC_ON | [...], 7, 0);`
replaced by
`state = _mod->SPIsetRegValue(RADIOLIB_SX127X_REG_PACKET_CONFIG_1,[...] | RADIOLIB_SX127X_CRC_OFF | [...], 7, 0);`

The CC1101 module has a function `setCrcFiltering(bool crcOn)` for this purpose. It would be great if the SX127x had the same function.

matthias-bs:
Added proposed solution:
Pull request [[SX1276] Request to add setCrcFiltering()](https://github.com/jgromes/RadioLib/pull/525)

matthias-bs:
Added proposed solution:
Pull request [[SX127x] added setCrcFiltering()](https://github.com/jgromes/RadioLib/pull/527)
(This time changed from a fresh fork. =:-) )

jgromes:
Resolved by merging #527.

matthias-bs:
Thanks a lot!

Issue #505, opened 2022-03-29
SX127x Missing packets in FSK mode


skywodd:
**Describe the bug**
When using an SX1276 module in FSK mode with all defaults settings (only frequency is set in `beginFSK()`), the DIO0 interrupt will only trigger half the time when a packet is sent.

The hardware itself seem to be working properly, by replacing `beginFSK(868.3)` with `begin(868.3)` in both TX and RX sketch, everything work as expected in LoRa mode.

**To Reproduce**

Test procedure:
- Prepare a board with an SX1276 module and the RX sketch below. 
- One packet is sent every **3** seconds using another identical board and the TX sketch below.
- Using an RTLSDR dongle, RF activity is confirmed (one packet every **3** seconds over the air).

Observed behavior:
- Only one pulse is seen on the DIO0 pin of the receiver board every **6** seconds using a logic analyser
- The receiver code only read and display one packet every **6** seconds.
- No error is returned by `readData()`or `startReceive()`.

TX sketch :
```
#include <RadioLib.h>

SX1276 radio = new Module(10, 3, 9, 4);

void setup() {
  Serial.begin(9600);

  int state = radio.beginFSK(868.3);
  if (state != RADIOLIB_ERR_NONE) {
    Serial.print(F("Begin failed, code "));
    Serial.println(state);
    while (true);
  }
}

void loop() {
  byte byteArr[] = {0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF};
  int state = radio.transmit(byteArr, 8);
  if (state != RADIOLIB_ERR_NONE) {
    Serial.print(F("Transmit failed, code "));
    Serial.println(state);
  }

  delay(3000);
}
```

RX Sketch :
```
#include <RadioLib.h>

SX1276 radio = new Module(10, 3, 9, 4);

volatile bool receivedFlag = false;
volatile bool enableInterrupt = true;

void setup() {
  Serial.begin(9600);

  int state = radio.beginFSK(868.3);
  if (state != RADIOLIB_ERR_NONE) {
    Serial.print(F("Begin failed, code "));
    Serial.println(state);
    while (true);
  }

  radio.setDio0Action(setFlag);

  state = radio.startReceive();
  if (state != RADIOLIB_ERR_NONE) {
    Serial.print(F("Receive failed, code "));
    Serial.println(state);
    while (true);
  }
}

void setFlag(void) {
  if(!enableInterrupt)
    return;
    
[...]

PR #525, opened 2022-05-23
[SX1276] Request to add setCrcFiltering()

matthias-bs:
Proposed solution for [SX1276] Request to add setCrcFiltering(bool crcOn = true)](https://github.com/jgromes/RadioLib/issues/524)

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -52,7 +52,7 @@ int16_t SX127x::begin(uint8_t chipVersion, uint8_t syncWord, uint16_t preambleLe
 
   // initialize internal variables
   _dataRate = 0.0;
-
+  
   return(state);
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/SX127x/SX127x.cpp
```diff
@@ -96,12 +96,17 @@ int16_t SX127x::beginFSK(uint8_t chipVersion, float br, float freqDev, float rxB
...
+  } else {
+    state = SX127x::setAFCAGCTrigger(RADIOLIB_SX127X_RX_TRIGGER_BOTH);
+  }
+
   RADIOLIB_ASSERT(state);
 
   // enable AFC
-  state = SX127x::setAFC(false);
+  state = SX127x::setAFC(true);
   RADIOLIB_ASSERT(state);
```
jgromes, line 110: The disabled AFC and AFC/AGC triggers are a workaround for https://github.com/jgromes/RadioLib/issues/505, the PR would overwrite those. I think you might be merging from some older version.

jgromes:
This is fine to add, however, looking through the PR, I noticed that it would overwrite a workaround for #505. Please remove that change from this PR, after that it will be fine.

PR #527, opened 2022-05-23
[SX127x] added setCrcFiltering()

matthias-bs:
proposed solution for [[SX1276] Request to add setCrcFiltering(bool crcOn = true)](https://github.com/jgromes/RadioLib/issues/524)


jgromes:
Merged, thanks!

PR #534, opened 2022-06-10
Allow SPIbeginTransaction and SPIendTransaction to be overridden by subclasses

caveman99:
This allows API clients to provide 'smarter' versions of Module that can do things like add thread safety so that multiple devices (and service threads) can share the same SPI bus. i.e. a subclass would lock some sort of mutex.

This is used in the Radiolib Fork for Meshtastic to share an SPI bus with SD Card and TFT Display but we want to move back to upstream and not maintain our own fork.

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -190,7 +190,7 @@ class Module {
 
       \param numBytes Number of bytes to transfer.
     */
-    virtual void SPItransfer(uint8_t cmd, uint8_t reg, uint8_t* dataOut, uint8_t* dataIn, uint8_t numBytes);
+    void SPItransfer(uint8_t cmd, uint8_t reg, uint8_t* dataOut, uint8_t* dataIn, uint8_t numBytes);
 
     // pin number access methods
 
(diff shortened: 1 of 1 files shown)


jgromes:
I'm guessing this way was easiest for you, however, I do have to wonder whether it's the SPI transfer method that should be the point at which mutex is handled - for instance, this will not work for modules with their own `SPItransfer` methods, such as nRF24, SX126x and SX128x.

I would suggest to override the `void Module::SPIbeginTransaction()` and `void Module::SPIendTransaction()`, since the only thing they call are the appropriate `SPIClass` methods, so it should be pretty easy to add a mutex lock/unlock.

caveman99:
Indeed is was easier, since we have a Concurrency LockGuard, which locks a semaphore in its constructor and frees it in the destructor, so on exiting the wrapper, the lock would be released. with SPIbeginTransaction and SPIendTransaction we'd have to do it in 2 steps, but still possible of course.

jgromes:
Merged! Thanks for making that modification - now it should also work with SX126x and the other modules.

caveman99:
thanks, will be included in our next preview release.

PR #540, opened 2022-06-24
RadioLib - getTimeOnAir for SX127x modules

Jaimi5:
New function in SX127x module that returns the time on air of a lora packet.

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -1064,7 +1064,7 @@ uint32_t SX127x::getTimeOnAir(size_t len) {
     // Get number of bits preamble
     float n_pre = (float) ((_mod->SPIgetRegValue(RADIOLIB_SX127X_REG_PREAMBLE_MSB_FSK) << 8) | _mod->SPIgetRegValue(RADIOLIB_SX127X_REG_PREAMBLE_LSB_FSK)) * 8;
     //Get the number of bits of the sync word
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks for the contribution!

One point though: the method will fail if the FSK modem is active. I would suggest to check the active modem and then either:

1. returning the "wrong modem" error code, or
2. calculating the time-on-air for the FSK modem, based on preamble length, sync word length, packet type (fixed/variable length), payload length, CRC presence and bit rate.

Jaimi5:
Hi @jgromes !
I updated a new version including the calculation time-on-air for FSK modem, but I have some questions and I don't really know if it is implemented right.
When reading syncWord length, I add +1, as the inverse when adding to the register, the library is doing a -1 in the length.
Is the CRC for FSK 16 bits long?
And finally, I don't really know if it is correct to check for the variable or fixed payload because of the parameter of the function?

Thanks! And sorry for having so many questions, I would like to implement it good enough!

jgromes:
> When reading syncWord length, I add +1, as the inverse when adding to the register, the library is doing a -1 in the length.
Is the CRC for FSK 16 bits long?

That's correct, the actual sync word sent is one byte longer than what is written in the register.

> Is the CRC for FSK 16 bits long?

Yes, that is always two bytes.

> I don't really know if it is correct to check for the variable or fixed payload because of the parameter of the function?

Not entirely sure what you mean by that - the packet length depends on whether the device is in fixed packet length mode, or variable mode. In variable packet length mode, there's one extra byte in the packet that contains the length of the payload.

jgromes:
Looking good now, thanks!

PR #542, opened 2022-07-06
Introduce the setDIOMapping method

obones:
As discussed in #532
Implementation done for CC1101, RFM69 and SX1278.
Validity of mapping tables has also been verified for their derived classes (SX1231, RFM96...)

### DIFF
--- a/src/modules/SX127x/SX1278.cpp
+++ b/src/modules/SX127x/SX1278.cpp
@@ -471,6 +471,20 @@ int16_t SX1278::explicitHeader() {
   return(setHeaderType(RADIOLIB_SX1278_HEADER_EXPL_MODE));
 }
 
+int16_t SX1278::setDIOMapping(RADIOLIB_PIN_TYPE pin, uint8_t value) {
+  if (pin > 5)
+    return RADIOLIB_ERR_INVALID_DIO_PIN;
+
+  if (pin < 4)
+    return(_mod->SPIsetRegValue(RADIOLIB_SX1278_REG_DIO_MAPPING_1, value, 7 - 2 * pin, 6 - 2 * pin));
+  else
+    return(_mod->SPIsetRegValue(RADIOLIB_SX1278_REG_DIO_MAPPING_2, value, 15 - 2 * pin, 14 - 2 * pin));
+}
+
+int16_t SX1278::setDIOPreambleDetect(bool usePreambleDetect) {
+  return _mod->SPIsetRegValue(RADIOLIB_SX1278_REG_DIO_MAPPING_2, (usePreambleDetect) ? RADIOLIB_SX1278_DIO_MAP_PREAMBLE_DETECT : RADIOLIB_SX1278_DIO_MAP_RSSI, 0, 0);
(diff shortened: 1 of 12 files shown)


src/modules/SX127x/SX1278.cpp
```diff
@@ -471,6 +471,20 @@ int16_t SX1278::explicitHeader() {
...
+  if (pin > 5)
+    return RADIOLIB_ERR_INVALID_DIO_PIN;
+
+  if (pin < 4)
+    return(_mod->SPIsetRegValue(RADIOLIB_SX1278_REG_DIO_MAPPING_1, value, 7 - 2 * pin, 6 - 2 * pin));
+  else
+    return(_mod->SPIsetRegValue(RADIOLIB_SX1278_REG_DIO_MAPPING_2, value, 15 - 2 * pin, 14 - 2 * pin));
+}
+
+int16_t SX1278::setDIOPreambleDetect(bool usePreambleDetect) {
```
jgromes, line 484: Is it necessary to have this as a separate method? Seems very specific compared to the other generic methods.

src/modules/SX127x/SX1278.h
```diff
@@ -10,6 +10,8 @@
 
 // SX1278 specific register map
 #define RADIOLIB_SX1278_REG_MODEM_CONFIG_3                     0x26
+#define RADIOLIB_SX1278_REG_DIO_MAPPING_1                      0x40
+#define RADIOLIB_SX1278_REG_DIO_MAPPING_2                      0x41
```
jgromes, line 14: This (and actually all of the changes in SX1278.h/.cpp) should be done in SX127x files, to be common to all the SX127x modules (unless there's a specific incompatibility between SX1278/76 and 72/73.

jgromes (commented):
Thanks, I left some comments - if we can resolve those then I think it will be OK to merge.

src/modules/SX127x/SX1278.cpp
obones: This is used to set bit 1 of `RegDioMapping2` which in turn controls what signal is effectively present on the DIO pin when using one of the `RSSI_PREAMBLE_DETECT` values. 
For instance, if you use `RADIOLIB_SX1278_DIO1_CONT_RSSI_PREAMBLE_DETECT` then DIO1 will output either "RSSI threshold reached" or  "Preamble detected" depending on bit 1 of `RegDioMapping2`
So yes, it is specific to the `SX1278` but not in the sense that there is a special method for a given DIO signal, rather a specific method to activate a specific function from the chip itself.

src/modules/SX127x/SX1278.h
obones: Thing is, I started with the chip I have here and went down the inheritance chain, but not up.
That being said, I checked and you are right, SX1278 and SX1273 have the same behavior when it comes to DIO pins, so I moved the required support code to the SX127x.h/.cpp files in the latest commit.

jgromes:
All good now, thanks for the contribution!

PR #554, opened 2022-08-04
remedy Module when RADIOLIB_BUILD_ARDUINO is not defined

majbthrd:
In its present form, it seems RadioLib cannot be compiled when RADIOLIB_BUILD_ARDUINO is not defined.

This is because the source code in the non-RADIOLIB_BUILD_ARDUINO variant of Module.cpp makes repeated use of functionality unique to when RADIOLIB_BUILD_ARDUINO is defined.

So, this is a proposed PR to remedy these issues when RADIOLIB_BUILD_ARDUINO is not defined.

One change in ./src/BuildOpt.h is to modify the RADIOLIB_GENERATE_CALLBACK_RET_FUNC and RADIOLIB_GENERATE_CALLBACK_SPI_RET_FUNC macros so that the default value is nullptr.  These values are only written by the RADIOLIB_BUILD_ARDUINO constructor, and the Module source code *presumes* that a nullptr check will work.  However, that check will only work if these values are initialized to nullptr to begin with.

In the present ./src/Module.cpp, the non-RADIOLIB_BUILD_ARDUINO variant makes use of "_initInterface", a value that only exists when RADIOLIB_BUILD_ARDUINO is defined.  It then presumes a callback value to be valid.  This PR places such code within an #ifdef  RADIOLIB_BUILD_ARDUINO.

In the present ./src/Module.cpp,  the non-RADIOLIB_BUILD_ARDUINO variant makes use of SPIbeginTransaction(), SPItransfer(uint8_t), and SPIendTransaction(), which are presently only defined when RADIOLIB_BUILD_ARDUINO defined.  The proposed solution is to make all three of these available regardless of RADIOLIB_BUILD_ARDUINO and defined as virtual functions to allow the user to overload these for the non-Arduino SPI implementation.

The final change is to allow RadioLib to be more easily integrated into projects.  In its present form, RadioLib requires modification of ./src/BuildOpt.h when RADIOLIB_BUILD_ARDUINO is not defined; the comments even direct users to do this.  However, this is not terribly practical when RadioLib is a git submodule of another project.  Moreover, every time the source code changes, the same modification must be re-applied.  So, to remedy, an #include "noarduino.h" was added to so that the user simply has to put such a file somewhere in the include path.  (They are, alas, still on the hook to update this noarduino.h file whenever yet another #define is added to ./src/BuildOpt.h, but the important difference is that the RadioLib submodule can remain untouched.)

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -886,6 +886,7 @@
   #define RADIOLIB_PLATFORM                           "Generic"
 
   // platform properties may be defined here, or somewhere else in the build system
+  #include "noarduino.h"
 
 #endif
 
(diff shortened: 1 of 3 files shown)


jgromes:
Thanks for this - the whole idea behind RADIOLIB_BUILD_ARDUINO macro is to allow building RadioLib on non-Arduino platforms. However, I don't think anyone ever actually tried using it - I put the framework together in a few days, and there was never any sort of feedback.

All the changes look good, merged. Thanks for the contribution!

PR #561, opened 2022-08-24
Introduce the setRSSIThreshold method

obones:
For RF69 and SX127x families of chips

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -872,11 +872,11 @@ float RF69::getRSSI() {
   return(-1.0 * (_mod->SPIgetRegValue(RADIOLIB_RF69_REG_RSSI_VALUE)/2.0));
 }
 
-int16_t RF69::setRSSIThreshold(float value) {
-  if (value < 127.5 || value > 0)
+int16_t RF69::setRSSIThreshold(float dbm) {
+  if (dbm < 127.5 || dbm > 0)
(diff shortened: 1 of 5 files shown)


src/modules/RF69/RF69.cpp
```diff
@@ -872,6 +872,13 @@ float RF69::getRSSI() {
   return(-1.0 * (_mod->SPIgetRegValue(RADIOLIB_RF69_REG_RSSI_VALUE)/2.0));
 }
 
+int16_t RF69::setRSSIThreshold(float dbm) {
+  if (dbm < 127.5 || dbm > 0)
+    return RADIOLIB_ERR_INVALID_RSSI_THRESHOLD;
```
jgromes, line 877: There's a mcro to do these sorts of checks: RADIOLIB_CHECK_RANGE(VAR, MIN, MAX, ERR), please use that instead. For example `RADIOLIB_CHECK_RANGE(dbm, -127.5, 0, RADIOLIB_ERR_INVALID_RSSI_THRESHOLD);`

Also the 127.5 should probably be negative.

jgromes:
Thank you, I left a comment.

src/modules/RF69/RF69.cpp
obones: I did not know about the `RADIOLIB_CHECK_RANGE` macro 
As to the negative value for 127.5, I will go hide in a dark corner for this bad copy/paste between my two repositories...

Anyway, the latest commit fixes both

PR #609, opened 2022-11-15
add Portduino as a platform

caveman99:
This way we can ged rid of the maintained fork for Meshtastic's own Portduino Platform (Arduino API on x86). Portduino is mimicking an UNO WiFi board and implements parts of the original Arduino API. The Rest of Meshtastic is using the upstream Radiolib already.

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -426,7 +426,7 @@
     #define RADIOLIB_CB_ARGS_SPI_END_TRANSACTION        (void, SPIendTransaction, void)
     #define RADIOLIB_CB_ARGS_SPI_END                    (void, SPIend, void)
 
-  #elif defined(ARDUINO_AVR_UNO_WIFI_REV2) || defined(ARDUINO_AVR_NANO_EVERY)
+  #elif defined(ARDUINO_AVR_UNO_WIFI_REV2) || defined(ARDUINO_AVR_NANO_EVERY) || defined(PORTDUINO)
     // Arduino megaAVR boards - Uno Wifi Rev.2, Nano Every
     #define RADIOLIB_PLATFORM                           "Arduino megaAVR"
     #define RADIOLIB_PIN_TYPE                           uint8_t


jgromes:
It seems strange to be using the same definition for Portduino as for Uno Wifi, which is fundementally different hardware. I think it would be better to have a separate definition for Portduino, even if it is just a copy-paste of everything from Uno WiFi. That being said, I don't see this as major issue.

What could be a problem is that as a new platform, Portduino should be covered by the CI, which builds all examples for all officially supported platforms. The CI uses arduino-cli to install platforms and build examples, and I'm not sure if Portduino supports that.

caveman99:
Portduino is not a full implementation of the Arduino API, and we indeed had a specific 'Portduino' definition section in the fork we maintained. It is geared to run with the 'native' build environment of Platformio through a tailor-made platform build script, so i doubt it will build with arduino-cli.

More than 90% of the API are just stubs to the underlying OS or 'noop', it's more like a shim to make Radiolib usable on non-embedded hardware. Right now we use it to build a simulated radio firmware for regression tests in our own CI but the goal is to use it to support radio hardware on linux, whether connected by GPIO or an USB adapter.

I'm fine with either approach, i just wanna get rid of maintaining a dozen of lib forks ;-)

jgromes:
Thanks for the clarification, this is very interesting. I think there were a couple other attempts to run RadioLib on Linux (should be listed in forks), and quite some time ago I was running RadioLib on Raspberry Pi using RasPiDuino framework.

I'm fine with having this as is, but unless CI can be implemented, I cannot guarantee futurue commits won't break something.

caveman99:
That's absolutely fine, our major builds are not depending on this and we build every platform on CI with every commit as well. If it breaks we have to fix it on our side ;-)

jgromes:
OK, then it's merged.

BTW yesterday I got a TTGO T-Beam off ebay, powered it on for the first time - and Meshtastic was already installed on it :)

caveman99:
Nice, thanks. :-)

PR #610, opened 2022-11-15
damaged header error fix

IZ7BOJ:
Damaged header error check only in IMPLICIT mode

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -623,7 +623,7 @@ int16_t SX127x::readData(uint8_t* data, size_t len) {
 
   if(modem == RADIOLIB_SX127X_LORA) {
     // check packet header integrity
-    if(_crcEnabled && (_mod->SPIgetRegValue(RADIOLIB_SX127X_REG_HOP_CHANNEL, 6, 6)) == 0) {
+    if((_mod->SPIgetRegValue(RADIOLIB_SX127X_REG_MODEM_CONFIG_1, 0, 0))==1 && _crcEnabled && (_mod->SPIgetRegValue(RADIOLIB_SX127X_REG_HOP_CHANNEL, 6, 6)) == 0) {
       // CRC is disabled according to packet header and enabled according to user
       // most likely damaged packet header
       clearIRQFlags();


jgromes:
I don't understant - the section you have modified checks wheter the received packet header says CRC should be enabled. In implicit mode, there is no packet header to check.

What are you trying to fix?


Screenshot_36

IZ7BOJ:
Hi Jan,
I'll try to explain better.
I used a TTGO board with the last version of OE5BPA Software which uses
radiolib.
If I transmit in explicit mode but without any CRC, The ttgo states a
radiolib error -24, which translates to a "RADIOLIB_ERR_LORA_HEADER_DAMAGED
(-24)".
As far as I understand, this error should be reported only if:
1) modem is in implicit mode
2) RxPayloadCrcOn is 0 on TX side
3) RxPayloadCrcOnis 1 on RX side

The case is "CRC is always wrong" of the following table (extracted from
the datasheet):

[image: immagine.png]

If the modem is in explicit mode (as in my case), the situation is the
following:
[image: immagine.png]

As far as I understand, if RxPayloadCrcOn is 0 on TX side, the RX should
simply not perform the crc check, and should not give any "damaged header"
error.

I hope I've been more clear now.
I already tested the pull request on my TTGO and now it decodes the beacons
in explicit mode with or without the crc.
73's
Alfredo IZ7BOJ



Il giorno mar 15 nov 2022 alle ore 18:16 Jan Gromeš <
***@***.***> ha scritto:

IZ7BOJ:
Hi Jan,
you say "In implicit mode, there is no packet header to check.".
I would say "In implicit mode, the field which indicates if the crc is
present or not, is not present".
But the CRC itself could be present, even in implicit mode!
Bests
Alfredo

Il giorno mar 15 nov 2022 alle ore 18:40 Alfredo Vania ***@***.***>
ha scritto:

jgromes:
> I used a TTGO board with the last version of OE5BPA Software which uses

I don't know what that is, can you post a link? Also the images seem to be missing.

>But the CRC itself could be present, even in implicit mode!

That's correct, but that is not what this error is about - it's about mismatch between what the user configured (CRC present or not), and what the LoRa header is reporting. The following are the options:

1. Explicit header mode, CRC on Tx enabled, CRC on Rx enabled - no issue, CRC will be checked on Rx
2. Implicit header mode, CRC on Tx enabled, CRC on Rx enabled - no issue, CRC will be checked on Rx
3. Explicit header mode, CRC on Tx disabled, CRC on Rx enabled - CRC Rx/Tx mismatch, packet will report no CRC presence while local configuration says there should be a CRC. Here, error is raised.
4. Implicit header mode, CRC on Tx disabled, CRC on Rx enabled - CRC Rx/Tx mismatch, but there is no packet header to inform us, CRC will attempt check on Rx but fail (because there is no way to know on Rx side if CRC is actually enabled on Tx or not).
5. Explicit header mode, CRC on Tx enabled, CRC on Rx disabled - CRC Rx/Tx mismatch, CRC will not be checked on Rx and there will be two extra "payload" bytes returned.
6. Implicit header mode, CRC on Tx enabled, CRC on Rx disabled - CRC Rx/Tx  mismatch, CRC will not be checked on Rx and there will be two extra "payload" bytes returned.
7. Explicit header mode, CRC on Tx disabled, CRC on Rx disabled - no issue, CRC will not be checked on Rx
8. Implicit header mode, CRC on Tx disabled, CRC on Rx disabled - no issue, CRC will not be checked on Rx

You can see that only one of the options leads to the error being returned, and requires explicit header mode. Your change would cause that condition to never be true becuase in implcit header mode, you do not know if the transmitting side added a CRC or not).

More to the point, I think you should start by investigating why is your transmitter configured to not send CRCs while the receiver is configured to expect them, regardless of header mode.

IZ7BOJ:
Hi Jan,
The SW that I use on TTGO is this:
https://github.com/lora-aprs/LoRa_APRS_iGate .

It's strange that you don't see attached images. Anyway I refer to par
4.1.1.6. of SX1276/77/78/79 Semtech datasheet Rev.4 .
At page 30 you can find tables which report the same analysis that you did.
I compared these tables with your mail and there is match except one
difference:


*Explicit header mode, CRC on Tx disabled, CRC on Rx enabled - CRC Rx/Tx
mismatch, packet will report no CRC presence while local configuration says
there should be a CRC. Here, error is raised. *

It's my case. That's the point. Page 30 of the datasheets says that in this
case CRC will not be checked, so any error should be raised.
My interpretation is the following: in explicit mode, the RX local settings
for the crc are overridden by the remote TX settings. If the TX header has
RxPayloadCrcOn set to 0, the RX knows that the CRC field is NOT present,
and it doesn't matter if the local setting says the opposite. The CRC is
not present so it doesn't make sense to check it.

Moreover, the datasheet itself says:



*"In Explicit Header Mode, the presence of the CRC at the end of the
payload is selected only on the transmitter side through the bit
RxPayloadCrcOn in the register RegModemConfig1.On the receiver side, the
bit RxPayloadCrcOn in the register RegModemConfig1 is not used and once the
payload has been received, the user should check the bit CrcOnPayload in
the register RegHopChannel. If the bit CrcOnPayload is at ‘1’, the user
should then check the Irq Flag PayloadCrcError to make sure the CRC is
valid.If the bit CrcOnPayload is at ‘0’, it means there was no CRC on the
payload and thus the IRQ Flag PayloadCrcError will not be trigged even if
the payload has errors."*

For this reason, I conditioned the "wrong packet header" error only to
implicit case.

Let me know if you confirm my interpretation of the datasheet.

73's
Alfredo IZ7BOJ

Il giorno mar 15 nov 2022 alle ore 22:00 Jan Gromeš <
***@***.***> ha scritto:

jgromes:
It could be argued that there should be a way to retrieve data even if CRC check fails, or if there is a mismatch in the configuration. 

I will close this now, since this PR would basically remove the check completely (what's doing is basically if(A and !A)), and return data from the reception methods even if CRC fails (RADIOLIB_ERR_CRC_MISMATCH), or if there's a mismatch in configuration/damaged header (RADIOLIB_ERR_LORA_HEADER_DAMAGED). You can choose to ignore the error code and keep the data, that's up to you.

jgromes:
@IZ7BOJ FYI I added this handling, so that even if header error is returned, you can still check the data and decide what to do with it.

PR #612, opened 2022-11-18
Update SX126x.cpp to configure CAD based on set SF

Amalinda:


Amalinda:
- This pull request tries to configure the CAD module based on the set SF. 
- Currently, the CAD configuration is fixed for all SFs. This does not result in the most optimal configuration. In this pull request, I integrate the recommendation from AN1200.48 to configure CAD.

Amalinda:
It can be found under this link [https://lora-developers.semtech.com/documentation/product-documents] by searching for keyword CAD.

Not sure if this direct link would work but here is one: https://semtech.my.salesforce.com/sfc/p/E0000000JelG/a/2R000000Q1ES/SPexo9njbhEQLcJVUg1i0Su8p3tpAtwX1jhMBGlXsQI

jgromes:
Thanks, I found 4 links to this document on Semtech site and none of them worked ... I hate SalesForce.
Briefly going through the document, I think it actually would be beneficial to let the user set their own CAD parameters, and only use the Semtech-recommended ones if the user chooses not to change them.

I'm going to accept this PR as-is and add this change myself, since it's going to slightly change the interface.

Thanks for the contribution!

PR #623, opened 2022-11-26
No bound checks on frequency deviation if ~OOK

phretor:
Sometimes it's acceptable to set frequency deviation to 0, even in FSK. For example, I use it to "fool" the radio into acting as a spectrum, by changing the carrier frequency while in FSK and then taking samples of the RSSI.

So, unless the radio is in OOK, which makes sense to do frequency deviation range checks, I'd suggest to not validate this range.

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -506,7 +506,8 @@ int16_t CC1101::setFrequencyDeviation(float freqDev) {
     newFreqDev = 1.587;
   }
 
-  if (_modulation != RADIOLIB_CC1101_MOD_FORMAT_ASK_OOK) {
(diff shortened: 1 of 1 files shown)


jgromes:
I don't think it's a good idea to skip the check entirely - instead, would it be acceptable to have the value 0 as a special case that is not validated?

phretor:
Good idea! I like how RFQuack is trying to abuse the radios in every possible way and RadioLib is supporting that while acting as a very polite firewall to our silly ideas 😬

jgromes:
Thanks, that was pretty much the main design goal - a simple to use API on the surface, but ultimately letting the user to do pretty much anything ;)

jgromes:
Looks good now - merged, thanks!

PR #626, opened 2022-12-01
[RF69] Fixed setPromiscuousMode(false) corner case

phretor:
Before this fix, disabling promiscuous mode would *not* re-enable preamble detection with preamble to the default length.

Signed-off-by: Federico Maggi <federico.maggi@gmail.com>


phretor:
Found the same issue in CC1101. I'm closing this one and creating a new, single PR.

PR #627, opened 2022-12-01
[RF69/CC1101] Fixed corner case in `setPromiscuousMode(false)`

phretor:
In RF69:

- `setPromiscuousMode(false)` would no re-enable preamble detection/generation and set preamble to default length
- `disableSyncWordFiltering()` would also `setPreambleLength(0)` while this is supposed to be in `RF69::setPromiscuosuMode`

In CC1101:
- `setPromiscuousMode(false)` would no re-enable preamble detection/generation and set preamble to default length


jgromes:
Looks good, thanks!

PR #631, opened 2022-12-03
[RF69] Missing 1000.0 multipler in

phretor:
I'd forgotten to multiply the calculated FD by a factor of 1000.0


jgromes:
Merged, thanks!

Issue #588, opened 2022-10-12
Adding support for STM32WL integrated transceivers


matthijskooijman:
The STM32WL55 series combines an MCU with a LoRa-transceiver in a single module. I'm working together with STMicroelectronics to get some support for these chips in the Arduino ecosystem, and we're currently investigating implementing and contributing support in RadioLib.

I'm currently in the process of investigating feasibility and already looking at what changes would be needed. I'm writing this issue to 1) see if there is interest in accepting such a contribution and 2) getting some feedback on the implementation approach.

I've added my implementation plan/idea below, I welcome comments about this. One thing that I would like some specific feedback on, is whether it would be acceptable to add some virtual methods to the `SX126x` class, to allow overriding them with STM32WL-specific behavior, or if you have other ideas on how to achieve this behavior (see below on details on how I could imagin this).

The STM32WL55 is a dual-core (M4 & M0+) MCU (of which the Arduino core currently only uses the M4) paired up with a LoRa transceiver which is essentially an SX1261/SX1262 (it has the exact same SPI command interface, but has both the low-power amplifier of the SX1261 and the high power amplifier of the SX1262). This means that most of the SX126x code should be usable without modification, though some changes are needed (mostly related to power selection or special handling of the GPIO and interrrupt connection).

### New subclass
To support the radio, a new `STM32WL` subclass of `SX126x` can be
created. There is no point in subclassing either `SX1261` or `SX1262`,
since most if not all of that code must be overridden anyway.

### Board-specific vs chip-specific
One point to note is that most customizations needed are specific to the
way the STM32WL is wired, but some things are specific to the way the
Nucleo board used is built.

The fact that an STM32WL chip is used can easily be detected using the
various macros set by the STM32Duino core (based on the board selected
for compilation), so that is known at compiletime.

What board is selected can also be detected using macros, so the code
can contain some code mapping board macros to board-specific settings,
but it is probably good to also allow these settings to be specified by
the sketch to support custom boards as well without having to modify the
library for such custom boars (configuration from the sketch is the
approach used for other radio chips as well).

### Board-specific configuration

 - For DCDC and TCXO configuration, the `STM32WL` subclass can provide
   appropriate defaults to the corresponding parameters to `begin()`,
   passing these values to the superclass. These defaults can based on
   the board macros defined by the stm32duino core/variant). By
   keeping the parameters available, this also allows supporting other
   boards with the same CPU by passing different values.

 - For TX power configuration, the sketch must be able to configure
[...]

PR #648, opened 2022-12-20
[SX126x] Fix comments & Dio2 for FSK

matthijskooijman:
These are two small issues I noticed while working on #588 that were separate from that issue, so best put into their own PR. These are fairly trivial fixes, I think. See commit messages for details.


jgromes:
Looking good, thanks!

PR #649, opened 2022-12-20
Add support for STM32WL integrated transceivers

### LINKED ISSUE #588: Adding support for STM32WL integrated transceivers

matthijskooijman:
This fixes #588 by adding support for STM32WL transceivers.

As discussed in that issue, this adds a new radio class to support the radio integrated into the STM32WL series of microcontrollers. The implementation is nearly complete, but there are a couple of things that need some additional review and some changes before they are ready to merge. Those commits have been marked with "Draft", and details of what is needed are given in the commit messages.

This PR is structured as follows:
 - The first three commits generalize the RfSwitch handling to support more than two pins and more different operation modes.
 - Then one commit that prepares SX126x for adding a new subclass (this splits off one change, some other smaller changes to SX126x are included in the STM32WLx commit later on).
 - The rest of the commits add the STM32WLx implementation. This starts with a big commit and then some smaller ones, that could probably have been a single commit but I hope that splitting them like this makes everything easier to review (and we can consider squashing them later, or leaving them split for the commit history).

~~This PR relies on some changes in the Arduino Core that are not yet merged. To compile the code, you need a git version of the core with https://github.com/stm32duino/Arduino_Core_STM32/pull/1839 applied.~~

*Edit: This PR needs version 2.4.0 of the STM32 Arduino core at https://github.com/stm32duino/Arduino_Core_STM32*

This PR contains some things that were not discussed in #588 yet, or implemented differently:
 - The class name is now `STM32WLx`, because `STM32WL` was already defined as a macro somewhere. It derives from `SX1262` instead of `SX126x`, since that allowed reusing some code after all.
 - The RFSwitch table code was done in a (hopefully) more portable way (not relying on pin numbers and states being integers) and without templates. It is now limited to 3 pins (but could be extended in the future with some work) and an unlimited amount of modes.
 - I've added an `STM32WLx_Module` class, which is a subclass of `Module` that handles overriding some of the callbacks and neatly hides all the (virtual) pin numbers and SPI configuration from the user (since the radio is internally hardwired, there is no need for the user/sketch to configure any of this).

I've tested basic transmission and reception, both regular and using interrupts on a Nucleo WL55JC1 board. This PR duplicates these four examples from SX126x for STM32WLx, since the initialization needs some significant changes. I haven't modified *all* examples, though, since that would result in more duplicate code (but I guess that duplicating the other five examples might be useful, since I'll need to test these things anyway, at least CAD and FSK, maybe also PingPong and Settings). I have tested interoperability with other radios early in the implementation (using an SX127x radio), I'll doublecheck that once the implementation is further finalized.

As shown in the new files' headers, copyright for this contribution is assigned to STMicroelectronics, but made available under the MIT license matching the existing license of this library.

Here's a bit of documentation that was removed from the code during review of the PR, but might be good to put on the wiki or elsewhere later:

#### RF Switch and low-power vs high-power transmissions

One notable difference with the %SX126x radios is that this radio
essentially combines the %SX1261 and %SX1262 by integrating both the
low-power (LP) and high-power (HP) amplifier. Because both output on
two different pins, the external RF switch on the board must be told
(through GPIO pins) which mode is used so it can connect the proper
output (or input, for RX) to the antenna.

Because of this, and because the DIO2 pin (which can be used on
external SX126x radios to let the radio handle RF switch control) is
not available externally on the STM32WL, you *must* configure an
RfSwitch table when using this radio.

Whether to use the LP or HP mode is automatically decided by the
setOutputPower() method, see its documentation for details.

Some boards might only have circuitry for just the low-power or
high-power transmission paths. For these boards, just omit the
unsupported mode from the table and setOutputPower will adjust its
behavior accordingly.

See Module::setRfSwitchTable() for general info on how to call this
method.  Note that because of the LP vs HP output, this radio defines
its own OpMode_t constants, which split the TX mode into two versions
and the example in the Module::setRfSwitchTable() method documentation
does not apply.

Instead, consider this example that sets up the RfSwitch as needed for
the Nucleo WL55JC1 board:

```C++
// In global scope, define the pin array and mode table
static const RADIOLIB_PIN_TYPE rfswitch_pins[] =
                         {PC3,  PC4,  PC5};
static const Module::RfSwitchMode_t rfswitch_table[] = {
  {STM32WLx::MODE_IDLE,  {LOW,  LOW,  LOW}},
  {STM32WLx::MODE_RX,    {HIGH, HIGH, LOW}},
  {STM32WLx::MODE_TX_LP, {HIGH, HIGH, HIGH}},
  {STM32WLx::MODE_TX_HP, {HIGH, LOW,  HIGH}},
  STM32WLx::END_OF_MODE_TABLE,
};

void setup() {
  ...
  // Then in setup (before calling begin), pass them to radiolib
  radio.setRfSwitchTable(rfswitch_pins, rfswitch_table);
  ...
}
```

Note that the setRfSwitchTable() method should be called *before* the
begin() method, to ensure the radio knows which modes are supported
during initialization.


jgromes:
This looks very good!

Regarding the examples, I'm fine with just leaving the Rx/Tx ones. CAD could be useful as well, but FSK and Settings, I don't think those are needed unless the behavior of the integrated transceiver differs significantly in this regard.

One thing that will still need to be set up is CI (it doesn't make sense to built the STM32WL examples on other platforms), but that can be taken care of by using a pattern that will skip all examples named STM32WL*. The same mechanism was used in previous versions of the library e.g. for HTTP on ESP32 (before that was dropped). It is probably better if I handle that.

I'll review and let you know. Is there an estiamte when a new release of the STM32 core with this support can be made available?

matthijskooijman:
> This looks very good!

Thanks!

> Regarding the examples, I'm fine with just leaving the Rx/Tx ones. CAD could be useful as well, but FSK and Settings, I don't think those are needed unless the behavior of the integrated transceiver differs significantly in this regard.

Ah, I just noticed that FSK and Settings are really just for documentation, and not full examples, so no real point in copying those indeed. I'll copy the CAD examples in a future push, because I at least want to test that they work :-)

> One thing that will still need to be set up is CI (it doesn't make sense to built the STM32WL examples on other platforms), but that can be taken care of by using a pattern that will skip all examples named STM32WL*. The same mechanism was used in previous versions of the library e.g. for HTTP on ESP32 (before that was dropped). It is probably better if I handle that.

Ah, I see. I just pushed a commit attempting to handle that, not sure if this is what you had in mind? 

> I'll review and let you know. Is there an estimate when a new release of the STM32 core with this support can be made available?

@fpistm is finalizing the review now, so should be somewhere in the coming days.

fpistm:
Hi @jgromes 

> Is there an estimate when a new release of the STM32 core with this support can be made available?

Yes, it is available :wink:
https://github.com/stm32duino/Arduino_Core_STM32/releases/tag/2.4.0

src/modules/SX126x/STM32WLx_Module.h
```diff
@@ -0,0 +1,52 @@
...
+
+  public:
+    STM32WLx_Module();
+
+#if !defined(RADIOLIB_GODMODE)
+  private:
+#endif
+
+    // TODO: Do we need copy constructor or assignment operator? Or mabye
+    // delete them? Module has defined them, but where are they used?
```
jgromes, line 40: They are not strictly necessary and are not used in the library itself or in the examples. However, if the user needs them for whatever reason, they are provided for completeness sake.

src/modules/SX126x/SX1262.cpp
```diff
@@ -20,10 +20,10 @@ int16_t SX1262::begin(float freq, float bw, uint8_t sf, uint8_t cr, uint8_t sync
   state = setFrequency(freq);
   RADIOLIB_ASSERT(state);
 
-  state = setOutputPower(power);
+  state = SX126x::fixPaClamping();
```
jgromes, line 23: What's the reason for swapping the order here?

src/modules/SX126x/SX126x.cpp
```diff
@@ -1581,7 +1585,11 @@ int16_t SX126x::fixPaClamping() {
   RADIOLIB_ASSERT(state);
 
   // update with the new value
-  clampConfig |= 0x1E;
+  if (enable)
+    clampConfig |= 0x1E;
```
jgromes, line 1589: Where are these numbers coming from?

src/Module.h
```diff
@@ -15,6 +15,43 @@
 */
 class Module {
   public:
+    /*!
+     * Description of RF switch pin states for a single mode.
+     *
+     * See setRfSwitchTable() for details.
+     */
+    struct RfSwitchMode_t {
+      unsigned mode;
```
jgromes, line 24: Please use fully specified types (e.g. uint8_t).

src/Module.cpp
```diff
@@ -551,20 +551,49 @@ void Module::regdump(uint8_t start, uint8_t len) {
...
+    {MODE_TX,    {LOW,  HIGH}},
+    END_OF_MODE_TABLE,
+  };
+  setRfSwitchTable(pins, table);
+}
+
+void Module::setRfSwitchTable(const RADIOLIB_PIN_TYPE (&pins)[3], const RfSwitchMode_t table[]) {
+  memcpy(_rfSwitchPins, pins, sizeof(_rfSwitchPins));
+  _rfSwitchTable = table;
+  for (auto pin : _rfSwitchPins)
```
jgromes, line 571: I would rather avoid using range loops since that requires C++11. It should be pretty simple to replace this with just a for loop (if number of pins is defined as macro, see one of the next comments)

src/modules/SX126x/SX126x.h
```diff
@@ -1047,6 +1043,9 @@ class SX126x: public PhysicalLayer {
 
     int16_t _lastError = RADIOLIB_ERR_NONE;
 
+    // Allow subclasses to define different TX modes
+    unsigned _tx_mode = Module::MODE_TX;
```
jgromes, line 1047: Please use fully specified types (e.g. uint8_t).

src/modules/SX126x/STM32WLx_Module.cpp
```diff
@@ -0,0 +1,114 @@
...
+  switch(dwPin) {
+    case RADIOLIB_STM32WLx_VIRTUAL_PIN_NSS:
+    case RADIOLIB_STM32WLx_VIRTUAL_PIN_BUSY:
+    case RADIOLIB_STM32WLx_VIRTUAL_PIN_IRQ:
+    case RADIOLIB_STM32WLx_VIRTUAL_PIN_RESET:
+      // Nothing to do
+      break;
+    default:
+      // TODO: Should we save the original callbacks in the constructor and then call those
+      // instead of the real Arduino functions directly?
```
jgromes, line 51: It's probably not strictly necessary, unless you want to jump through all the hoops of generator macros in Module header. I don't think there's much chance of this being used outside of Arduino.

jgromes (changes requested):
@matthijskooijman I left a coupel of comments in the review - overall there's nothing major that I can see.

I also fixed the CI and slightly cleaned up the examples. Thinking about those, it might be better to move them from the SX126x folder to a new STM32WLx folder, just for clarity.

src/modules/SX126x/STM32WLx.h
```diff
@@ -0,0 +1,171 @@
...
+/*!
+  \class STM32WLx
+
+  \brief Derived class for STM32WL modules.
+
+  The radio integrated into these modules is essentially the same as the
+  Semtech %SX126x external radio chips, so most of the documentation for
+  those also applies here.
+
+  #### RF Switch and low-power vs high-power transmissions
```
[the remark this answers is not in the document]
matthijskooijman: The reason I put this here, is that it deviates from the normal RfSwitch documentation, which uses `Module::MODE_*` constants. On STM32WL, separate mode constants and an extra mode is available, so this has be to documented somewhere. Putting this in the `STM32WLx::setRfSwitchTable()` docs would make sense, except that method does not exist (it just inherits the SX126x version without chagnes), so putting it in the class header seemed to most obvious place (though I guess we could explicitly inherit it with a `using` directive, or a method that just calls its parent, and document that perhaps).

Moving this into the wiki does not seem right to me, since then the API documentation in the code is missing relevant details about how to use the rf switch and how to set up modules that only support either LP or HP. Maybe duplicating (instead of moving) some of that information in the wiki could be a good idea, though I don't see an obvious place for this directly, https://github.com/jgromes/RadioLib/wiki/Modules only has very concise info for different module types).

src/modules/SX126x/SX1262.cpp
matthijskooijman: From the commit message:

> This modifies SX1262::begin and beginFSK to call setOutputPower after
applying the workaround, so that can undo the workaround if needed.

More specifically: fixPaClamping applies the workaround unconditionally, and setOuputPower (overridden by `STM32WLx`) applies it only when needed, so the latter must run after the former so it can undo the workaround if needed.

Alternatives to this approach could be:
 1. Make fixPaClamping virtual and put the logic about whether to apply or not inside that method, instead of in setOutputPower.
 2. Let `STM32WLx` override the `begin` and `beginFSK`, duplicating its contents and omitting the call to  `fixPaClamping()`.

src/modules/SX126x/SX126x.cpp
matthijskooijman: See comment a few lines up:

```
  // fixes overly eager PA clamping
  // see SX1262/SX1268 datasheet, chapter 15 Known Limitations, section 15.2 for details
```

Which points to:

image

src/Module.h
matthijskooijman: I think I used `unsigned` because this takes an enum, which I think resolves to `unsigned`. But you are right that just using a `uint8_t` is probably better (if that doesn't produce warnings, but I expect not), we won't have so many modes anyway.

Note that I didn't define the variable as a `Module::OpMode_t`, since it can also hold e.g. `STM32WLx::OpMode_t`.

src/Module.h
```diff
@@ -242,23 +279,107 @@ class Module {
...
+
+      void setup() {
+        ...
+        // Then somewhere in setup, pass them to radiolib
+        radio.setRfSwitchTable(rfswitch_pins, rfswitch_table);
+        ...
+      }
+      \endcode
+
+      TODO: Should we put this 3 (also in other places) into a constant?
```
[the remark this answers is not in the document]
matthijskooijman: The danger I see with defining a macro or constant (say `RFSWITCH_NUM_PINS=3`) is that sketches might start using this, e.g.:

```
      // In global scope, define the pin array and mode table
      static const RADIOLIB_PIN_TYPE rfswitch_pins[RFSWITCH_NUM_PINS] =
                             {PA0,  PA1,  RADIOLIB_NC};
      static const Module::RfSwitchMode_t rfswitch_table[] = {
        {Module::MODE_IDLE,  {LOW,  LOW}},
        {Module::MODE_RX,    {HIGH, LOW}},
        {Module::MODE_TX,    {HIGH, HIGH}},
         Module::END_OF_MODE_TABLE,
      };
```

Then, if we later decide to change the value to 4, the above code will define an extra pin element, which is initialized to 0 (or maybe even undefined if it is a local variable and not a global, but I'm not 100% sure about this) and will thus cause RadioLib to unexpectedly write values to pin 0.

For clarity, the upgrade path I had in mind when we ever *do* need to support 4 pins, is to increase the size of `_rfSwitchPins` and then add an additional overload of `setRfSwtichTable`, so you will have two overloads:

    void setRfSwitchTable(const RADIOLIB_PIN_TYPE (&pins)[3], const RfSwitchMode_t table[]);
    void setRfSwitchTable(const RADIOLIB_PIN_TYPE (&pins)[4], const RfSwitchMode_t table[]);

The former can ensure that the unused element is set to `NC`, so existing sketches that define 3-element pin arrays will just work as before. But this only works when the array size in sketches does not increase, hence my hesitancy of adding a constant for this.

(note that in this case the `RfSwitchMode_t` type does need to change from 3 pins to 4 pins, enlarging all existing `rfswitch_table` variables in sketches automatically, but extra zeroes do not hurt there. This means that no compatibility handling is needed for the tables).


Maybe adding a good compromise would be to add a `MAX_RFSWITCH_PINS` constant, adding "MAX" to make it less likely that people will use it in sketches, and maybe add some comments to explicitly discourage this? I don't think we can make this a protected constant in the Module class, because it will be part of the API (in `RfSwitchMode_t` and probably also in the `setRfSwitchTable` argument type).

Also, you say "It would be better IMO to define the number of pins as a macro", does that mean you prefer using a macro over a `const`? I usually prefer `const`, since those adhere to normal scoping rules and give better compiler errors when names are accidentally reused elsewhere. Also, making this a `const` inside the module class might help a bit more to hide it from users even if it cannot be made protected.

matthijskooijman:
> @matthijskooijman I left a coupel of comments in the review - overall there's nothing major that I can see.

Cool, thanks. Also for the quick and solid review. So far it's been a pleasure collaborating with you on this project :-)

I've responded to some of your comments, the other (obvious) ones I'll fix later today.

> I also fixed the CI and slightly cleaned up the examples.

Ah, I do remember typing that filter, not sure why it didn't end up in the commit. Thanks for adding it :-)

> Thinking about those, it might be better to move them from the SX126x folder to a new STM32WLx folder, just for clarity.

Sounds reasonable, I'll do that.


I'll try to make further changes to the PR by adding new commits to simplify review, but I'm inclined to squash some things together later (e.g. I could squash in your commits into mine already) unless you prefer not to?

src/modules/SX126x/STM32WLx_Module.h
matthijskooijman: Ok, makes sense to have them. However, I see that if you do not define any, C++ will automatically define a [copy constructor]() and [copy assignment operator](https://en.cppreference.com/w/cpp/language/copy_assignment) (unless they are deleted, but I don't think that happens here), so I think I can just remove the comment and it will be ok.

However, looking at the copy assignment operator for Module, I see that it only copies a subset of the fields. I'm not sure if that's intentional, but I doubt that the copy assignment is at all useful right now. In particular:
 - None of the callbacks are copied. It seems these fields were added in 8111bc058fe5818082c347370a7af00bf1ff286f, but they were not added to the copy assignment constructor.
 - The `_spi`, `_spiSettings` and `_initInterface` fields are not copied. The former two were originally copied, but this was removed for some reason in 8111bc058fe5818082c347370a7af00bf1ff286f. The latter was I think never copied (but didn't look closely).
 - The `_rfSwitchPins` and `_rfSwitchTable` fields are not copied. I hadn't considered adding them, because the `_rxEn` and `_txEn` fields they replaced weren't copied either (it seems these fields were added after the copy operator was added).

I actually think that to be useful, these copy operators should really just copy *all* fields, right? Or are there fields that should be omitted? If not, I think they can just be omitted, relying on the compiler to generate default versions that just copy all fields . In C++11 you can declare them with `= default` to force generating the defaults, or (I think) throw an error if defaults are not possible (e.g. with reference or const members), might be useful to add that (in appropriate C++ preprocessor guards).

jgromes: You're right, the copy operators should copy all fields. I didn't know compiler generates them by default (though it does make sense actually, come to think about it).

In the scope of this PR, you can just remove the comment. I will also remove the copy and assignment constructors from Module class (since it seems that the reason they only copy some fields is that I simply forgot to add them to the constructors ... compiler should be smarter than human in this case).

src/modules/SX126x/SX126x.cpp
jgromes: I meant the rest of it (highlighted incorrect part, sorry):

```c
  // update with the new value
  if (enable)
    clampConfig |= 0x1E;
  else
    clampConfig = (clampConfig & ~0x1E) | 0x80;
```

This seems to imply that if set to false, it will undo the workaround. I don't understand why it sets the MSB though, the datasheet doesn't mention that.

src/Module.h
jgromes: I've just seen too many bugs caused by assumed bit width of `int` (most recently in the on-board storage service of an unnamed spacecraft) to be comfortable with it.

If implicit conversion to e.g. uint8_t works, then I'm fine with it. If not, then explicit casting should be done.

src/modules/SX126x/SX1262.cpp
jgromes: Thanks, in that case I think this is resolved (it just wasn't immediately clear to me why this was needed).

src/Module.h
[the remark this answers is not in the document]
jgromes: The compromise with naming the macro (or constant) MAX* and documenting it properly is a good solution. At the end of the day, most people will keep just using the 2-pin RF switches, which is handled by `Module::setRfSwitchPins`. If their RF setup needs more complex driving, then they are free to define their own switch table.

src/modules/SX126x/STM32WLx.h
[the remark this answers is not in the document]
jgromes: I think this is a bit more high-level description of how the STM32WL functions, not so much related to any single method in STM32WLx.

The API documentation is also missing information about other high-level things, such as "how do you send a message?" - that's what the examples are for. And for even more generic documentation (such as https://github.com/jgromes/RadioLib/wiki/High-power-Radio-Modules-Guide), there's the wiki. In general I would say that the purpose of the doxygen docs is to describe "how do I use this method", as opposed to "how do I use this device".

jgromes:
@matthijskooijman feel free to squash commits, that's fine.

src/modules/SX126x/SX126x.cpp
matthijskooijman: > This seems to imply that if set to false, it will undo the workaround. I don't understand why it sets the MSB though, the datasheet doesn't mention that.

Ah, I think I made mistake there, should have been 0x08 (which is "0100" for bits 4-1). I'll fix that.

src/Module.h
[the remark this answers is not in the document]
matthijskooijman: Resolved in 4e26a8a9e0f73f029b3efa3a6301d06bd7d3f29a

src/modules/SX126x/STM32WLx.h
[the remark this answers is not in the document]
matthijskooijman: > The API documentation is also missing information about other high-level things, such as "how do you send a message?"

This is true, but in this case a signficant part of the documentation is about how the API for *this* module differs from the generic / SX126x module, which is otherwise lost in the API docs.

What I did now was add documentation for the `STM32WLx::setRfSwitchTable()` method, focusing on the API and not the module workings. The documentation in the `STM32WLx` class header was removed, leaving just a small remark about the amplifier difference and which APIs this affects. I preserved the original comment in the first post of this PR for later use on e.g. the wiki if needed.

See 5cc28361ceed9c1c569dcb16bc34f584d0e77966.

src/modules/SX126x/STM32WLx_Module.h
matthijskooijman: Resolved in 02bd4293b7418210931bd40c543c4556e7984fcb.

src/modules/SX126x/STM32WLx_Module.cpp
matthijskooijman: Resolved in 02bd4293b7418210931bd40c543c4556e7984fcb.

src/modules/SX126x/SX126x.cpp
matthijskooijman: Resolved in feede0b2fce9207133be4af68267377fec7b3ffe (I also slightly clarified the comment).

src/Module.h
matthijskooijman: Resolved in a64ee65a11b104cec72ae28a512d875660c080a0.

src/modules/SX126x/SX126x.h
matthijskooijman: Resolved in a64ee65a11b104cec72ae28a512d875660c080a0.

src/Module.cpp
matthijskooijman: Resolved in 4e26a8a9e0f73f029b3efa3a6301d06bd7d3f29a.

matthijskooijman:
I've addressed all comments in new fixup commits and squashed some of your commits into previous commits, so I did a force push. I've also moved the examples into their own directory, but I already squased that into the original commits.

I've left the comments open for you to resolve once you're satisfied with the resolution.

Once everything is resolved, there's a few more changes I want to make to the PR:
 - Apply bd66422ec276828019faca7212d767740af16500 (use `\copydoc` for `setRfSwitchPins`) to all radios
 - Add `setRfSwitchTable` to all radios that currently have `setRfSwitchPins`.
 - Squash fixups and the verbatim example copy commit.

I also want to modify `setRfSwitchState` to return a status code (so you get an error when a mode is omitted from the table but is used, instead of the radio silently not working or maybe even be damaged), but that requires a bit more work on all callsites (which I'm planning to outsource to ChatGPT, but the scripting around that is turning into a project of its own, so that will take a bit more time).

matthijskooijman:
@jgromes, did you get a chance to look at my last changes yet? I don't want to put pressure, but I'm holding off the last couple of changes (which are mostly more of the same thing) until you have given your blessing on the current changes to prevent double work :-)

matthijskooijman:
I've pushed another update to this PR with further cleanups and remaining TODOs. As far as I'm concerned, it is now ready to be merged.

Apart from rebasing on top of master and squashing commits, I've made the following actual changes to the code compared to my previous push:
 - Applied commit 90b28d77 (Remove duplicated setRfSwitchPins documentation) to all radios
 - Applied commit 3779faf6 (Add setRfSwitchTable() wrapper methods) to all radios
 - Added CAD examples in commit e116a20d ([STM32WLx] Add examples)
 - Updated interrupt examples to remove `enableInterrupt` like in master in commit e116a20d ([STM32WLx] Add examples)
 - Fixed `STM32WLx::setOutputPower` in commit 5e47d944 ([STM32WLx] Add module for STM32WL MCUs with integrated radio). It would previously use the LP amplifier up to 15dBm, based on docs that said it should support up to 15dBm. In testing, I noticed that it did not work, since 15dBm actually requires special handling (configuring 14dBm and then using some different PA config values). I did not add the special handling, but just limited to 14dBm, just like the existing SX1261 code does.

While testing I also noticed some weirdness with the HP amplifier power configuration (changing the TX power on one board would not change the RSSI on another board in the way I would expect, while it worked ok for the LP amplifier), but I suspect that that might have been a problem in my measurement setup, or maybe an existing problem with how the HP/SX1262 amplifier is configured, since I observed the same problem with an external SX1262 transceiver on an Arduino Uno. I want to further investigate this when I have some more time, but I do not think this is really related to this PR, so it should not need to block this PR.

I previously also proposed modifying `setRfSwitchState` to return a value, but my chatGPT refactor script for that is taking a bit more work that anticipated. Since this just a nice addition that is not required in any way to use this PR, I suggest also not letting this block this PR, I'll finish this in the near future and submit a new PR for it.


With that, I think this PR should be mergeable as it is now.

thebentern:
This is pretty exciting! Over on the Meshtastic project we've had our eye on some of boards that utilize the STM32WLx and are looking forward to the RadioLib support. :-)

jgromes:
@matthijskooijman thanks for the additions, I'll try to review everything one last time over the next few days :)

jgromes (approved):
From my point of view, everything is looking fine to be merged.

@matthijskooijman I also want to thank you, personally - this is a major contribution to RadioLib, and it is obvious a lot of personal time and dedication went into it. So, thank you, this project wouldn't be what it is now without community support and contributions.

Also, big thanks to @fpistm and everyone at STMicroelectronics who helped with this as well!

matthijskooijman:
\o/ Thanks!

>  it is obvious a lot of personal time and dedication went into it. So, thank you, this project wouldn't be what it is now without community support and contributions.

As for the "personal time", I have to forward those thanks to ST who have been paying for my time working on this. For the dedication, I can take credit myself ;-)

In any case, I'd like to also thank you and previous contributors - by building on your previous work, we now have a very feature-rich library for point-to-point communication on STM32WL, much better than if we would have written something from scratch in the same time :-) Also, I very much enjoyed the collaboration so far, your dedication and technical skills made it a pleasure to work on this for me.

Looking at the future: I'm happy to stay involved with maintaining the STM32WL support, so if you run into problems in this area, feel free to tag me in issues and I'll try to chip in (no promises though - the ST project is finished and my personal time is always overbooked ;-p).

fpistm:
Thank you all :)

amirna2:
Thank you for this great contribution. Is this compatible with the STM32WLE55JC?
I purchased a couple of Seeed Studio E5 development boards for a project and I would like to first prototype it in the Arduino ecosystem and RadioLib.
https://www.seeedstudio.com/LoRa-E5-Dev-Kit-p-4868.html?queryID=7d286374878ee8330deed17fa1dd7440&objectID=4868&indexName=bazaar_retailer_products

acceleroto:
> Thank you for this great contribution. Is this compatible with the STM32WLE55JC?
> I purchased a couple of Seeed Studio E5 development boards for a project and I would like to first prototype it in the Arduino ecosystem and RadioLib.

I've been meaning to try this out with my Wio-E5 project, but I haven't made the time yet. Looking through the code, there might be a hiccup or two, but it seems like it should work.

I haven't verified this, but you might hit a speedbump if you set the E5's output power to 14 dBm. From my code browsing, 17, 20 or 22 seemed like they'd work though. Modifying the RFSwitch table might fix that though. Again, I haven't tried yet.

jgromes:
@amirna2 @acceleroto please keep discussions in separate threads, unless you have something specific to this PR. Thanks!

Issue #647, opened 2022-12-20
[SX1278] Stream transmit FSK mode doesn't work


notsleep196:
Hello!

I tried to run "Stream Transmit" example and faced the issue.
The output is: 
```
[SX1278] Initializing ... success!
[SX1278] Sending a very long packet ... Guru Meditation Error: Core  1 panic'ed (Interrupt wdt timeout on CPU1). 

Core  1 register dump:
PC      : 0x4008a38c  PS      : 0x00060935  A0      : 0x800895e2  A1      : 0x3ffbec1c  
A2      : 0x3ffb9138  A3      : 0x3ffb8890  A4      : 0x00000004  A5      : 0x00060923  
A6      : 0x00060923  A7      : 0x00000001  A8      : 0x3ffb8890  A9      : 0x00000018  
A10     : 0x3ffb8890  A11     : 0x00000018  A12     : 0x3ffc199c  A13     : 0x00060923  
A14     : 0x007bee88  A15     : 0x003fffff  SAR     : 0x00000019  EXCCAUSE: 0x00000006  
EXCVADDR: 0x00000000  LBEG    : 0x40083cd5  LEND    : 0x40083cdd  LCOUNT  : 0x00000027  
Core  1 was running in ISR context:
EPC1    : 0x400dc21b  EPC2    : 0x00000000  EPC3    : 0x00000000  EPC4    : 0x00000000


Backtrace:0x4008a389:0x3ffbec1c |<-CORRUPTED


Core  0 register dump:
PC      : 0x4008a50f  PS      : 0x00060035  A0      : 0x8008920b  A1      : 0x3ffbe7dc  
A2      : 0x3ffbee88  A3      : 0xb33fffff  A4      : 0x0000abab  A5      : 0x00060023  
A6      : 0x00060021  A7      : 0x0000cdcd  A8      : 0x0000abab  A9      : 0xffffffff  
A10     : 0x00000000  A11     : 0x00000000  A12     : 0x3ffc17a0  A13     : 0x00000007  
A14     : 0x007bee88  A15     : 0x003fffff  SAR     : 0x0000001a  EXCCAUSE: 0x00000006  
EXCVADDR: 0x00000000  LBEG    : 0x00000000  LEND    : 0x00000000  LCOUNT  : 0x00000000  


Backtrace:0x4008a50c:0x3ffbe7dc |<-CORRUPTED




ELF file SHA256: 0000000000000000

Rebooting...
```

[...]

PR #651, opened 2022-12-22
[SX127x] Fix FSK Stream mode TX and RX

notsleep196:
Hello!

As i described in [Issue #602 ](https://github.com/jgromes/RadioLib/issues/602) and [Issue #647 ](https://github.com/jgromes/RadioLib/issues/647) there are some problems with streams in FSK on SX127x modules. So i submitted a PR with fixes for problems i described in my issues.

List of changes:

1. Remove clearIrqFlags() method call in fifoAdd() and fifoGet() in  SX127x.cpp, because it causes the package to split into several parts (as i saw on RX side)
2. Remove adding of "magic" byte, because it is not necessary
3. Updated Stream_Transmit.ino example, because it caused wdt reset

### DIFF
--- a/examples/Stream/Stream_Transmit/Stream_Transmit.ino
+++ b/examples/Stream/Stream_Transmit/Stream_Transmit.ino
@@ -86,6 +86,8 @@ volatile bool transmittedFlag = false;
 // disable interrupt when it's not needed
 volatile bool enableInterrupt = true;
 
+bool isFifoEmpty = false;
+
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks, I'm especially glad you managed to get rid of that extra padding byte, that was bugging me. This fix should also be done to RF69 (as it handles very much like the SX127x series), but I will do that separately.

Thanks for the contribution!

PR #663, opened 2023-01-16
[SX126x] Add method to support SX126x Rx Boosted Gain mode

andrew-moroz:
This adds a method allowing SX126x radios to be toggled between Rx Boosted Gain and Rx Power Saving Gain (the default setting) as described in section 9.6 of the SX126x datasheets. It also optionally allows the gain mode setting to be persisted across sleep/warm start.

I wasn't sure if perhaps it would be preferable to break the 'retain' argument (whether to persist the selected Rx gain setting) out into its own setter. I stopped short of doing that here, partly because the datasheet does not say anything specific about how to revert to _not_ persisting (though the implication seems to be that if the gain retention registers are set to any values other than those specified, the gain setting will not be retained). But, happy to refactor if needed.

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -870,6 +870,26 @@ int16_t SX126x::setRxBandwidth(float rxBw) {
   return(setModulationParamsFSK(_br, _pulseShape, _rxBw, _freqDev));
 }
 
+int16_t SX126x::setRxBoostedGainMode(bool rxbgm) {
+  // read the current register value
+  uint8_t rxGain = 0;
+  int16_t state = readRegister(RADIOLIB_SX126X_REG_RX_GAIN, &rxGain, 1);
+  RADIOLIB_ASSERT(state);
+
+  // gain mode register value (SX1261/2 datasheet v2.1 section 9.6)
+  if(rxbgm) {
+    rxGain = 0x96; // Rx Boosted Gain
+  } else {
+    rxGain = 0x94; // Rx Power Saving Gain
+  }
+
+  // update RX gain setting register
+  state = writeRegister(RADIOLIB_SX126X_REG_RX_GAIN, &rxGain, 1);
+  RADIOLIB_ASSERT(state);
+
+  return(state);
+}
+
 int16_t SX126x::setDataShaping(uint8_t sh) {
   // check active modem
   if(getPacketType() != RADIOLIB_SX126X_PACKET_TYPE_GFSK) {
--- a/src/modules/SX126x/SX126x.h
+++ b/src/modules/SX126x/SX126x.h
@@ -709,6 +709,15 @@ class SX126x: public PhysicalLayer {
     int16_t setRxBandwidth(float rxBw);
 
     /*!
+      \brief Enables or disables Rx Boosted Gain mode as described in SX126x datasheet section 9.6 (SX1261/2 v2.1, SX1268 v1.1)
+
+      \param rxbgm True for Rx Boosted Gain, false for Rx Power Saving Gain
+
+      \returns \ref status_codes
+    */
+    int16_t setRxBoostedGainMode(bool rxbgm);
+
+    /*!
       \brief Sets time-bandwidth product of Gaussian filter applied for shaping.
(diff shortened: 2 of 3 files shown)


src/modules/SX126x/SX126x.cpp
```diff
@@ -870,6 +870,41 @@ int16_t SX126x::setRxBandwidth(float rxBw) {
...
+    rxGain = 0x94; // Rx Power Saving Gain
+  }
+
+  // update RX gain setting register
+  state = writeRegister(RADIOLIB_SX126X_REG_RX_GAIN, &rxGain, 1);
+  RADIOLIB_ASSERT(state);
+
+  // add Rx Gain register to retention memory if requested
+  if(retain) {
+    uint8_t value0 = 0x01;
```
jgromes, line 891: Please add comment stating where these contants come from (e.g. SX126x datasheet vX.Y section 9.6).

src/modules/SX126x/SX126x.h
```diff
@@ -708,6 +708,16 @@ class SX126x: public PhysicalLayer {
     */
     int16_t setRxBandwidth(float rxBw);
 
+    /*!
+      \brief Enables or disables Rx Boosted Gain mode as described in SX126x datasheet section 9.6 (SX1261/2 v2.1, SX1268 v1.1)
+
+      \param rxbgm True for Rx Boosted Gain, false for Rx Power Saving Gain
+      \param retain True to persist gain setting through sleep/wake cycles
+
+      \returns 0
```
jgromes, line 717: Change to `\returns \ref status_codes`

src/modules/SX126x/SX126x.h
```diff
@@ -708,6 +708,16 @@ class SX126x: public PhysicalLayer {
...
+    */
+    int16_t setRxBoostedGainMode(bool rxbgm = false, bool retain = false);
```
jgromes, line 719: I would suggest to remove the first default argument - it doesn't seem to make much sense to call `setRxBoostedGainMode()`, why is `false` the default?

jgromes (changes requested):
Thank you for the contribution! I left some minor comments. In addition, please add `setRxBoostedGainMode` as keyword to keywords.txt (keep in mind this file has to use true tabs, not spaces).

Regarding restoring non-retention of Rx gain, did you investigate what is the default value of the `RADIOLIB_SX126X_REG_RX_GAIN_RETENTION_X` registers? And does it change upon power-on reset?

src/modules/SX126x/SX126x.h
andrew-moroz: Ah yep agreed, defaulting this to false does not make sense – removed.

andrew-moroz:
Awesome, thanks for the feedback! Made the changes you requested.

Re: retention of Rx gain setting, I just did some more testing locally. I observed that the value of the first retention register is always 0 on startup, but the values of the other two fluctuate and do not seem to have consistent defaults. But moreover, I was actually not able to get Rx gain retention working at all with my test device (a RAK4631 running Meshtastic 2.0.13). Upon sleep/warm start, the value in the Rx gain setting register is always 0x94, regardless of whether the retention registers were set. (I guess this might be a result of particularities in the way Meshtastic handles sleep, but it is also quite possible I am misunderstanding something in the datasheet or code.)

In light of the above, maybe it would be better to keep things simple and drop the 'retain' argument and functionality for now?

jgromes:
> I was actually not able to get Rx gain retention working at all with my test device

In that case, let's drop it - I would rather not have this functionality, than have something that maybe works, maybe doesn't and might have some unknown side effects.

You could try asking on the LoRa developer forum (https://forum.lora-developers.semtech.com/c/Semtech-Transceivers/SX126x-family/29), maybe somebody from Semtech will be kind enough to help.

andrew-moroz:
Ok sounds good, removed 'retain' arg and functionality.

jgromes (approved):
Thank you, in that case this is good to merge.

jgromes:
... and done! Thank you very much for the contribution!

tckb:
I know this PR is old and done. @andrew-moroz  what is the purpose of this method other whats on the method? in other words, where and how enabling this option (i.e., gain mode) would increase the RX sensitivity? I am currently looking for ways to increase sensitivity (apart from working on LNA and antenna setup) thank you!

PR #674, opened 2023-01-28
sx126x-rx-gain-persist: add ability to persist Rx gain setting when calling SX126x::setRxBoostedGainMode

andrew-moroz:
This restores the 'persist' functionality when setting Rx boosted gain mode on SX126x devices. I had previously thought this wasn't working, but then realized I had misunderstood how it is intended to work. Just verified that setting these three registers as described in the datasheet _does_ indeed persist the gain setting when using Rx duty cycle mode, where otherwise it immediately reverts to its default power-saving gain setting upon the first switch to Tx and back to Rx.

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -870,7 +870,7 @@ int16_t SX126x::setRxBandwidth(float rxBw) {
   return(setModulationParamsFSK(_br, _pulseShape, _rxBw, _freqDev));
 }
 
-int16_t SX126x::setRxBoostedGainMode(bool rxbgm) {
+int16_t SX126x::setRxBoostedGainMode(bool rxbgm, bool persist) {
   // read the current register value
   uint8_t rxGain = 0;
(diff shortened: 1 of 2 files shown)


jgromes:
Thank you for looking into this further. So how exactly does the Rx gain retention work?

jgromes:
Merged, thanks for the contribution!

andrew-moroz:
Awesome, thanks! Btw, I did find that zeroing out the gain retention registers makes the module stop persisting Rx gain mode. So I had considered doing this when `persist` is false, so that e.g. calling `setRxBoostedGainMode(false, false)` would end persistence. This seems like an edge use case, so not sure whether it's overkill. In any case I am a bit reticent to spam you with 3 incremental PRs for what could have been done in one – but will happily submit an update if you feel it would make the feature more complete 🙂

jgromes:
As a rule of thumb, there should be a way to explicitly revert any configuration. However, this seems to me like an undocumented behavior with possible side effects, so in this case I would rather make an exception.

Issue #683, opened 2023-02-27
[SX1262]: Chip not correctly recognized


GUVWAF:
I'm using the latest commit 00998b5 with Meshtastic firmware 2.0.22. It seems the `SX126x::findChip()` function is recognizing my module as an SX1261, although according to the [datasheet](https://docs.rakwireless.com/Product-Categories/WisBlock/RAK4631/Datasheet/#features), it is a SX1262 using nRF52840. Here is the full SPI transcript when starting up:

> CMD	80	
DATW	0	AA	

> CMD	80	
DATW	0	A2	

> CMD	C0	
DATR	0	22	0	22	

> CMD	1D	3	20	
DATR	0	A2	0	53	0	58	0	31	0	32	0	36	0	31	0	20	0	56	0	32	0	44	0	20	0	32	0	44	0	30	0	32	0	0	

> SX126x not found! (1 of 10 tries) RADIOLIB_SX126X_REG_VERSION_STRING:
0000320 53 58 31 32 36 31 20 56 32 44 20 32 44 30 32 00 | SX1261 V2D 2D02.

This then repeats 10 times, after it prints “No SX126x found!”, and the result of function `begin` is `RADIOLIB_ERR_CHIP_NOT_FOUND`.

As expected, the module cannot be correctly operated after this.

jgromes:
I also have a module that the manufacturer (Dorji RF) claims is SX1262. However, the device reports as SX1261. Tomorrow I will remove the shield and check, but it seems to me that the silicon should be trusted in case of discrepancy. Does it work when you change your code to SX1261?

GUVWAF:
Yes, changing to SX1261 works. Indeed, I would expect the chip is not lying.
It would be interesting though if it is really a SX1261, as it was set to 22dBm output power, while the SX1261 can only do 15dBm.

jgromes:
I asked a question about the device ID registers on the LoRa dev forum (https://forum.lora-developers.semtech.com/t/sx126x-device-id/1508), hopefully someone from Semtech can clarify. I'm not sure if the device would actively reject an out-of-bound output power request. RadioLib only performs range checking based on information from the user (the wireless module type).

GUVWAF:
Thanks. I am curious to see whether you get an answer on it, as I found a while ago the answer was as follows: https://forum.lora-developers.semtech.com/t/detect-transceiver-version/1377

Rustie0125:
[...]

PR #684, opened 2023-02-28
[SX126x/SX128x] Add user-configurable IRQ flags and mask to receive methods

GUVWAF:
As discussed in #682, this adds two parameters to the interrupt receive methods of SX126x and SX128x modules. In this way, a user can set the IRQ flags and mask (the IRQs that will trigger DIO1) as they wish. 

For convenience and readability, I added the `RADIOLIB_SX126X_IRQ_RX_DEFAULT` definition, which is the combination of RX_DONE, TIMEOUT, CRC_ERROR and HEADER_ERROR. The `irqFlags` default to this, while `irqMask` defaults to RX_DONE, like it was up until now.

I tested with Meshtastic firmware 2.0.22 using the SX1261 module (although it _should_ be an SX1262 as discussed in #683) that without giving these parameters, it behaves as before. When adding the HEADER_VALID flag, it can successfully detect using `getIrqStatus()` that it is actively receiving, for both `startReceive()` and `startReceiveDutyCycleAuto()`. Since I do not have an SX128x module, I didn’t test this hardware.

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -532,8 +532,8 @@ int16_t SX126x::finishTransmit() {
   return(standby());
 }
 
-int16_t SX126x::startReceive(uint32_t timeout) {
-  int16_t state = startReceiveCommon(timeout);
+int16_t SX126x::startReceive(uint32_t timeout, uint16_t irqFlags, uint16_t irqMask) {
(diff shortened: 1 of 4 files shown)


jgromes:
Thanks for the contribution! Just to let you know I didn't forget about this, I just haven't had the time yet to test it. Hopefully over the weekend.

jgromes:
Everything looks good to be merged, thanks for the contribution!

PR #700, opened 2023-03-16
add functions to PhysicalLayer interface

gasagna:
Hi, following our discussion I have added a few functions to improve the `PhysicalLayer` interface. Surely there are others missing, but these are those i need for now!

### DIFF
--- a/src/protocols/PhysicalLayer/PhysicalLayer.cpp
+++ b/src/protocols/PhysicalLayer/PhysicalLayer.cpp
@@ -110,12 +110,23 @@ int16_t PhysicalLayer::receive(uint8_t* data, size_t len) {
   return(RADIOLIB_ERR_UNSUPPORTED);
 }
 
+int16_t PhysicalLayer::sleep() {
+  return(RADIOLIB_ERR_UNSUPPORTED);
+}
+
 int16_t PhysicalLayer::standby() {
   return(standby(RADIOLIB_STANDBY_DEFAULT));
 }
 
 int16_t PhysicalLayer::standby(uint8_t mode) {
   (void)mode;
+  return(RADIOLIB_ERR_UNSUPPORTED);
+}
+
+int16_t PhysicalLayer::startReceive(uint32_t timeout, uint16_t irqFlags, uint16_t irqMask) {
+  (void)timeout;
+  (void)irqFlags;
+  (void)irqMask;
   return(RADIOLIB_ERR_UNSUPPORTED);
 }
 
@@ -228,6 +239,14 @@ float PhysicalLayer::getFreqStep() const {
 size_t PhysicalLayer::getPacketLength(bool update) {
   (void)update;
   return(0);
+}
+
+float PhysicalLayer::getRSSI() {
+  return(RADIOLIB_ERR_UNSUPPORTED);
+}
+
+float PhysicalLayer::getSNR() {
+  return(RADIOLIB_ERR_UNSUPPORTED);
 }
 
 int32_t PhysicalLayer::random(int32_t max) {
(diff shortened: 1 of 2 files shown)


jgromes:
Everything looks OK, thanks for the contribution. Feel free to open another PR in case you feel the need to add more methods.

gasagna:
@jgromes, I discovered a bug in my PR today, for the `startReceive` function implementation. I had to define default values for the three arguments, because I thought that subclasses would define their own. The signature in `PhysicalLayer.h` is 
```c++
virtual int16_t startReceive(uint32_t timeout = 0, uint16_t irqFlags = 0, uint16_t irqMask = 0);
```
while, for instance, in `SX126X.h` this is the signature:
```c++
int16_t startReceive(uint32_t timeout = RADIOLIB_SX126X_RX_TIMEOUT_INF, uint16_t irqFlags = RADIOLIB_SX126X_IRQ_RX_DEFAULT, uint16_t irqMask = RADIOLIB_SX126X_IRQ_RX_DONE);
```

However, calling `startReceive(RADIOLIB_SX126X_RX_TIMEOUT_INF)` on a `SX1262` object does not seem to pick up the correct default values for the `irqFlags` and `irqMask` arguments, and it looks like the compiler is picking up the default values in the declaration in `PhysicalLayer.h` Did I forget an `override`?

jgromes:
@gasagna I was not able to reproduce this - when I call `startReceive(RADIOLIB_SX126X_RX_TIMEOUT_INF)` and print the value of e.g. `irqFlags`, I get 0x262, which is the correct default value.

gasagna:
Thanks, where do you read the `irqFlags` from?

jgromes:
I see - I was able to replicate this using a simple wrapper class:
```c++
class Wrapper {
  public:
    PhysicalLayer& _phy;
    Wrapper(PhysicalLayer& phy) : _phy(phy) { }
    int16_t startReceive(uint32_t timeout) {
      Serial.print("Wrapper::startReceive, timeout=");
      Serial.println(timeout);
      _phy.startReceive(timeout);
      return(0);
    } 
};

(...)

Wrapper w(radio);
w.startReceive(10);
```

This indeed produces the following:

```
Wrapper::startReceive, timeout=10
SX126x::startReceive, irqFlags=0
```

It is very similar to how standby is done: add `SX126x::startReceive(uint32_t timeout = RADIOLIB_SX126X_RX_TIMEOUT_INF)` and remove the defaults from the original `SX126x::startReceive`. Then the implementation of `startReceive` with a single timeout argument must call the full version `startReceive(timeout, RADIOLIB_SX126X_IRQ_RX_DEFAULT, RADIOLIB_SX126X_IRQ_RX_DONE)`.

However, there's a bigger issue - the `PhysicalLayer::startReceive` method has to be common for all modules. CC1101, nRF24, RF69 and Si443x have `startReceive` without arguments. SX127x has a `mode` and `len` arguments, and SX126x/8x have the timeout and IRQ flags/mask. For the implementation in `PhysicalLayer` to make sense, it needs to be comon for all the modules.

gasagna:
Thanks for confirming this. Not sure how to best address it. The question is if you'd like `PhysicalLayer` to be a complete base class for all the other radio drivers. I remember you have discussed this in the context of a LoRaWAN implementation in some issue in this repo.

jgromes:
I think it would be best if we turn `PhysicalLayer` into a complete base class. It started as a collection of useful methods and abstraction for ham radio modes, but since then it clearly evolved.

Since this is will take some architectural decisions, I will do it myself and let you know once it's implemented.

gasagna:
Thanks!

jgromes:
@gasagna startReceive compatibility across different modules should be fixed now, building a common baseline is going to take a bit more time.

PR #711, opened 2023-03-25
[SX1278] New getInstRSSI to get current RSSI

G4lile0:
This idea came from an user of tinyGS , that suggest to check the current RSSI of the module to check the actual noise floor, I just added this feature to the tinyGS beta and it work great.  If it's ok for you I can create it as well for SX126x and SX128x modules. 
Thanks

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -1221,7 +1221,7 @@ float SX126x::getDataRate() const {
   return(_dataRate);
 }
 
-float SX126x::getRSSI() {
+float SX126x::getRSSI(bool packet) {
 
   if (packet) { 
   // get last packet RSSI from packet status
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -1222,10 +1222,20 @@ float SX126x::getDataRate() const {
 }
 
 float SX126x::getRSSI() {
+
+  if (packet) { 
   // get last packet RSSI from packet status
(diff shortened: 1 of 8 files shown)


jgromes:
Thanks, I'm OK with adding this functionality, however, the implementation seems suboptimal to me - it mostly copies the existing `SX1278::getRSSI`, the only difference is which register is used when reading the value in LoRa mode. For FSK, it also behaves exactly the same as the current `SX1278::getRSSI`, which could lead to confusion.

The best way to do it would be the following in my opinion:

1. Create a common `SX127x::getRSSI(bool packet, bool skipReceive, int16_t offset)` method. This is will implement the functionality that is currently split in `SX1272` and `SX1278`.
2. Add `SX1272::getRSSI(bool packet = true, bool skipReceive = false)` which just call the common SX127x method, but set the correct offset for LoRa RSSI.

For SX126x, there's already `SX126x::getRSSIInst`, but you are correct that it is missing in SX128x. To have a common interface, I would suggest `SX126x::getRSSI(bool packet = true)`

G4lile0:
Thank you Jan, this makes a lot of sense.

G4lile0:
Sure, let me send you a proposal.

G4lile0:
I wasn't sure for SX1272 modification, so I just modified SX1278 and SX126x.
I haven't delete getRSSIInst from SX126x to keep the compatibility .

jgromes:
Thanks, I took the liberty to unify the interface so that there's just one common `getRSSI` method. With this, I think we can merge, thanks for the contribution!

PR #726, opened 2023-04-10
Fix generic build

Mesteery:
I confirm that it works on generic build, but I didn't tested it on Arduino.

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -1044,10 +1044,10 @@
     #define RADIOLIB_DEBUG_PRINTLN(...) { RADIOLIB_DEBUG_PORT.println(__VA_ARGS__); }
   #else
     #if !defined(RADIOLIB_DEBUG_PRINT)
-      #define RADIOLIB_DEBUG_PRINT(...) { frintf(RADIOLIB_DEBUG_PORT, __VA_ARGS__); }
+      #define RADIOLIB_DEBUG_PRINT(...) { fprintf(RADIOLIB_DEBUG_PORT, __VA_ARGS__); }
     #endif
     #if !defined(RADIOLIB_DEBUG_PRINTLN)
-      #define RADIOLIB_DEBUG_PRINTLN(...)  { printf(RADIOLIB_DEBUG_PORT, __VA_ARGS__ "\n"); }
+      #define RADIOLIB_DEBUG_PRINTLN(...)  { fprintf(RADIOLIB_DEBUG_PORT, __VA_ARGS__ "\n"); }
     #endif
   #endif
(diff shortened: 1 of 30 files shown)


src/BuildOpt.h
```diff
@@ -1040,14 +1040,37 @@
...
+        if (!buffer) {
+            return 0;
+        }
+        va_start(arg, format);
+        vsnprintf(buffer, len + 1, format, arg);
+        va_end(arg);
+    }
+    len = RADIOLIB_DEBUG_PORT.write((const uint8_t*) buffer, len);
+    if (buffer != temp) {
+        delete[] buffer;
```
jgromes, line 1062: Why is buffer being deleted? It's allocated statically.

jgromes (commented):
Thanks, I left some comments. Nothin major, but there is ghoing to be some impact on lower-performance platforms with debug enabled.

src/BuildOpt.h
Mesteery: I took this implementation from esp8266 Arduino (cf. the link)

src/BuildOpt.h
```diff
@@ -1040,14 +1040,37 @@
...
+    }
+    return len;
+}
+    #define RADIOLIB_DEBUG_PRINT(...) serialPrintf(__VA_ARGS__)
```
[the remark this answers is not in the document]
Mesteery: > If a generic printf() function will be added, then it will be better to have it as a separate macro RADIOLIB_DEBUG_PRINTF instead of changing the behavior of the currently implemented PRINT/PRINTLN.

My idea was to fix the difference of the behaviour between Arduino's `RADIOLIB_DEBUG_PRINT` and the fallback's `RADIOLIB_DEBUG_PRINT`, Serial.print does not support format like sprintf, and sprintf can't takes other argument than a `const char`.

If we create a `RADIOLIB_DEBUG_PRINTF`, what PRINT/PRINTLN will become? And how we can ensure that they have the same behaviour no matter the build platform?

src/BuildOpt.h
jgromes: I didn't notice the dynamic allocation a couple lines above, in that case it's OK.

src/BuildOpt.h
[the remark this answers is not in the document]
jgromes: Hmm, you're right. In that case let's keep this new implementation.

jgromes (changes requested):
This can be merged once the open points are closed.

jgromes (approved):
Thanks, all looks good now!

PR #728, opened 2023-04-12
Undefine RADIOLIB_DEBUG_PRINT_BIN macro after use

Mesteery:
RADIOLIB_DEBUG_PRINT_BIN is used only once, and is defined in a specific scope.


jgromes:
Maybe it would be easier to not have bin debug print at all, and just use hex. The bin print is only applicable for Arduino Serial class, it's not really a thing for sprintf and other more common methods.

jgromes:
Merged, thanks!

PR #730, opened 2023-04-12
Improve hardware abstraction layer

Mesteery:
This PR completly refactor the hardware abstraction layer. I think it's ready to be reviewed and I'm really sorry because it's a big PR and I didn't split the commits to ease the review. 
It compiled successfully on arduino cores (https://github.com/Mesteery/RadioLib/actions/runs/4680264029 and https://github.com/Mesteery/RadioLib/actions/runs/4681321278/jobs/8293761734 after stm32wlx fix)

- I was forced to remove `dir` default in methods like `setGdo0Action` because of the use of `this` (to access eg. `hal->GpioInterruptRising`)
- I'm not too happy with the RADIOLIB_NC check in the HAL because I don't think it's normal that the user has to check this kind of thing (in case he provides his own HAL). But this is the simplest and cleanest way, so it doesn't bother me more than that

*This is a breaking change.*

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -221,7 +221,7 @@
 #endif
 
   #if !defined(RADIOLIB_NC)
-  #define RADIOLIB_NC                                 (0xFF)
+  #define RADIOLIB_NC                                 (0xFFFFFFFF)
   #endif
   #if !defined(RADIOLIB_DEFAULT_SPI)
   #define RADIOLIB_DEFAULT_SPI                        SPI
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -7,21 +7,16 @@
   #define RADIOLIB_BUILD_ARDUINO
 #else
   // generic build
+  #include <stdio.h>
   #define RADIOLIB_BUILD_GENERIC
 #endif
 
 #if defined(RADIOLIB_BUILD_ARDUINO)
-
 /*
  * Platform-specific configuration.
  *
  * RADIOLIB_PLATFORM - platform name, used in debugging to quickly check the correct platform is detected.
- * RADIOLIB_PIN_TYPE - which type should be used for pin numbers in functions like digitalRead().
- * RADIOLIB_PIN_MODE - which type should be used for pin modes in functions like pinMode().
- * RADIOLIB_PIN_STATUS - which type should be used for pin values in functions like digitalWrite().
- * RADIOLIB_INTERRUPT_STATUS - which type should be used for pin changes in functions like attachInterrupt().
- * RADIOLIB_DIGITAL_PIN_TO_INTERRUPT - function/macro to be used to convert digital pin number to interrupt pin number.
- * RADIOLIB_NC - alias for unused pin, usually the largest possible value of RADIOLIB_PIN_TYPE.
+ * RADIOLIB_NC - alias for unused pin, usually the largest possible value of uint8_t.
  * RADIOLIB_DEFAULT_SPI - default SPIClass instance to use.
  * RADIOLIB_NONVOLATILE - macro to place variable into program storage (usually Flash).
  * RADIOLIB_NONVOLATILE_READ_BYTE - function/macro to read variables saved in program storage (usually Flash).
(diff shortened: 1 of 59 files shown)


jgromes:
Thanks, is this ready for review or still in progress?

Mesteery:
It seems to be ready for review. I've tested it on my pi with real hardware, it also works!

Mesteery:
By the way, could we use directly `uint8_t` instead of `byte` in the lib, so we cam remove this line for generic build: https://github.com/jgromes/RadioLib/blob/7faeb46ada637675e4c96677de370dd773435bc4/src/BuildOpt.h#L276

jgromes:
As far as I can tell, there should be no usage of the `byte` type anywhere in the library, if there is then it's definitely a mistake. The only exception are the examples, which will remain Arduino-specific even after this is merged.

I'm a bit busy at the moment, but will try to do a proper review over the weekend.

src/BuildOpt.h
```diff
@@ -7,21 +7,16 @@
   #define RADIOLIB_BUILD_ARDUINO
 #else
   // generic build
+  #include <stdio.h>
```
jgromes, line 10: Why is this needed?

jgromes: Similar with other includes (ctypes, math.h etc.) in other files.

src/BuildOpt.h
```diff
@@ -105,833 +65,218 @@
...
+
+  #define DEC 10
+  #define HEX 16
+  #define OCT 8
+  #define BIN 2
+
+  #include <algorithm>
+  #include <stdint.h>
 
+  typedef uint8_t byte;
```
jgromes, line 276: As per the conversation in the PR, this can likely be removed.

src/Hal.h
```diff
@@ -0,0 +1,42 @@
...
+#define _RADIOLIB_HAL_H
+
+class Hal {
+  public:
+    const uint8_t GpioModeInput;
+    const uint8_t GpioModeOutput;
+    const uint8_t GpioLevelLow;
+    const uint8_t GpioLevelHigh;
+    const uint8_t GpioInterruptRising;
+    const uint8_t GpioInterruptFalling;
```
jgromes, line 13: I don't think it's safe to use uint8_t as the generic "pin type", I would propose to change this to uint32_t as most of the platforms RadioLib runs on are 32-bit.

src/protocols/ExternalRadio/ExternalRadio.h
```diff
@@ -3,12 +3,17 @@
 
 #include "../../TypeDef.h"
 #include "../../Module.h"
+#include "../../ArduinoHal.h"
```
jgromes, line 6: This should probably be included only for RADIOLIB_BUILD_ARDUINO.

src/Module.h
```diff
@@ -1,8 +1,8 @@
+#include "TypeDef.h"
+#include "Hal.h"
```
jgromes, line 2: Why is this moved before the guard ifdef?

src/Hal.cpp
```diff
@@ -0,0 +1,17 @@
...
+      GpioLevelHigh(high),
+      GpioInterruptRising(rising),
+      GpioInterruptFalling(falling) {}
+
+void Hal::init(){};
+void Hal::term(){};
+void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
+void Hal::noTone(uint8_t pin){};
+void Hal::yield(){};
+uint8_t Hal::pinToInterrupt(uint8_t pin) { return pin; };
```
jgromes, line 17: What is the reason these methods are not pure virtual like the rest?

jgromes (changes requested):
I left some comments, but it's nothing major. If everything can be implemented/answered, then I think this can be merged. Thanks!

jgromes:
Also, I noticed there's now a couple warnings in the CI (building for Arduino Uno), related to the (pure) virtual comment I left.

```
Building AFSK_Imperial_March.ino ... 
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp: In member function 'virtual void Hal::tone(uint8_t, unsigned int, long unsigned int)':
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:24: warning: unused parameter 'pin' [-Wunused-parameter]
 void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
                        ^~~
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:42: warning: unused parameter 'frequency' [-Wunused-parameter]
 void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
                                          ^~~~~~~~~
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:67: warning: unused parameter 'duration' [-Wunused-parameter]
 void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
                                                                   ^~~~~~~~
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp: In member function 'virtual void Hal::noTone(uint8_t)':
/home/runner/work/RadioLib/RadioLib/src/Hal.cpp:15:26: warning: unused parameter 'pin' [-Wunused-parameter]
 void Hal::noTone(uint8_t pin){};
                          ^~~
Sketch uses 20276 bytes (62%) of program storage space. Maximum is 32256 bytes.
Global variables use 990 bytes (48%) of dynamic memory, leaving 1058 bytes for local variables. Maximum is 2048 bytes.
```

src/BuildOpt.h
Mesteery: There indeed useless for Arduino build since they are already included, but on generic build they are required to use some functions like `sprinf`, `max`, `min`, `fabs`, etc.
These includes did not cause any problems with arduino, so I left them unconditioned at the top of the file using functions from them.

src/BuildOpt.h
Mesteery: Yes I forgot to remove it, since there is yet some files using `byte` in the lib.

src/protocols/ExternalRadio/ExternalRadio.h
Mesteery: ArduinoHal.h would be empty if RADIOLIB_BUILD_ARDUINO is empty, so is it important to re-check?

src/Module.h
Mesteery: I don't remember much about it, but maybe it's because some define were not found, or maybe it's completely by mistake. I remember at least one case like that but I don't know where. I'll try again.

src/Hal.cpp
Mesteery: They are not pure because on some platform they doesn't exist, and will simply be defined as nop. We can simply make them optional and implement them as not for platform that doesn't support them.
init and term are optional since they are `Hal`-specific and can or not be useful.

Mesteery:
> Also, I noticed there's now a couple warnings in the CI (building for Arduino Uno), related to the (pure) virtual comment I left.
> 
> ```
> Building AFSK_Imperial_March.ino ... 
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp: In member function 'virtual void Hal::tone(uint8_t, unsigned int, long unsigned int)':
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:24: warning: unused parameter 'pin' [-Wunused-parameter]
>  void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
>                         ^~~
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:42: warning: unused parameter 'frequency' [-Wunused-parameter]
>  void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
>                                           ^~~~~~~~~
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp:14:67: warning: unused parameter 'duration' [-Wunused-parameter]
>  void Hal::tone(uint8_t pin, unsigned int frequency, unsigned long duration){};
>                                                                    ^~~~~~~~
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp: In member function 'virtual void Hal::noTone(uint8_t)':
> /home/runner/work/RadioLib/RadioLib/src/Hal.cpp:15:26: warning: unused parameter 'pin' [-Wunused-parameter]
>  void Hal::noTone(uint8_t pin){};
>                           ^~~
> Sketch uses 20276 bytes (62%) of program storage space. Maximum is 32256 bytes.
> Global variables use 990 bytes (48%) of dynamic memory, leaving 1058 bytes for local variables. Maximum is 2048 bytes.
> ```

I left them because I find writing `(void)var;` not pretty like in some files to remove warning. But finally I will just do like this, it's better than useless warning :+1:

src/Hal.h
Mesteery: I don't understand , why wouldn't it be safe?

src/BuildOpt.h
```diff
@@ -105,833 +65,218 @@
...
+
+  #if !defined(RADIOLIB_ARDUINOHAL_PIN_MODE_CAST)
+  #define RADIOLIB_ARDUINOHAL_PIN_MODE_CAST
+  #endif
+  #if !defined(RADIOLIB_ARDUINOHAL_PIN_STATUS_CAST)
+  #define RADIOLIB_ARDUINOHAL_PIN_STATUS_CAST
+  #endif
+  #if !defined(RADIOLIB_ARDUINOHAL_INTERRUPT_MODE_CAST)
+  #define RADIOLIB_ARDUINOHAL_INTERRUPT_MODE_CAST
+  #endif
```
Mesteery, line 253: (Not important again) I wanted to know if it's not too "ugly" the repetition of `ifndef .. define .. endif`. It seems to be quite syntactically heavy but I haven't found an alternative.

src/BuildOpt.h
jgromes: > there is yet some files using byte in the lib.

Where exactly is this?

src/protocols/ExternalRadio/ExternalRadio.h
jgromes: At the moment, yes, it would be empty, but I would rather avoid this breaking in the future. Also IIRC there's other places where ArduinoHal.h is included which are guarded, so let's make it consistent. It's a compile-time check anyway.

jgromes:
> I left them because I find writing (void)var; not pretty like in some files to remove warning

I would say that in these cases it's justified - actually, the only times when (void)var; is used is when it's a parameters that is needed for some common API, which is also teh case here.

src/BuildOpt.h
jgromes: I would prefer it like this, if the user decides to define the macro in some other way (for example by calling gcc with -DRADIOLIB_ARDUINOHAL_INTERRUPT_MODE_CAST=something).

src/BuildOpt.h
jgromes: Thanks for the examplanation, in that case this can remain.

src/Hal.h
jgromes: > I don't think there is a platform with more than 255 pins

But the value can contain more information than just a pin number. In fact most MCUs address GPIO pins using port and pin type. IIRC some STM32 Arduino cores use the number to save both GPIO port as well as GPIO pin, which may add up to more than 8 bits.

src/Hal.cpp
jgromes: OK, thanks for the clarification, this can remain as-is.

src/BuildOpt.h
Mesteery: https://github.com/jgromes/RadioLib/blob/88549baf8326ee9f679e09e04f7b3c1a0e8d7a2f/src/protocols/Morse/Morse.h#L127 via gh code search.

jgromes: Ah, I see - that should be `uint8_t`, feel free to change it (or I can do after merging this).

src/BuildOpt.h
Mesteery: Yes sure, I was rather searching a different syntax or way to do that.

jgromes: As far as I know there's nothing else, I've seen this syntax quite a lot.

jgromes (approved):
All looking good now, thanks!

jgromes:
@Mesteery I was about to merge this, but then I realized it could also be useful to have an example of how to use the new HAL. Since you already have yours, maybe you could add it to e.g. `examples/Non_Arduino/...`? We would have to modify the CI file so that it doesn't try to build it as Arduino example, but that is pretty straight-forward (it has a filter).

Or we can merge this as-is and open another PR for the example, either way is fine if you're willing to share it of course.

jgromes:
Perfect, then I can try it right away using RPi and a Waveshare LoRaWAN hat!

Mesteery:
Do you have any news, I'm curious to know if it worked!

jgromes:
@Mesteery unfrotunately not yet, recently I haven't had enough time to put into this project :( Hoping to get this done over the weekend again, sorry it's taking longer.

jgromes:
@Mesteery I fainlly got around to merge this - thank you very much for the great contribution!

In the following days, I will do a bit of cleanup and then release this as 6.0.0.

PR #739, opened 2023-04-30
improve build workflow for better flexibility

Mesteery:
This makes the workflow lighter and more readable and allows to add more easily newer platforms in the future (like raspberrypi, by simply adding other parameters like `isarduino` in matrix and adapting the end)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,13 +6,14 @@ on:
   pull_request:
     branches: [ master ]
   workflow_dispatch:
-
-jobs:
-
-  build:
-    strategy:
-      matrix:
-        board:
+    inputs:
+      id:
+        description: The ID of the platform on which the build is run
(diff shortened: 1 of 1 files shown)


jgromes:
Thanks for this! So if I understand the changes corectly, all of the current functionality should be preserved, with the option to run CI of any platform by manually inputting the board id, right? If that is the case, is there a difference from running the CI by re-running the job from the web interface?

Mesteery:
Yes nothing nothing, I just used all the capacitites of `matrix` to make it more readable. And added a way to manually run a build for a specific board (or all of them), without commiting something with a specific message like before.

jgromes:
OK, thanks for the confirmation that nothing changed. I'm not that familiar with GitHub CI (as our company uses GitLab). Also it is possible input wasn't a thing when I first created the workflow.

PR #763, opened 2023-06-01
Bug in waitForMicroseconds prevTimingLen in V6.0.0

KevWal:
In Module::waitForMicroseconds if RADIOLIB_INTERRUPT_TIMING is defined the variable prevTimingLen has been incorrectly upated in v6.0.0 introducing a bug that stops compilation.

Fix: renamed _prevTimingLen to prevTimingLen to match other updates in 6.0.0


jgromes:
Merged, thanks - I should probably add an example for interrupt timing to catch these issues in the future.

PR #772, opened 2023-06-16
Pager: allow manual override of function bits

h3ndrik:
I wanted to send an activation code to a Skyper. This requires to set the function code to 0b10 and we can't manually set function codes.

This request introduces a new argument to the transmit function in order to overwrite the value of 'function'.

While at it, i restructured the bit shifting to be in line with the other code. (Do the bit shift when calculating the word inside the code, not inside a macro.)


jgromes:
Merged, thanks for the contribution - when originally writing the implementation, I only had this very helpful table from the specification, so it's probably a good idea to have the function bits accessible.

Screenshot_51

PR #779, opened 2023-06-27
modules: Don't read excess status data

alistair23:
The SX126x get status command returns a single status byte. The status byte is automatically read in the `Module::SPItransferStream()` function when we increment buffLen (see
https://github.com/jgromes/RadioLib/blob/master/src/Module.cpp#L259).

By setting numBytes we incorrectly end up reading 2 bytes. Instead set numBytes to zero so we only read the single status byte.


jgromes:
Well spotted, thanks! Setting the value to 0 actually makes sense, since the GetStatus command does not read any data after returning the status byte - unlike all other SX126x read commands.

PR #780, opened 2023-06-28
modules: SX126x: Remove SPIparseStatus error checking

alistair23:
The SX126x status is an unreliable way to detect errors. It is reported in forums online that it reports errors even when operations are succssful [1]. The Semtech reference implementation never checks the status value [2].

So let's remove some of the error checking on the status byte. In my testing it incorrectly reports errors. If we want to perform error checking we can use the GET_DEVICE_ERRORS command instead.

1: https://forum.lora-developers.semtech.com/t/sx1262-failure-to-execute-command/269/11
2: https://github.com/search?q=repo%3ALora-net%2FLoRaMac-node+SX126xGetStatus&type=code

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -2022,10 +2022,6 @@ int16_t SX126x::config(uint8_t modem) {
 int16_t SX126x::SPIparseStatus(uint8_t in) {
   if((in & 0b00001110) == RADIOLIB_SX126X_STATUS_CMD_TIMEOUT) {
     return(RADIOLIB_ERR_SPI_CMD_TIMEOUT);
-  } else if((in & 0b00001110) == RADIOLIB_SX126X_STATUS_CMD_INVALID) {
-    return(RADIOLIB_ERR_SPI_CMD_INVALID);
-  } else if((in & 0b00001110) == RADIOLIB_SX126X_STATUS_CMD_FAILED) {
-    return(RADIOLIB_ERR_SPI_CMD_FAILED);
   } else if((in == 0x00) || (in == 0xFF)) {
     return(RADIOLIB_ERR_CHIP_NOT_FOUND);
   }


jgromes:
Can you share a specific case where this is causing a problem? Unless there is a very good reason to not check some error, I would prefer to leave this in.

In the forum post you have linked, I have seen two well described cases of GetStatus returning not-OK result, one caused by low supply voltage and another one caused by sending out incorrect bytes when reading. Both of these are usage outside of specification, so I don't think this is a good proof of unreliability.

alistair23:
What I'm seeing is that whenever I run the `SetStandby` command (`0x80`) the next `GetStatus` command (`0xC0`) returns `0x2A`.

So setting the standby returns: `Failure to execute command` and `STBY_RC`. It seems to be in `STBY_RC` mode, yet returns an error. Not verifying the output of `standby()` also works, which is probably a better solution

jgromes:
Sorry, but I'm unable to replicate that behavior: in my testing, GetStatus after SetStandby behaves like this:

```
CMDW    80
SI      0
SO      A2
CMDR    C0
SI      0
SO      22
```

This works even after cold start (SetStandby is the first command called:

```
[SX1262] Initializing ... CMDW  80
SI      0
SO      A2
CMDR    C0
SI      0
SO      22
```

How do you know this error is a false positive? Don't get me wrong, the documentation for the error codes leaves a lot to be desired - I have no idea what does "Failure to execute command" mean according to Semtech. However, there could be some subtle reason,p erhaps hardware-related, as to why exactly this non-OK status is being returned.

Before declaring this feature faulty, I would need to have a reproducible case where it fails. If this cannot be provided, then I propose to provide a compile-time option to disable status checking, but have checking enabled by default.

For reference, the above test was done using Dorji DRF1262G without TCXO.

alistair23:
For reference I'm using one of these: https://www.waveshare.com/w/upload/c/c1/CoreSX1262_Sch.pdf

I don't **know** the error is a false positive. The main reason is that the official Semtech library never checks the status register, which makes me think that it's not a reliable source of information.

I'm not really sure that I can confirm there is no error, but communication does work.

This commit might be a better fix, as it leaves most of the other checks in place: https://github.com/alistair23/RadioLib/commit/3836e5b1496d160a41afb9caaa9df2f4a41077c8

jgromes:
IMO the fact that the official driver does not implement something is not a sufficient to consider that feature unreliable. Besides, there have been discrepancies between the official implementation and documentation in the past, see e.g. here: https://github.com/Lora-net/LoRaMac-node/issues/696

The commit you referenced is not a good solution, because it is only applied in one specific place. If the GetStatus command really is unreliable, then it can fail in any place the check is performed.

So until Semtech comes out and publishes an updated datasheet, errata or even a forum post that says the command should not be used, I don't see any reason to not use it. If you really want to disable these checks, there's already a mechanism that allows it, `RADIOLIB_SPI_PARANOID` macro which is enabled by default. If you disable it, none of these checks will be performed.

PR #781, opened 2023-06-28
Module: Get status from first byte after the command

alistair23:
Instead of getting the status from the very first byte in the input buffer, which will be what we read when we send the first byte. Let's instead get the status from the first byte after the command. This provides a more accurate status value.


jgromes:
Thanks, this is actually closer to what was implemented in the original transferStream (before buffered SPI).

PR #793, opened 2023-07-08
NonArduino: Tock: Fixup some bugs and expand the example

alistair23:


examples/NonArduino/Tock/libtockHal.h
```diff
@@ -190,6 +190,10 @@ class TockHal : public RadioLibHal {
     void spiEnd() {
     }
 
+    void yield() {
+      yield();
+    }
+
```
jgromes, line 196: Will this call a global yield or will it recursively call itself? If the intention is to call a global function, then I would suggest explicitly specifying it with `::yield();`.

examples/NonArduino/Tock/main.cpp
```diff
@@ -30,9 +30,19 @@
 // include the hardware abstraction layer
 #include "libtockHal.h"
 
+// Include some libtock-c helpers
+#include <libtock/temperature.h>
+#include <libtock/humidity.h>
+
+#define BUFFER_LEN 64
+
```
jgromes, line 38: I don't think it's a good idea to include specific sensors in the example. The example should be the necessary minimum. This pulls in more libraries and it also causes the functionality of the example to depend on the functionality of the sensors.

jgromes (commented):
I added a couple comments which I think should be addressed.

examples/NonArduino/Tock/main.cpp
alistair23: I was thinking that it makes the example more useful for users. It gives more of an idea how the library can be used.

If you really don't think it should be there I can remove it. Would a receive example (useing #ifdef) be alright to add as though?

alistair23:
Ok, I have dropped the sensor data

PR #801, opened 2023-07-21
Tock: A few more cleanups and fixes

alistair23:


jgromes:
Is there any specific reason why the settings passed to radio.begin() must be non-default? Usually I try to refrain from doing that, because if the API changes in the future, you have to keep in mind to update the examples too.

alistair23:
I'm not completely sure why, but these settings don't trigger any errors. So for the specific setup in the SoC they seem to be the ones to use

jgromes:
That's interesting, what sort of errors get triggered with different settings?

alistair23:
I think it reported error 0x707... I'd have to check again though

jgromes:
OK, thanks. Since all of the changes are Tock-specific, I will merge this as is. However, it would be better to report possible problems with certain settings as a normal issue, I don't think it's something that will only happen when using Tock. After all, you can use the expLoRaBLE board with Arduino.

alistair23:
Yeah, it shouldn't be Tock specific. To clarify I do see this error with the default `begin()`

```
failed, code -707
```

Which is:

```
/*!
  \brief SX126x failed to execute SPI command.
  Often this means that the module is trying to use TCXO while
  XTAL is connected (or vice versa). Make sure your crystal setup
  (e.g. TCXO reference voltage) matches your hardware by setting
  "tcxoVoltage" to 0 when using XTAL module, or to appropriate value
  when using TCXO module.
*/
#define RADIOLIB_ERR_SPI_CMD_FAILED                            (-707)
```

Which makes sense as `tcxoVoltage` need to be 0. The other values probably could be left as default, but currently this is what I have tested and works

jgromes:
Have you tried `radio.XTAL = true;` before calling begin? Might save you trouble with having to configure the other begin parameters.

alistair23:
So much better!

https://github.com/jgromes/RadioLib/pull/802

PR #807, opened 2023-08-05
Fixed wrong register definition 

### LINKED ISSUE #805: [RF69] (aka RFM69) truncates last Bit in TX process. Fix included.
I'm testing Adafruit Feather M0 boards which consists of Atmel SAMD21 microcontroller and HopeRF RFM69HCW radio. My goal is a communication between two such boards at a bitrate of 1200bps.

The issue is that not every packet gets through without corruption. I tried the same FSK settings of 1200bps with the RF69 library from https://github.com/lowpowerlab where the problem never happened. With the help of an SDR and a FSK demodulator the problem became obvious. During transmission the very last bit is truncated to about half of its original symbol length. Accordingly this last bit can no longer be sampled reliably by the receiver. The packet error rate was roughly 10% to 20%.

After many hours of investigation the fix for this bug seems quite clear. According to the datasheet from HopeRF this lines in RF69.h

```
#define RADIOLIB_RF69_SEQUENCER_OFF                             0b00000000
#define RADIOLIB_RF69_SEQUENCER_ON                              0b10000000
```
should be
```
#define RADIOLIB_RF69_SEQUENCER_OFF                             0b10000000
#define RADIOLIB_RF69_SEQUENCER_ON                              0b00000000
```

I'm wondering why nobody so far took notice of this severe issue. My explanation is that with a significantly slower MCU than this 48MHz Cortex M0 it might not happen. After all an artificial delay of a few hundred microseconds at the right place in the code is enough and the last bit is no longer truncated. With a higher bitrate than 1200bps it might also not happen because the symbol time is anyway shorter.

ramaza:
Defines for RADIOLIB_RF69_SEQUENCER_OFF and RADIOLIB_RF69_SEQUENCER_ON where mixed up. Fixes #805

### DIFF
--- a/src/modules/RF69/RF69.h
+++ b/src/modules/RF69/RF69.h
@@ -98,8 +98,8 @@
 
 // RF69 modem settings
 // RADIOLIB_RF69_REG_OP_MODE                                                  MSB   LSB   DESCRIPTION
-#define RADIOLIB_RF69_SEQUENCER_OFF                             0b00000000  //  7     7   disable automatic sequencer
-#define RADIOLIB_RF69_SEQUENCER_ON                              0b10000000  //  7     7   enable automatic sequencer
+#define RADIOLIB_RF69_SEQUENCER_OFF                             0b10000000  //  7     7   disable automatic sequencer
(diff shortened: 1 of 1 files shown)


jgromes:

PR #810, opened 2023-08-09
Remove unnecessary condition

nicklasb:
This will never be NULL and thus causes an error in ESP-IDF

```
managed_components/RadioLib/src/protocols/APRS/APRS.cpp: In member function 'int16_t APRSClient::begin(char, char*, uint8_t, bool)':

managed_components/RadioLib/src/protocols/APRS/APRS.cpp:27:8: error: the address of 'APRSClient::src' will never be NULL [-Werror=address]
   27 |   if((!src) && (this->phyLayer != nullptr)) {
```

### DIFF
--- a/src/protocols/APRS/APRS.cpp
+++ b/src/protocols/APRS/APRS.cpp
@@ -24,7 +24,7 @@ int16_t APRSClient::begin(char sym, char* callsign, uint8_t ssid, bool alt) {
     table = '/';
   }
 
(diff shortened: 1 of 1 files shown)


jgromes:
Actually, I think that entire condition should be removed - it doesn't make sense to return when `phyLayer` is set, in fact it will always be set for APRS over LoRa. I'm not sure why I added that check in the first place.

PR #834, opened 2023-09-20
disable syncword generation & detection, add method to set preamble polarity

baycom:
Ons patch allows the sync mechanism in the SX127X chips being disabled if setSyncWord is called with len=0 . Another patch adds a method for setting the preamble polarity. Both functions are being used in this project https://github.com/baycom/rps


src/modules/SX127x/SX127x.cpp
```diff
@@ -782,6 +782,24 @@ int16_t SX127x::setPreambleLength(size_t preambleLength) {
   return(RADIOLIB_ERR_UNKNOWN);
 }
 
+int16_t SX127x::setPreamblePolarity(int polarity) {
```
jgromes, line 785: Since there are only two possible values, I think it would be better to use a `bool` polarity, with `true` using the default and false for the inverted.

jgromes, line 785: Please add a call to `setPreamblePolarity` in `SX127x::beginFSK`, so that at reset, the value is always set to a known default.

src/modules/SX127x/SX127x.h
```diff
@@ -876,6 +876,13 @@ class SX127x: public PhysicalLayer {
     */
     int16_t setPreambleLength(size_t preambleLength) override;
 
+    /*!
+      \brief Sets FSK preamble polarity. Allowed values are RADIOLIB_SX127X_PREAMBLE_POLARITY_55 or RADIOLIB_SX127X_PREAMBLE_POLARITY_AA. Not allowed in LoRa mode.
+      \param polarity Preamble polarity bit in FSK mode).
+      \returns \ref status_codes
+    */
+    int16_t setPreamblePolarity(int polarity);
```
jgromes, line 884: Please add `setPreamblePolarity` to keywords.txt

jgromes (changes requested):
Thank you for the contribution! I left a few comments, though nothing major.

src/modules/SX127x/SX127x.cpp
baycom: If the value is just true or false how can you tell which one is 0x55 or 0xAA? There is no standard preamble in the FSK world.

baycom: There is already a direct register SPIsetRegValue in configFSK which does the same. Could remove it here and put the new method in beginFSK.

src/modules/SX127x/SX127x.h
baycom: done

src/modules/SX127x/SX127x.cpp
baycom: ok, done.

jgromes: It's true there is no standard in general - what I meant by default is the current polarity, which is fixed at 0x55 (sorry for not being clear).

jgromes: That's exactly what I meant, thanks.

baycom: Do you think it would be more convenient having a setPreamblePolarity55(true|false) method instead of the one above?

jgromes: I think it's fine with the current name. Maybe `invertPreamble(true|false);` could be more descriptive, analogous to `SX127x::invertIQ(true|false)`?

jgromes:
@baycom to not block this PR any longer, I did the changes myself.

Thank you so much for the contribution!

PR #835, opened 2023-09-22
[LoRaWAN] Implement full session persistence & more v1.1 specification

StevenCellist:
Hi Jan, great work on the LoRaWAN integration. As someone pointed out to me, the first 90% is working, but now we need to add the other 90%. Here's my contribution!   
I've got a list of items that still need to be done but which are more closely related to the actual hardware integration regarding frequency, datarate etc. Once we've gone through this PR, I'll open up some issues so we can tackle those as well.

Implemented the following items:
- check JoinNonce
- implement downlink frame counters (32-bit), checked as well
- implement sending multiple MAC commands
- save all relevant session parameters to persistent storage and load upon session restore
- only allow FPort = 0 when sending MAC-only payload
- implement LoRaWAN v1.1 errata (linked below the v1.1 specification download)
- added new function `deleteMacCommand` that properly deletes an enqueued MAC command
- renamed APB to ABP

Minor note about the change in LoRaWANMacCommand_t (`size_t` -> `uint8_t`: this saves precious NVM storage)


HeadBoffin:
Smoke test is fine - running it overnight to accumulate some uplinks & subsequent potential downlinks from TTN to process

HeadBoffin:
@StevenCellist, we really need to get ADR sorted for this not to break local networks - SF12 isn't good citizenship!

StevenCellist:
@jgromes what is your plan on implementing the PR? Is it under review (which I sure believe takes some time!) or are you waiting for further development?
I would love to add other things & continue the work, but the sooner essentials are integrated into RadioLib, the less people start working with RF- and LNS-unfriendly devices :)

jgromes:
@StevenCellist sorry for being a bit less responsive than usual - I will try to review the PR ASAP. However, there is one major point that I think can be addressed right away: not all platforms supported by RadioLib have persistent storage support. How will those be affected? Since this PR includes checking things such as the last used nonce, I would expect those to fail to join now.

Also, in general I think it would be better to have an option to bypass these checks. Even TTN has the option to allow duplicate nonces.

StevenCellist:
@jgromes no worries, take your time! But a simple heads up as to when you'll likely get around to it is welcome :)

Regarding the use of persistent storage, there are two options, of which I *strongly* prefer #1:
1) Restrict the use of LoRaWAN to devices with persistent storage; either internal or external. Adding a few (mega)bytes of PSRAM or whatever is simple and low-cost, and is less hacky then disabling all security checks on the network side - which exist for a reason too. I'm sure @HeadBoffin will like to chime in on this as well.
2) Add a whole range of #defines and un-readability, and comments to explain all caveats.

But yes, can add a LoRaWAN-type godmode that just disables the security checks if that's what you want for RadioLib. However I'd heavily discourage disabling the whole NVM.

BTW - even your code would fail at that point, as you retrieve the frame counter for uplinks from NVM ;)

HeadBoffin:
There is providing options to bypass checks and then there are options for bypassing checks.

GodMode that a developer has to turn on with some mild effort is one thing. Providing a magic switch on the front door is likely to end up with the less technically savvy shooting themselves in the foot trying to figure out what's going on and not likely to be well received for those helping them debug when they find out the LNS has been subverted and that the device is behaving more like ABP than OTAA. Or has managed to run totally amok and is now saturating the airwaves at SF12.

But that's more about @jgromes policy / view on the LW implementation. It would be nice to have a very compliant clean stack that can do Arduino or bare metal. And I hear on the grapevine that someone who can run the compliance tests is happy to help out. I doubt there are shed loads of cash awaiting for creating an almost totally certifiable stack, in the LoRaWAN Alliance sense, but there is definitely some kudos to be had.

That said, there are SAMD EEPROM libraries which are the main missing link and a mop up of any missing ARM boards can be done after. It was something I was looking at before the firehose of coding output from @StevenCellist hit my desk and I've not been able to keep up since.

Alternatively supporting v1.0.3 with its random DevNonce's could be reviewed. v1.1 is so much more a super set of the v1.0.x that this shouldn't be too onerous. Another potential is to provide detailed instructions on resetting the DevNonce - via the console or the CLI - so that a device can join and do its job but have to rejoin if power goes out. Mostly power doesn't go out, most of this activity is because ESP32 deep sleep & preserving state in LMIC is a constantly arising issue, but having to spot a device is trying to rejoin wouldn't be great. For production it's likely it will hit the next number in the sequence reasonably quickly anyway, depending on how TR007 is implemented. So at the very least, having a sleepy ESP32 stack available to the world is a good thing.

jgromes:
@StevenCellist regarding option 1, i.e. adding persistent storage - the major disadvantage is that this isn't portable. Until now, RadioLib has worked on any off-the-shelf development kit that has exposed SPI bus and a couple GPIO pins to connect to the radio.

> as you retrieve the frame counter for uplinks from NVM

Agree, but it would be able to join, "just" the frame counters would be always set to the same value; I wonder if the network servers actually check that ...

TL;DR I think the best course of action now to unblock this PR is to disable LoRaWAN on platforms that do not have non-volatile storage provided by their Arduino core.

@HeadBoffin 

> That said, there are SAMD EEPROM libraries which are the main missing link and a mop up of any missing ARM boards can be done after.

SAMD is the big one, but also the Arduino mbed boards (Nano 33, Portenta H7 and RP2040) don't support any sort of EEPROM emulation. I find that puzzling since ESP32 has had that for years, and for a few bytes like this I definitely do not want to have a filesystem. A simple EEPROM flash emulation with rudimentary wear leveling would be my go-to solution in a less generic environment. But overall I don't like the idea of either relying on external dependencies (that are not part of the Arduino core) or implementing our own mechanism.

On a side note: 

> It would be nice to have a very compliant clean stack that can do Arduino or bare metal

I don't think a full compliance is the primary goal here; the goal is to have a simple to use LoRaWAN implementation that can be deployed quickly on various hardware.

HeadBoffin:
> "just" the frame counters would be always set to the same value; I wonder if the network servers actually check that ...

Yup, they definitely do - the #1 gotcha for APB but not an issue for an OTAA join where they are reset.

> TL;DR I think the best course of action now to unblock this PR is to disable LoRaWAN on platforms that do not have non-volatile storage provided by their Arduino core.

It moves forward to let Steven do more MAC stuff whilst I look at SAM EEPROM.


> > That said, there are SAMD EEPROM libraries which are the main missing link and a mop up of any missing ARM boards can be done after.
> SAMD is the big one, but also the Arduino mbed boards (Nano 33, Portenta H7 and RP2040) don't support any sort of EEPROM emulation. 

It can be emulated with the libraries available. Nano 33 is SAMD and there are NVM implementations for RP2040. I can check in to STM32H747.

> I find that puzzling since ESP32 has had that for years, and for a few bytes like this I definitely do not want to have a filesystem. 

Various ARM implementations have had NVM for years but for what ever reason they've not surfaced in to the Arduino world. In LoRaWAN land, ESP32 has been a third class citizen until now - LMIC isn't easy to save state as mostly AVR & SAMD boards deep sleep without losing RAM so there's never been an incentive to get a good implementation in to the main library. It's all a matter of perspective. You & Steven have turned this around. No file systems need be harmed in the creation of NVM.

> A simple EEPROM flash emulation with rudimentary wear leveling would be my go-to solution in a less generic environment. But overall I don't like the idea of either relying on external dependencies (that are not part of the Arduino core) or implementing our own mechanism.

Theses are libraries that are in the Arduino eco system - they just don't work as an overlay to the original EEPROM library so not as simple as the ESP32 drop in replacement. An abstraction layer, which is what almost all of Arduino is anyway shouldn't be a stretch. The benefits in return are a whole extra range of boards to use.

But as I say, let's get started with the boards you are happy with and take it from there. Rome wasn't built in a day.

> > It would be nice to have a very compliant clean stack that can do Arduino or bare metal
> I don't think a full compliance is the primary goal here; the goal is to have a simple to use LoRaWAN implementation that can be deployed quickly on various hardware.

Very compliant ≠ full compliance. I know the spec. I know what compliance takes. I'm not that keen nor naieve to suggest working towards something where you've done the first 90%, Steven seems happy to work on the next 90% and then we all have to work on the last 90%. It would be a fools errand, some of the edge cases defy belief, but being a good TTN citizen, ie not shredding the local airwaves / network is a reasonable goal.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -55,12 +55,97 @@ int16_t LoRaWANNode::begin() {
     return(RADIOLIB_ERR_NETWORK_NOT_JOINED);
   }
 
+  uint16_t nvm_table_version = mod->hal->getPersistentParameter<uint32_t>(RADIOLIB_PERSISTENT_PARAM_LORAWAN_VERSION_ID);
```
jgromes, line 58: This seems unused in this PR, is the intention to use it later?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -681,23 +835,63 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len) {
...
+  uint32_t fcntDownPrev = 0;
+  if (isAppDownlink) {
+    fcntDownPrev = mod->hal->getPersistentParameter<uint32_t>(RADIOLIB_PERSISTENT_PARAM_LORAWAN_A_FCNT_DOWN_ID);
+  } else {
+    fcntDownPrev = mod->hal->getPersistentParameter<uint32_t>(RADIOLIB_PERSISTENT_PARAM_LORAWAN_N_FCNT_DOWN_ID);
+  }
+
+  // assume a 16-bit to 32-bit rollover when difference in LSB is smaller than MAX_FCNT_GAP
+  // if that isn't the case and the received fcnt is smaller or equal to the last heard fcnt, then error
+  if (fcnt16 <= fcntDownPrev && 0xFFFF - (uint16_t)fcntDownPrev + fcnt16 > RADIOLIB_LORAWAN_MAX_FCNT_GAP) {
```
jgromes, line 874: Please add parentheses to make the order of operations explicit (this can lead to warning or error for some compiler configurations).

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -749,23 +942,55 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len) {
...
+      RADIOLIB_DEBUG_HEXDUMP(foptsBuff, foptsBufSize);
+
+      isMACPayload = true;
+      this->uplink(foptsBuff, foptsBufSize, RADIOLIB_LORAWAN_FPORT_MAC_COMMAND);
+    }
+
+    // write the MAC command queue to nvm for next uplink
+    uint8_t queueBuff[sizeof(LoRaWANMacCommandQueue_t)];
+    memcpy(&queueBuff, &this->commandsUp, sizeof(LoRaWANMacCommandQueue_t));
+    mod->hal->writePersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_PERSISTENT_PARAM_LORAWAN_FOPTS_ID), queueBuff, sizeof(LoRaWANMacCommandQueue_t));
```
jgromes, line 974: Is it necessary to save this in NVM?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -749,23 +942,55 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len) {
       remLen -= processedLen;
       foptsPtr += processedLen;
     }
+
+    // if FOptsLen for the next uplink is larger than can be piggybacked onto an uplink, send separate uplink
+    if(this->commandsUp.len > 15) {
```
jgromes, line 947: I'm not sure it's a good idea to automatically send an uplink in this case - if we do, the device should then open the RX windows, right?

jgromes (commented):
Finally I had a bit of time to review this, I left a couple comments. Coming back to the platforms without NVM support, I realized that we can keep supporting them by forcing the to rejoin every time. The counters can be saved in volatile storage. So after this is merged, I will address that.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Yes, if future versions decide that the layout or size of some parts need to be shuffled, that parameter can be used to 'transition' from one version to another while keeping the session alive which in some cases could be crucial.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: If the device is going to sleep after a downlink, the MAC responses need to be stored for the next uplink, otherwise they would be lost. Especially on ESP32 boards where the RAM is discarded through deepsleep (and powercycle).

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: I agree that this may not be the best place to send the uplink - it's probably best to complete the downlink first. But I couldn't really come up with an elegant way to separate the uplink from the downlink. We would maybe need to 'signal' the user to trigger an uplink from the main code, but if that doesn't happen, the MAC commands will only pile up worse and worse.

src/protocols/LoRaWAN/LoRaWAN.cpp
HeadBoffin: In a later revision @StevenCellist may want to check best practise - sending an uplink to ack some MAC commands before sleeping if the sleep is going to be a while may be a good idea - particularly if there are more needed and/or the queue took priority over a user level downlink that may have some import.

TL;DR: Keep a short cycle going until all the messages have been processed.

src/protocols/LoRaWAN/LoRaWAN.cpp
HeadBoffin: I suspect that the listen for downlink is a MUST. As eluded to above, not returning to the user until commands are flushed out isn't unusual as functionality, even if rarely experienced. I think that's how LMIC does it.

HeadBoffin:
> Coming back to the platforms without NVM support, I realized that we can keep supporting them by forcing the to rejoin every time. 

With an incrementing DevNonce, that leads to a potential mess but with the other platforms their sleep current is very low & they save RAM so no need to rejoin.

> The counters can be saved in volatile storage

Works for OTAA, works for ABP if counters checks are turned off. I suspect many users deploying on Arduino environment will be OK / not care until it comes to bite them.

> . So after this is merged, I will address that.

I have SAMD NVM as work in progress - please let me know what the plan is so we don't duplicate effort.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: OK, in that case could you please add a comment stating this, as well as to explicitly cast the variable to void so that it is not throwing an unused variable warning?

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: The main thing that I'm worried about here is that we end up taking control from the user for a long time. I don't know how likely is the situation of user uplinks not being able to keep up with MAC commands.

The way I see it there are three options now:

1. Don't do the uplink automatically at all and let the user know that the queue is full. We could then have a method e.g. `flushMacCommands` that the user can call to do this MAC-only uplink/downlink.
2. Keep the uplink here and signal to user (e.g. by return code) that he needs to call downlink again. I don't like that too much since it seems to kind of break the normal flow from the users perspective - first they call uplink, then downlink and then they need to call downlink again for a reason that is out of their control.
3. Add the downlik here. This blocks for the longest amount of time, but in the current state it is likely the easiest and the cleanest solution.

Future solution would have to include interrupt-driven processing of the commands so that all of this really does happen in the background (as the standard calls for).

jgromes:
@HeadBoffin 

> I have SAMD NVM as work in progress - please let me know what the plan is so we don't duplicate effort.

That's appreciated, but I don't think it should be directly part of RadioLib. I think that should be work for the SAMD Arduino cores, as their "EEPROM library" emulation.

src/protocols/LoRaWAN/LoRaWAN.cpp
HeadBoffin: > I don't know how likely is the situation of user uplinks not being able to keep up with MAC commands.

Very unlikely - it's not a use case for LoRaWAN for people to try to keep up with the legal duty cycle and it's an extremely rare event that you end up with a pile of MAC commands moving up & down - you can trigger your own mess with a LinkCheckReq, a TimerReq, an ADR update, but a device getting a downlink after the initial foray at JoinAccept & ADR adjustments is rare - if everything is working and the RF is stable so no ADR changes, then a device can go months without a downlink.

> Future solution would have to include interrupt-driven processing of the commands so that all of this really does happen in the background 

But it will still block the user's use of the radio.

The processing time should be very short, it's giving the user back some cycles whilst waiting for Rx1 that people ask for but then never really make good use of. You can't run a real time system doing something really important if you need to hit the Rx1 & 2 windows, but again, with some evaluation, if the system is serviced before initiating committing to the radio, the system gets some time during the delay, if the Rx1 window gets trashed by an interrupt it's no worse than RF interference stopping reception. 

> (as the standard calls for).

Which standard?

MAC commands coming down can be processed as a low priority task, they just have to be enacted before the next uplink.

HeadBoffin:
> @HeadBoffin
> 
> > I have SAMD NVM as work in progress - please let me know what the plan is so we don't duplicate effort.
> 
> That's appreciated, but I don't think it should be directly part of RadioLib. I think that should be work for the SAMD Arduino cores, as their "EEPROM library" emulation.

I wasn't thinking of doing anything specific, just implementing the best library candidate - which will require some mods for things like:

```
#if !defined(RADIOLIB_EEPROM_UNSUPPORTED)
#include <EEPROM.h>
#endif
```

in ArduinoHal.cpp as at first glance not many of the existing libraries use an EEPROM.h header. But easily done by expanding the #if for board definitions.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: Thanks @HeadBoffin for the inputs - in this case I think going with option 3 from my previous post is the best option now. @StevenCellist this is the major thing blocking this PR.

Also, @HeadBoffin 

> Which standard?

I meant the LoRaWAN specification, which explicitly states the following - I understood that as meaning that the user/application should not be required to do anything (maybe apart from the initial join action) in order for MAC commands to work.

Screenshot_68

StevenCellist: Sorry, now I'm the one short on time :)
I'll get the PR fixed in the weekend, I expect! And after that work on the other stuff mentioned in the issue

StevenCellist:
@jgromes your comments should be resolved now

StevenCellist:
Ah... Strings. Copied that from the examples but that was not my brightest idea.

StevenCellist:
@jgromes I am not sure what is causing the checks to fail to install arduino-cli? I assume it is unrelated to my PR; is it something you can look into?

jgromes:
@StevenCellist I'll look into it, the important thing is that esp-idf and RPi builds succeed, so there is no issue in the PR.

PR #843, opened 2023-10-10
Incorrectly checking sx128x command status

chemary:
In `SX128x::SPIparseStatus` the bit mask to check the command status is incorrect, it's checking bits 3:1 when the documentation says that are 4:2
image

You can replace the constants and clearly see that they don't make sense
```c++
if((status & 0b00001110) == RADIOLIB_SX128X_STATUS_CMD_TIMEOUT)
if((status & 0b00001110) == RADIOLIB_SX128X_STATUS_CMD_ERROR)
if((status & 0b00001110) == RADIOLIB_SX128X_STATUS_CMD_FAILED)
// replacing the constant becomes
if((status & 0b00001110) == 0b00001100)
if((status & 0b00001110) == 0b00010000)
if((status & 0b00001110) == 0b00010100)
```

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -1499,11 +1499,11 @@ int16_t SX128x::config(uint8_t modem) {
 }
 
 int16_t SX128x::SPIparseStatus(uint8_t in) {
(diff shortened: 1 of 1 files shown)


jgromes:
Seems like a copy-paste error from the SX126x driver. Merged, thank you for the contribution!

PR #860, opened 2023-10-28
Update api adapt esp core 3.0.0-alpha2

lewisxhe:
esp32 core has been updated from IDF V4.x.x to IDF V5.x.x, and RadioLIb uses the updated API. The function has been tested normally in 3.0.0-alpha2,
Please see the specific changes [esp32-hal-ledc.h](https://github.com/espressif/arduino-esp32/blob/6d9ebea5008be540aaea5ad3cffc6ceb583692e3/cores/esp32/esp32-hal-ledc.h#L47C4-L47C4)

### DIFF
--- a/src/ArduinoHal.cpp
+++ b/src/ArduinoHal.cpp
@@ -145,10 +145,16 @@ void inline ArduinoHal::tone(uint32_t pin, unsigned int frequency, unsigned long
     // ESP32 tone() emulation
     (void)duration;
     if(prev == -1) {
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5,0,0)
       ledcAttachPin(pin, RADIOLIB_TONE_ESP32_CHANNEL);
+#endif
     }
     if(prev != frequency) {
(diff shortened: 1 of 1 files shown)


jgromes:
All looks good, merged - thank you very much for the contribution!

lyusupov:
Guys, please, be aware that ESP_IDF_VERSION macro may not had been defined in Arduino Core 1.0.x series  ( they use IDF < 4 )

jgromes:
@lyusupov thanks for pointing that out - in that case, I will add a check for `defined(ESP_IDF_VERSION)` to catch those version (although 2.0.0 was released more than two years ago at this point).

lyusupov:
> I will add a check for defined(ESP_IDF_VERSION) to catch those version

Great !

 > although 2.0.0 was released more than two years ago at this point

There was a breaking change between 1.x and 2.x that they had replaced LL hardware layer from (ESP32 only) custom design onto a API that supports other SoC (-S2,  -S3, -C3, ...)

My opinion that1.0.x is still the best (stable and performance wise) if an application needs no compatibility with more recent SoCs.

PR #866, opened 2023-11-03
LoRaWAN OTAA example fix

nmaas87:
node.begin(); does not exist anymore, in OTAA speech, this should be node.restoreOTAA();. However, I am unsure what should be the correct way to fix the ABP example.


nmaas87:
Already handled by #867

PR #868, opened 2023-11-04
Fixed Persistent Storage Issue for RP2040 with Arduino Pico Framework

nmaas87:
RP2040 does not have an EEPROM but always reserves the last 4K chunk of the flash for a software EEPROM - given the Arduino Pico framework is in use. It is exactly handled as ESP32 "SoftEEPROMs", meaning it does copy the "flashEEPROM" to memory on .begin(); and does need to .commit(); to write it back. We saw in the past that a node could successfully get an OTAA on a RP2040, but could never re-join - due to the missing commit and wrong init - which is with that fixed. As the "SoftEEPROM" is always written at the end of the flash, it also survives an Arduino Sketch reflash if not wiped afterwards by node.wipe(); More info and documentation here: https://arduino-pico.readthedocs.io/en/latest/eeprom.html

### DIFF
--- a/src/ArduinoHal.cpp
+++ b/src/ArduinoHal.cpp
@@ -104,7 +104,7 @@ void inline ArduinoHal::spiEnd() {
 
 void ArduinoHal::readPersistentStorage(uint32_t addr, uint8_t* buff, size_t len) {
   #if !defined(RADIOLIB_EEPROM_UNSUPPORTED)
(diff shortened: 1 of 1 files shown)


jgromes:
Good job on figuring it out, and thank you very much for this contribution! Merging before the big LoRaWAN PR because this is pretty self-contained.

PR #883, opened 2023-11-27
Check for RADIOLIB_SPI_PARANOID = 1 rather than just defined

### LINKED ISSUE #882: RADIOLIB_SPI_PARANOID cannot be easily turned off downstream
It would be lovely if BuildOpt.h would define RADIOLIB_SPI_PARANOID=1, and then this macro was checked with a simple "#if RADIOLIB_SPI_PARANOID", to allow a downstream project to disable it with -DRADIOLIB_SPI_PARANOID=0.

jp-bennett:
Closes #882

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -358,13 +358,13 @@
 #endif
 
 /*
- * Uncomment to enable "paranoid" SPI mode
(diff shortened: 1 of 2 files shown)


jgromes:
Thank you for pointing this out - you are right, this is indeed easier. In fact I will apply this also to all the other on/off BuildOpt macros as well.

PR #884, opened 2023-11-27
[LoRaWAN] Move TX power logic to function

StevenCellist:
Some applications might want to alter the Tx power, e.g. for testing LoRaWAN network coverage. Also, this moves the setOutputPower logic to one single function for better readability.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -698,6 +698,8 @@ int16_t LoRaWANNode::uplink(uint8_t* data, size_t len, uint8_t port, bool isConf
     this->isMACPayload = false;
   }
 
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks, this looks good. Just one point, would it be possible to change the method name to `setOutputPower`? That's used for the transceivers as well.

StevenCellist:
Understandable, but within the LoRaWAN scheme it's specified as Tx Power and would prevent some mistakes between function calls.
But completely your call, it's not much of a difference. Let me know :)

jgromes:
Hmm, that's actually a good point - event the transmit/receive methods are called uplink and downlink, so let's keep the current name. Though in that case, `setTxPower` is missing from keywords (keep in mind the files uses true tabs, not spaces).

StevenCellist:
Ah, that was the problem.. got it and done

PR #885, opened 2023-11-27
[LoRaWAN] Add datarate into event structure

StevenCellist:
As ADR can change the configured datarate, and for some bands the uplink datarate can be different to the downlink datarate, I think it is useful to be able to retrieve this using the event structure.

Yeah I'm not much of a Git expert so I messed up something with the commits while synchronizing the forks - it's now mislabelled previous commits as part of this PR but actually it's just one.


jgromes:
> it's now mislabelled previous commits as part of this PR but actually it's just one.

Doesn't matter, since by default I squash merged commits into a single one anyways ;)

Merged, thanks!

PR #893, opened 2023-11-30
Maintain consistency across SX126x and SX128x with macro rename

S5NC:
See https://github.com/jgromes/RadioLib/commit/82258105b798bc3e2a57a3de42d8a44e370f16a1#r133872034


jgromes:
Looks good, thanks for the fix!

PR #900, opened 2023-12-09
[LoRaWAN] Expose downlink framecounters

StevenCellist:
.. something with power-users wanting to know the number of downlinks after a deepsleep/reboot when there's no downlink event available (yet) :)

(yeah I still didn't fix the old commits, will try to fix that at once)


jgromes:
Merged, thanks for the contribution!

> (yeah I still didn't fix the old commits, will try to fix that at once)

Sometimes it's easiest to just clone the repository fresh.

PR #914, opened 2023-12-29
[LoRaWAN] Fix ABP initialization, support MAC in payload

### LINKED ISSUE #912: LoRaWAN mode ABP not working
LoRaWAN's ABP mode doesn't work currently. This is under investigation and I plan to fix this in early January.
The required changes also ask for some (minor) restructuring which I hope should make the whole implementation more manageable & readable.

StevenCellist:
* ABP initialization was not working, but is now fixed and tested working.
* When there are more than 15 bytes of MAC payload, this was not handled as MAC payload but as application payload. Now, this is recognized by looking at the FPort and handled accordingly. This fix is very relevant for ABP devices, which cause a very fat downlink (of 43 MAC bytes in my case).

This fat downlink of 9 MAC commands causes a response of 18 MAC bytes which is currently not processed correctly (it should trigger a MAC-only uplink at FPort 0); currently, it just fills the available 15 bytes of FOpts, which causes a second downlink that will request the remaining MAC payload. I will work on a fix, but this is quite workable for now. 

**_Please make sure that you are in good reach of a gateway when you initialize an ABP device to prevent endless downlinks_**.

Fixes #912 and by extent #904

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -1194,7 +1194,7 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len, LoRaWANEvent_t* event)
   // process FOpts (if there are any)
   if(foptsLen > 0) {
     // there are some Fopts, decrypt them
-    uint8_t fopts[max(RADIOLIB_LORAWAN_FHDR_FOPTS_LEN_MASK, (int)foptsLen)];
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you for the fix! I think everything is OK, just one thing: you are using `min()` and `max()` which may or may not exist (see https://github.com/jgromes/RadioLib/actions/runs/7357961464/job/20030435017?pr=914#step:6:1262). Instead use `RADIOLIB_MIN` and `RADIOLIB_MAX` macros which are platform-agnostic and don't depend on math.h

StevenCellist:
Ah, right, I forgot about those..

jgromes:
All good now and merged, thanks for the contribution!

PR #941, opened 2024-01-21
examples: NonArduino: Raspberry Pi Pico

cameron-goddard:
Add support for the Raspberry Pi Pico with the native [Pico SDK](https://github.com/raspberrypi/pico-sdk) instead of through the Arduino core.

This has been tested with a Pico and an SX1276 radio module.


examples/NonArduino/Pico/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+// now we can create the radio module
+// NSS pin:  26
+// DIO0 pin:  14
+// RESET pin:  22
+// DIO1 pin:  15
+SX1276 radio = new Module(hal, RFM_NSS, RFM_DIO0, RFM_RST, RFM_DIO1);
+
+
+int main() {
+  stdio_init_all();
```
jgromes, line 58: It's probably not needed to call this here, as `stdio_init_all()` is called from `PicoHal::init()`, which is in turn called in `Module::init()` as the first step in `radio.begin()`

examples/NonArduino/Pico/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  gpio_init(RFM_NSS);
+  gpio_init(RFM_RST);
+
+  gpio_set_dir(RFM_NSS, GPIO_OUT);
+  gpio_set_dir(RFM_RST, GPIO_OUT);
```
jgromes, line 64: Pin direction configuration is handled by the library. However, if it requires a call to `gpio_init`, I would propose to move this call to `PicoHal::pinMode`.

examples/NonArduino/Pico/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  // reset the RF module
+
+  sleep_ms(10);
+  gpio_put(RFM_RST, 0);
+  sleep_ms(10);
+  gpio_put(RFM_RST, 1);
```
jgromes, line 71: Again, this is not needed, the module will be reset during `radio.begin()`.

Just trying to remove the redundant portions of code ;)

examples/NonArduino/Pico/PicoHal.h
```diff
@@ -0,0 +1,142 @@
...
+#include "hardware/spi.h"
+#include "hardware/timer.h"
+
+// create a new Raspberry Pi Pico hardware abstraction 
+// layer using the Pico SDK
+// the HAL must inherit from the base RadioLibHal class
+// and implement all of its virtual methods
+class PicoHal : public RadioLibHal {
+public:
+  PicoHal(spi_inst_t *spiChannel, uint32_t misoPin, uint32_t mosiPin, uint32_t sckPin, uint32_t spiSpeed = 500 * 1000)
```
jgromes, line 18: The default SPI speed is quite low (500 kHz vs 2 MHz in Arduino). Is there a specific reason for that? Asking because of #938.

jgromes (commented):
Thank you for the contribution! I left some comments (nothing major), though there is one point that should be addressed - CI. I can add it, but I need to know how to install the SDK. Do you have some pointers how to do that?

jgromes:
One more thing, it would be nice to have at least build.sh and clean.sh scripts.

PR #952, opened 2024-02-01
include esp_attr.h where IRAM_ATTR is used

remenyo:
These two commits contain a small fix in two files that allowed my PlatformIO project to build.

They reference `IRAM_ATTR` attribute, but do not not include esp_attr.h; now they do, with the same ifdef guard as the attributes themselves.

If this issue is not apparent for other people and only my system is wrong, feel free to close this request.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -1,5 +1,8 @@
 #include "LoRaWAN.h"
 #include <string.h>
+#if defined(ESP8266) || defined(ESP32)
+#include "esp_attr.h"
+#endif
 
 #if !RADIOLIB_EXCLUDE_LORAWAN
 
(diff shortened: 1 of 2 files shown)


jgromes:
Thank you for the contribution - it's interesting this did not come up before, especially in ESP-IDF builds. I guess some headers are added automatically there, but I don't see any reason to not have this explicitly. Merged, thanks!

arendst:
There is no ``esp_attr.h`` on ESP8266 so compilation fails.

I see no issue on ESP32 removing this PR either...

jgromes:
@arendst that is interesting - I tried building this for ESP8266 using Arduino core, and you are right, it's broken (https://github.com/jgromes/RadioLib/actions/runs/8039040028/job/21955616874)

However, it seems that the `esp_attr.h` file does exist in ESP8266 SDK (https://github.com/espressif/ESP8266_RTOS_SDK/blob/master/components/esp8266/include/esp_attr.h), just not in its Arduino core, which is very strange, especially since it does exist in Arduino core for ESP32.

I changed the macro to check for `defined(ESP_PLATFORM)`, which is set for ESP8266 SDK and ESP-IDF, but not set in Arduino ESP8266 core. It now compiles for ESP8266 and ESP32 Arduino cores as well as for ESP-IDF and for ESP32 in Platformio.

PR #958, opened 2024-02-08
Fix stringop-truncation warning in AX25 callsign getter

remenyo:
stringop-truncation warning was caught by GCC12 in the callsign getter, as it copies 6 characters instead of 7.

I see in in the header file : [RADIOLIB_AX25_MAX_CALLSIGN_LEN + 1](https://github.com/jgromes/RadioLib/blob/7945ffb996106329472cd828c98ee86af3d4cbc7/src/protocols/AX25/AX25.h#L316) I guess +1 is for the null termination.

So I added +1 to the copy in the getter as well, as i guess it was the intent to terminate the resulting char array as well.

---

Note that I don't use this feature of the library, (-> I havent tested it) nor I know if this is the correct way to fix.

### DIFF
--- a/src/protocols/AX25/AX25.cpp
+++ b/src/protocols/AX25/AX25.cpp
@@ -440,7 +440,7 @@ int16_t AX25Client::sendFrame(AX25Frame* frame) {
 }
 
 void AX25Client::getCallsign(char* buff) {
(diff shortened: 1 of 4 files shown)


jgromes:
Thanks for pointing this out - I think I did see this warning at some point in CI runs but did not see it as a high priority. Merged, thank you for the contribution!

Issue #950, opened 2024-01-31
LoRaWAN: fcntUp wraps after 50 transmissions


barnslig:
### Bug description

The `fcntUp` wraps around after 50 transmissions, resulting in no more messages being received by the network server.

### Reproduction

The bug can be reproduced as follows in god mode without sending 50 uplinks:

```cpp
#include <Arduino.h>
#include <RadioLib.h>

SX1262 radio = new Module(8, 14, 12, 13);
LoRaWANNode node(&radio, &EU868);

void setup()
{
  Serial.begin(115200);

  delay(2500);

  node.wipe();

  for (unsigned int i = 0; i < 100; i += 1)
  {
    node.restoreFcntUp();

    node.fcntUp += 1;

    Serial.printf("fcntUp: %u | iteration: %u\n", node.fcntUp, i);

    node.saveFcntUp();
  }
}

void loop()
{
}

```

The output is as follows and shows the wrap around after 50 iterations:

<details>
<summary>Log output</summary>

```
fcntUp: 1 | iteration: 0
fcntUp: 2 | iteration: 1
fcntUp: 3 | iteration: 2
fcntUp: 4 | iteration: 3
fcntUp: 5 | iteration: 4
fcntUp: 6 | iteration: 5
fcntUp: 7 | iteration: 6
fcntUp: 8 | iteration: 7
fcntUp: 9 | iteration: 8
fcntUp: 10 | iteration: 9
fcntUp: 11 | iteration: 10
fcntUp: 12 | iteration: 11
fcntUp: 13 | iteration: 12
fcntUp: 14 | iteration: 13
fcntUp: 15 | iteration: 14
fcntUp: 16 | iteration: 15
fcntUp: 17 | iteration: 16
fcntUp: 18 | iteration: 17
fcntUp: 19 | iteration: 18
fcntUp: 20 | iteration: 19
fcntUp: 21 | iteration: 20
fcntUp: 22 | iteration: 21
fcntUp: 23 | iteration: 22
fcntUp: 24 | iteration: 23
fcntUp: 25 | iteration: 24
fcntUp: 26 | iteration: 25
fcntUp: 27 | iteration: 26
fcntUp: 28 | iteration: 27
fcntUp: 29 | iteration: 28
fcntUp: 30 | iteration: 29
fcntUp: 31 | iteration: 30
fcntUp: 32 | iteration: 31
fcntUp: 33 | iteration: 32
fcntUp: 34 | iteration: 33
[...]

PR #970, opened 2024-02-14
[LoRaWAN] Fix fcntUp wrap-around after 50 transmissions

barnslig:
This PR contains the bug fix proposed by @StevenCellist in #950.

I tested the bug fix in a live environment and was able to send more than 50 messages.


jgromes:
@barnslig thank you for the contribution!

@StevenCellist can this be merged? I think you posted you were working on this at some point.

StevenCellist:
Looks exactly like the solution I posted in #950 which has been confirmed to solve the issue, so I'd say: sure. Let's try out my new powers ;)

Issue #965, opened 2024-02-12
[LORAWAN] sending uplink packet… failed code  -4


altinem:
Hello,
First of all, I would like to apologize for opening a closed issue yesterday.  I am new to Lora and Lorawan environments.  I have a Heltec  Wifi/Lora v3 with SX1262 lora chip and using Visual Code with Platformio to flash it. I also have  a SendeCap M2 gateway connected to the TTN. I have flashed the board with Lorawan_end_device with OTAA in the example folder and  The OTAA was successful. However, i am getting  “sending uplink packet…failed code. -4”  error code. I received an email back today from the author of this project saying that my message is too long to send message. It cannot  be because I am using the sketch from example folder it only sends “Hello World” .  Could you please help me with this issue? Again, I did not add any code to the sketch. I only change the TTN keys and the region . Thank you in advance.

StevenCellist:
No matter how few changes you made, please share your sketch and some of the serial output. Very very much preferred inside code blocks using the `<>` tool.

altinem:
`/*
  RadioLib LoRaWAN End Device Example

  This example joins a LoRaWAN network and will send
  uplink packets. Before you start, you will have to
  register your device at https://www.thethingsnetwork.org/
  After your device is registered, you can run this example.
  The device will join the network and start uploading data.

  LoRaWAN v1.1 requires the use of EEPROM (persistent storage).
  Please refer to the 'persistent' example once you are familiar
  with LoRaWAN.
  Running this examples REQUIRES you to check "Resets DevNonces"
  on your LoRaWAN dashboard. Refer to the network's 
  documentation on how to do this.

  For default module settings, see the wiki page
  https://github.com/jgromes/RadioLib/wiki/Default-configuration

  For full API reference, see the GitHub Pages
  https://jgromes.github.io/RadioLib/
*/

// include the library
#include <RadioLib.h>

// SX1262 has the following pin order:
// Module(NSS/CS, DIO1, RESET, BUSY)
SX1262 radio = new Module(8, 14, 12, 13);

// SX1278 has the following pin order:
// Module(NSS/CS, DIO0, RESET, DIO1)
//SX1278 radio = new Module(10, 2, 9, 3);

// create the node instance on the EU-868 band
// using the radio module and the encryption key
// make sure you are using the correct band
// based on your geographical location!
//LoRaWANNode node(&radio, &EU868);

// for fixed bands with subband selection
// such as US915 and AU915, you must specify
[...]

PR #974, opened 2024-02-15
[LoRaWAN] Improve examples, add getter for DevAddr

StevenCellist:
- Decrease default payload length (#965)
- Remove specific 'persistent' example and make the other examples persistent by default
- Expose the DevAddr through `getDevAddr()`
- Minor fixes to comments
- Minor fixes to Serial output

Note that the MAC response for DeviceTime and LinkCheck are not solved with this PR, as those will be for @HeadBoffin.


jgromes:
All looking good, merged. Thanks!

Issue #968, opened 2024-02-13
Test LoRaWAN LinkCheck & DeviceTime


HeadBoffin:
### Discussed in https://github.com/jgromes/RadioLib/discussions/967

<div type='discussions-op-text'>

<sup>Originally posted by **michapr** February 13, 2024</sup>
Hi,

using the LoRaWAN_End_Device_Reference.ino I cannot get valid results for the LinkCheck and DeviceTime values.
Connection itself is ok, I can see the transferred data in TTN console.

code (from examples file):
```
    uint8_t margin = 0;
    uint8_t gwCnt = 0;
    if(node.getMacLinkCheckAns(&margin, &gwCnt)) {
      Serial.print(F("[LoRaWAN] LinkCheck margin:\t"));
      Serial.println(margin);
      Serial.print(F("[LoRaWAN] LinkCheck count:\t"));
      Serial.println(gwCnt);
    }

    uint32_t networkTime = 0;
    uint8_t fracSecond = 0;
    if(node.getMacDeviceTimeAns(&networkTime, &fracSecond, true)) {
      Serial.print(F("[LoRaWAN] DeviceTime Unix:\t"));
      Serial.println(networkTime);
      Serial.print(F("[LoRaWAN] LinkCheck second:\t1/"));
      Serial.println(fracSecond);
    }
```
Result:
```
LoRaWAN] LinkCheck margin:	0
[LoRaWAN] LinkCheck count:	0
[LoRaWAN] DeviceTime Unix:	0
[LoRaWAN] LinkCheck second:	1/0
```

What maybe wrong here?

Thanks!

**EDIT:**
another interesting detail:
If add the request
`node.sendMacCommandReq(RADIOLIB_LORAWAN_MAC_LINK_CHECK);`
I will get an answer for
`node.getMacDeviceTimeAns(&networkTime, &fracSecond, true)`
(with same results as above...)
</div>

HeadBoffin:
Will test and respond by Thursday pm GMT latest.

HeadBoffin:
It isn't working as expected, code has been identified, unexpected meeting has been scheduled so this is likely to be fixed early next week.

PR #980, opened 2024-02-19
[LoRaWAN] Improve persistence behaviour, add dwell time error up debug output

### LINKED ISSUE #976: LoRaWAN OTAA saves join session data in background but not f_cnt
If you do a join on a device that can persist the session, it is saved & recalled as expected. But if you don't save session after each uplink, the f_cnt isn't updated so subsequent reboots of the device restore the session but then has to send uplinks until they exceed the last highest count.

**To Reproduce**
Join on OTAA, don't do node.saveSession(); after an uplink. Leave to send a few uplinks. Reboot it.

**Expected behaviour**
If there is a saved session, restore that, if not don't persist join info in node.beginOTAA until the first call of saveSession then save the whole lot and subsequently any changes & f_cnt.

If the developer doesn't call saveSession, then on reboot it will re-join with a reset f_cnt per LW spec. If they do call it after each uplink then everything will be A-OK. 

Side note: Perhaps hash the EUI's + Keys and compare with saved session, wipe if changes are detected - otherwise we can change creds on a device but restore a session for another device, possibly?

**Additional info (please complete):**
 - MCU: ESP32 but anything that can persist
 - Wireless module type LoRaWAN
 - Arduino IDE version  [e.g. 1.8.5]
 - Library version 6.4.2

StevenCellist:
Changelog is pretty much in the commit titles (except that I didn't re-fork so there are two old commits as well).
Fixes #976, now also tested through PIO Remote at @HeadBoffin's setup that is starting to get running. Before this makes it into a new version, we can hopefully give it a good shot through a certain 'test-suite', we're getting there!

Most important change is that join-accept information is now only saved through `node.saveSession()`; `node.beginOTAA()` now only saves the Nonces to adhere to LW spec. The resulting behaviour should also be less confusing for new users.
This change is achieved by keeping the keys in EEPROM all zero until `saveSession()` is called. As soon as `beginOTAA()` is called, it checks if the keys are all zero (in which case a new join must be issued with the existing, increased Nonces); if the keys are present, `saveSession()` was apparently called at some point and we can safely assume that restoring is fine.
ABP mode does about the same, except that it can't do a new join, so if its keys are all zeroes it will restart from 0-frame counters.


StevenCellist:
WIP.
@jgromes maybe an idea to have a label for that? Just a random thought.

jgromes:
@StevenCellist we can convert a PR back to a draft if you need more time for working on this.

StevenCellist:
Just more improvements in progress with the help of my friend. Keeping it a PR will maybe help people finding it in case they want to use it. But draft also fine by me.. mostly: don't merge yet :)

StevenCellist:
Now also fixes #968, expecting more WIP

StevenCellist:
PR changelog:
* [LoRaWAN] Fix DeviceTime & LinkCheck (#968)
* [LoRaWAN] Make dwell time error explicit
* [LoRaWAN] Fix downlink FPort reporting incorrect value for network downlinks
* [LoRaWAN] Make first uplink use FCnt 0 as per spec, instead of starting from 1
* [LoRaWAN] Move session persistence fully to `saveSession()`, such that `beginOTAA()` and `beginABP()` logic now only follows pure specification
* [LoRaWAN] Update all examples to reflect changes
* [ArduinoHAL] Only write new values by reading before writing

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -277,7 +280,41 @@ int16_t LoRaWANNode::restoreChannels() {
...
+  mod->hal->writePersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_EEPROM_LORAWAN_NWK_SENC_KEY_ID), zeroes, RADIOLIB_AES128_BLOCK_SIZE);
+}
+
+bool LoRaWANNode::isActiveSession() {
+  Module* mod = this->phyLayer->getMod();
+  this->devAddr = mod->hal->getPersistentParameter<uint32_t>(RADIOLIB_EEPROM_LORAWAN_DEV_ADDR_ID);
+  mod->hal->readPersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_EEPROM_LORAWAN_APP_S_KEY_ID), this->appSKey, RADIOLIB_AES128_BLOCK_SIZE);
+  mod->hal->readPersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_EEPROM_LORAWAN_FNWK_SINT_KEY_ID), this->fNwkSIntKey, RADIOLIB_AES128_BLOCK_SIZE);
+  mod->hal->readPersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_EEPROM_LORAWAN_SNWK_SINT_KEY_ID), this->sNwkSIntKey, RADIOLIB_AES128_BLOCK_SIZE);
+  mod->hal->readPersistentStorage(mod->hal->getPersistentAddr(RADIOLIB_EEPROM_LORAWAN_NWK_SENC_KEY_ID), this->nwkSEncKey, RADIOLIB_AES128_BLOCK_SIZE);
```
jgromes, line 299: I don't think we should be overwriting the current keys just to test if a saved session exists. I'm not sure it actually breaks something, but it seems like a major side effect. If this method is really ment to just test the session exists, then it should save the keys into a local buffer.

Also, does this still work if the persistent storage is not initialized to 0x00s, but to 0xFFs?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -697,17 +722,18 @@ int16_t LoRaWANNode::beginABP(uint32_t addr, uint8_t* nwkSKey, uint8_t* appSKey,
   if(fNwkSIntKey) { checkSum ^= LoRaWANNode::checkSum16(fNwkSIntKey, 16); }
   if(sNwkSIntKey) { checkSum ^= LoRaWANNode::checkSum16(sNwkSIntKey, 16); }
 
-  bool validCheckSum = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_CHECKSUM_ID) == checkSum;
-  bool validMode = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_MODE_ID) == RADIOLIB_LORAWAN_MODE_ABP;
+  bool isValidCheckSum = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_CHECKSUM_ID) == checkSum;
+  bool isValidMode = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_MODE_ID) == RADIOLIB_LORAWAN_MODE_OTAA;
```
jgromes, line 726: Seeing as this is in beginABP, should this check really be for OTAA mode?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -277,7 +280,41 @@ int16_t LoRaWANNode::restoreChannels() {
   }
   return(RADIOLIB_ERR_NONE);
 }
-#endif
+
+void LoRaWANNode::clearSession() {
```
jgromes, line 284: Shouldn't clearSession also zero devAddr, seeing as `isActiveSesion` checks it?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -697,17 +722,18 @@ int16_t LoRaWANNode::beginABP(uint32_t addr, uint8_t* nwkSKey, uint8_t* appSKey,
...
-  bool validCheckSum = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_CHECKSUM_ID) == checkSum;
-  bool validMode = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_MODE_ID) == RADIOLIB_LORAWAN_MODE_ABP;
+  bool isValidCheckSum = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_CHECKSUM_ID) == checkSum;
+  bool isValidMode = mod->hal->getPersistentParameter<uint16_t>(RADIOLIB_EEPROM_LORAWAN_MODE_ID) == RADIOLIB_LORAWAN_MODE_OTAA;
 
-  if(!force && validCheckSum && validMode) {
-    // the device has joined already, we can just pull the data from persistent storage
-    RADIOLIB_DEBUG_PRINTLN("Found existing session; restoring...");
-    
+  if(isValidCheckSum && isValidMode && !force && this->isActiveSession()) {
```
jgromes, line 728: I think this logic can be simplified (I had a bit of a rough time trying to follow the current one). Since `isValidCheckSum` and `isValidMode` are not evaluated independently, it could be the following:

```c++
if(isValidCheckSum & isValidMode) {
  if(!force && this->isActiveSession) {
    return(this->restore());
  }
  // either forced or no active session
  this->clearSession();
} else {
  // either invalid checksum or mode, do the debug and wipe either way
  this->wipe();
}
```

jgromes: A similar point also applies to a similar section in beginOTAA

jgromes (changes requested):
@StevenCellist very good job as always! I left I couple comments/clarifications.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: > I don't think we should be overwriting the current keys just to test if a saved session exists. I'm not sure it actually breaks something, but it seems like a major side effect. If this method is really ment to just test the session exists, then it should save the keys into a local buffer.

In itself a fair shout - I cannot immediately come up with a possible breaking situation, but safer with separate buffers. Will fix.

> Also, does this still work if the persistent storage is not initialized to 0x00s, but to 0xFFs?

This function would not be called unless the mode and checksum are also verified, which would both quickly fail if all is set to 0xFF. So I am not worried about that. If we still want it to be safe, I can put some extra thought into it..

jgromes: > This function would not be called unless the mode and checksum are also verified, which would both quickly fail if all is set to 0xFF. So I am not worried about that. If we still want it to be safe, I can put some extra thought into it..

I suppose it is private, so we have full control over when it gets called. No need to dig into the unitialized case any further, thanks for the clarification.

jgromes:
@StevenCellist FYI I pushed a minor update to LoRaWAN in 9f0bdffc50f71c04123741b47099fd3f5a16e570

StevenCellist:
Thanks for the heads-up. I will fix the mentioned points tomorrow!

StevenCellist:
In addition to earlier changelog (missed + new item):
* [LoRaWAN] Improve debug output readability
* [LoRaWAN] Ignore rejoin when a session is active and join is not forced

PR is tested on the following configurations:
- US915 OTAA
- EU868 OTAA
- EU868 ABP

Actually now also confirmed that forced MAC uplinks (when too large to piggyback) also work as intended!

@HeadBoffin is willing to give it a quick go tomorrow; once he's given his blessing, it's ready for merge!
@jgromes can we maybe do a (small) version bump to get the FCnt bug fix out, as well as all the other QoL changes? Work is underway for the wiki, and having a solid release to work with would help :)

HeadBoffin:
Smoke test of non-persistant & persistent looks good to go!

jgromes:
Amazing guys, thanks for this!

@StevenCellist I see no reason to not bump to 6.4.3, I will do that ASAP.

jgromes:
@StevenCellist sure, we can wait a bit. I also wanted to merge #987 before this next update, which has been done, so update of examples is probably the last thing left.

PR #990, opened 2024-02-27
IDF Component Registry manifest

nicklasb:
Ok then. 

I added a idf_components.yml and did a --dry-run using compote, it didn't complain, at least. 

So what is done is:
1. Sign in at [components.espressif.com](https://components.espressif.com/). I use Github.
2. In the profile menu, click tokens and create a token. I gave it all the rights as I way looking around the CLI.
3. Add a `~.espressif/idf_component_manager.yml` or equivalent on your platform. 
4. Make it look like:
```
profiles:
  default:
    api_token: "ADD THE FULL TOKEN HERE"
```
5. Now you are logged in on the command line when you run IDF stuff. 
6. In the repository folder, run: `compote component upload --name Radiolib`

If you haven't gotten any errors, you should be able to search for find Radiolib at https://components.espressif.com/ .
I hope I haven't forgotten any steps, is really tired today, but seeing that you seem to work with getting stuff up in space, I am sure you can get stuff up on Espressif as well if I have made some mistake. :-)

Good luck!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ extras/SX126x_Spectrum_Scan/out/*
 
 # cmake
 build/
+
+# Compote build output
+dist
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -1,0 +1,11 @@
+version: "6.4.2"
+description: "Universal wireless communication library. User-friendly library for sub-GHz radio modules (SX1278, RF69, CC1101, SX1268, and many others), as well as ham radio digital modes (RTTY, SSTV, AX.25 etc.) and other protocols (Pagers, LoRaWAN)."
+tags: "radio, communication, morse, cc1101, aprs, sx1276, sx1278, sx1272, rtty, ax25, afsk, nrf24, rfm96, sx1231, rfm96, rfm98, sstv, sx1278, sx1272, sx1276, sx1280, sx1281, sx1282, sx1261, sx1262, sx1268, si4432, rfm22, llcc68, pager, pocsag, lorawan"
+url: "https://github.com/jgromes/RadioLib"
+repository: "https://github.com/jgromes/RadioLib.git"""
+license: "MIT"
+dependencies:
+  # Required IDF version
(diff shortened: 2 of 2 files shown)


jgromes:
Thanks! I'll read up on some docs before merging this, to make sure I understand the workflow. But it seems to be similar to how platformio works.

Come to think of it I should probably automate publishing with some CI, it's getting quite tedious to be doing it manually, which is:

1. Change versions in the appropriate files
2. Push the updated files
3. Wait for CI to build every example for every platform
4. Create a tag (this gets picked up by Arduino library manager)
5. Checkout the tag and publish to platformio
6. Publish to espressif

nicklasb:
Yeah, that is probably a good move. Could be good idea to try once manually first though to get any kinks out.
Promise it is just not because I need it to be able to get my library up. :-)

jgromes:
@nicklasb all looks good, the package is up: https://components.espressif.com/components/jgromes/radiolib. Expect an update to 6.4.3 soon ;)

For future reference when putting the release CI together, the command to upload the package must include the namespace: `compote component upload --name RadioLib --namespace jgromes`, otherwise a rather unhelpful `ERROR: You are not allowed to perform this action.` is shown.

nicklasb:
Ok, way cool. it worked! (https://components.espressif.com/components/robusto/robusto/versions/0.1.5-alpha)

For me it worked without the namespace initially, oddly. I asked for the "robusto" namespace though, so I have two namespaces, might be something like that. I didn't mention the namespace aspect as you have jgromes/ everywhere, and I assumed that you do not want to diverge from that.

There is one minor annoyance, though, and that is that the ESP-IDF CMakelist.txt reference to RadioLib had to be changed to lowercase, which might be of note to those who use RadioLib in ESP-IDF as a git-dependency following the case of the repo, or use the exact source of the nonArduino/ESP-IDF-example. Robusto is lowercase everywhere so it won't get stuck on that.

PR #998, opened 2024-03-02
Receive messages for multiple POCSAG RICs

jasiek:


src/protocols/Pager/Pager.cpp
```diff
@@ -500,6 +525,15 @@ void PagerClient::write(uint32_t codeWord) {
   }
 }
 
+bool PagerClient::addressMatched(uint32_t addr) {
+  for (size_t i = 0; i < filterNumAddresses; i++) {
```
jgromes, line 529: This should probably check `filterAddresses != NULL` .

src/protocols/Pager/Pager.cpp
```diff
@@ -28,6 +28,14 @@ PagerClient::PagerClient(PhysicalLayer* phy) {
   #if !RADIOLIB_EXCLUDE_DIRECT_RECEIVE
   readBitInstance = phyLayer;
   #endif
+
+  filterMask = 0xFFFF;
```
jgromes, line 32: If the intention is to initialize the mask to all ones, then it should be `0xFFFFF`, to be 20 bits.

jgromes (changes requested):
Thank you for the contribution! I left some comments, I think that this can be merged.

src/protocols/Pager/Pager.cpp
```diff
@@ -243,9 +251,29 @@ int16_t PagerClient::transmit(uint8_t* data, size_t len, uint32_t addr, uint8_t
...
+  return startReceiveInternal();
+}
+
+int16_t PagerClient::startReceive(uint32_t pin, uint32_t *addr, uint32_t mask, size_t numAddress) {
+  // save the variables
+  readBitPin = pin;
+
+  delete[] filterAddresses;
+  filterAddresses = new uint32_t[numAddress];
+  memcpy(filterAddresses, addr, numAddress);
```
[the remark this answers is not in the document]
jasiek: Yeah, I was thinking about that too. So do you mean to keep a variable for when there's a single address to service the original version of the method, and then keep a pointer for a list of addresses?

What's the use case for masks anyway? I'm new to POCSAG, so I had assumed it would always be sufficient to pass in a list of addresses to cover all use cases.

src/protocols/Pager/Pager.cpp
jasiek: Yep, will fix.

jasiek: Do we want some kind of error handling in this case, or do we just silently not match anything?

src/protocols/Pager/Pager.cpp
jasiek: Yep.

src/protocols/Pager/Pager.cpp
jgromes: It's a private method, so no need to to report any error and just always return false if addresses are NULL. However, we should make sure it cannot crash ;)

src/protocols/Pager/Pager.cpp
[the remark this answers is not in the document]
jgromes: > do you mean to keep a variable for when there's a single address to service the original version of the method, and then keep a pointer for a list of addresses?

That's probably the simples way to do that. It actually flows nicely into the other comment about checking NULL address list - if it's null, check the address variable. If not, check the address list.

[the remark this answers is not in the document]
jgromes: > What's the use case for masks anyway?

It's very useful for checking continuous address ranges. If you need to match 256 addresses at time, it would be rather cumbersome having to spell them out in a list. Even better example is if you want to match all adresses - just set the mask to zero, no need to provide 2^20 addresses to check. However, this works best for continuous address ranges, that's why your PR is useful.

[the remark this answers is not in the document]
jasiek: That makes sense. If I were to set the mask to receive all messages, is there any risk that some would be dropped with high traffic or that is not a concern at all?

jasiek:
I've updated and force-pushed my changes, you may need to click on the "changes" tab to see it.

src/protocols/Pager/Pager.cpp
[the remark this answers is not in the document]
jgromes: > is there any risk that some would be dropped with high traffic or that is not a concern at all?

I don't think it will be an issue, though I did not test the implementation in a high traffic environment. Definitely not going to be na issue because of the address matching - if you set it to match everything, it will simply just alwasy return true whne checking the address.

jgromes (approved):

jgromes:
Looks good now - merged, thanks!

PR #1002, opened 2024-03-05
Added ESP32_RTC_EEPROM and LoraWAN_ESP32_Deep_Sleep example

ropg:
Allows ESP32 to run LoRaWAN from deep sleep mode without writing a significant chunk of flash every send.

Includes new files `ESP32_RTC_EEPROM.[cpp|h]` in src/utils/, which provide the mechanism to use ESP32's RTC RAM as the backend storage for simulated EEPROM instead of flash. These provide `toNVS()` and `fromNVS()` to backup and restore from flash, as well as automatically restoring from last flash backup if RTC memory was found empty.

The included new example `LoRaWAN_ESP32_Deep_Sleep` and the code comments therein should explain how to use the RTC RAM persistence while still being resilient against reset, power failure, code reflashes, battery changes, etc.

The only proposed change to existing code in this PR involves `#include`ing [ESP32_RTC_EEPROM](https://github.com/ropg/ESP32_RTC_EEPROM) in the case of ESP32 and moving that include from the .cpp to the .h file so that the EEPROM object is exposed to users of the library. This has the added advantage that (both on the RTC RAM version as on the strictly NVS flash one), users are then free to load and save contents to and from 8" floppies or other external sources by just putting and getting `EEPROM.length()` bytes of data from `EEPROM.getDataPtr()`

### DIFF
--- a/examples/LoRaWAN/LoRaWAN_ESP32_Deep_Sleep/LoRaWAN_ESP32_Deep_Sleep.ino
+++ b/examples/LoRaWAN/LoRaWAN_ESP32_Deep_Sleep/LoRaWAN_ESP32_Deep_Sleep.ino
@@ -1,0 +1,196 @@
+/*
+  RadioLib ESP32 Deep Sleep LoRaWAN example
+
+  THIS EXAMPLE ONLY WORKS ON ESP32. It assumes the ESP32 stays powered, but puts
+  it in a very low-power "deep sleep" mode between sends to the network.
+
+  This example joins a LoRaWAN network and will send uplink packets. Before you
+  start, you will have to register your device at
+  https://www.thethingsnetwork.org/ After your device is registered, you can run
+  this example. The device will join the network and start uploading data.
+
+  LoRaWAN v1.1 requires storing parameters persistently. RadioLib does this by
+  using the Arduino EEPROM storage. The ESP32 does not have EEPROM memory, so
+  the Arduino code from Espressif has code that emulates EEPROM in its "NVS"
+  flash key-value storage. This unfortunately means that every change of a
+  single but re- writes the entire emulated "EEPROM" to flash again. 
+
+  That would be great for the occasional configuration change, but we need to
+  keep state after every single interaction with the LoRaWAN network. Flash
+  memory can only be erased and written to a limited number of times. This is
+  why on ESP32 RadioLib stores data using ESP32_RTC_EEPROM, which writes to the
+  memory of the ESP32's onboard Real-Time Clock which is kept powered during the
+  chip's "deep sleep", which uses almost no power.
+
+  This example then backs up that RTC RAM memory to flash
+    a) every time it is woken up with a wiped RTC RAM,
+    b) whenever it receives a downlink packet, or 
+    c) every so many times as set by BACKUP_EVERY.
+
+  This backup is then automatically loaded when the RTC RAM is empty (after a
+  reset or power failure) so that we hang on to enough state information to
+  rejoin the network in a secure manner. 
+
+  For default module settings, see the wiki page
+  https://github.com/jgromes/RadioLib/wiki/Default-configuration
+
+  For full API reference, see the GitHub Pages
+  https://jgromes.github.io/RadioLib/
+
+  For LoRaWAN details, see the wiki page
+  https://github.com/jgromes/RadioLib/wiki/LoRaWAN
+
+
+  Last updated March 2024 for RadioLib 6.4.2
+*/
+
+// Create a backup copy of the RTC RAM to flash every so many times
+#define BACKUP_EVERY 100
+
+// include the library
+#include <RadioLib.h>
+
+// SX1262 has the following pin order:
+// Module(NSS/CS, DIO1, RESET, BUSY)
+SX1262 radio = new Module(8, 14, 12, 13);
+
+// SX1278 has the following pin order:
+// Module(NSS/CS, DIO0, RESET, DIO1)
+// SX1278 radio = new Module(10, 2, 9, 3);
+
+// create the node instance on the EU-868 band
+// using the radio module and the encryption key
+// make sure you are using the correct band
+// based on your geographical location!
+LoRaWANNode node(&radio, &EU868);
+
+// for fixed bands with subband selection
+// such as US915 and AU915, you must specify
+// the subband that matches the Frequency Plan
+// that you selected on your LoRaWAN console
+// LoRaWANNode node(&radio, &US915, 2);
+
+// Variables that are placed in RTC RAM survive deep sleep. This counter tells
+// us how many times we've booted since the last reset or power loss.
+RTC_DATA_ATTR int count = 1;
+
+void setup() {
+  Serial.begin(115200);
+  
+  // initialize radio (SX1262 / SX1278 / ... ) with default settings
+  Serial.print("[Radio] Initializing ... ");
+  int state = radio.begin();
+  if(state == RADIOLIB_ERR_NONE) {
+    Serial.println("success!");
+  } else {
+    Serial.printf("failed, code %i\n", state);
+    while(true);
+  }
+
+  // JoinEUI - previous versions of LoRaWAN called this AppEUI
+  // for development purposes you can use all zeros - see wiki for details
+  uint64_t joinEUI = 0x0000000000000000;
+
+  // DevEUI - The device's Extended Unique Identifier
+  // TTN will generate one for you
+  uint64_t devEUI =  0x----------------;
+
+  // encryption keys used to secure the communication
+  // TTN will generate them for you
+  // see wiki for details on copying & pasting them
+  uint8_t appKey[] = { 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--,   
+                       0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x-- };
+  uint8_t nwkKey[] = { 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--,   
+                       0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x--, 0x-- };
+
+  // Manages uplink intervals to the TTN Fair Use Policy
+  node.setDutyCycle(true, 1250);
+
+  bool gotDownlink = false;
+
+  String strUp;
+  String strDown;
+
+  // Begin the join to the network
+  Serial.print("[LoRaWAN] Attempting over-the-air activation ... ");
+  if (count == 1) {
+    Serial.print(" (forcing join) ");
+    // We're forcing a new join message when we wake up with wiped RTC RAM to
+    // make sure there is no outdated information sent and cmmunications are
+    // restored promptly.
+    state = node.beginOTAA(joinEUI, devEUI, nwkKey, appKey, RADIOLIB_LORAWAN_DATA_RATE_UNUSED, true);
+  } else {
+    state = node.beginOTAA(joinEUI, devEUI, nwkKey, appKey);
+  }
+  if(state >= RADIOLIB_ERR_NONE) {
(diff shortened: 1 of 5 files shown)


StevenCellist:
I think what we should be asking most here, is whether RadioLib should be about anything else than radios. Judging from points mentioned in #999 by @jgromes himself, I'd say that it's not a good idea to burn our fingers on a can of worms. Very sure, the LoRaWAN stack can provide handles to user-supplied callbacks, but @jgromes is not and should not be held responsible for anything EEPROM related. The current state happened but has already multiple times been point of discussion, but with possible far-future support for SAMD and STM32 flash saving for keys & nonces, it would become a swamp if not much worse.

We would need to find a place where example implementations can live, but I recommend outside RadioLib - it's not RadioEEPROMRTCFRAMLib. Possibly @ropg would be so kind as to be care-taker of a to-be-collection of scripts for a range of platforms? :)

ropg:
Peace,

I very much respect the work that's been done here or I wouldn't be here trying to improve on it. I am aware of, apologize for and will try to curb my sometimes a bit overzealous enthusiasm, and I would love to help make things better. I can only offer to help, I understand this is your project much more than it is mine.

Now can we please just all take a deep breath and a step back and look at the situation at hand?

### 1.

We have my PR that, as far as existing code is concerned, literally just replaces

```cpp
#if !defined(RADIOLIB_EEPROM_UNSUPPORTED)
#include <EEPROM.h>
#endif
```
in one file with this in another
```cpp
#if !defined(RADIOLIB_EEPROM_UNSUPPORTED)
  #if defined(RADIOLIB_ESP32)
    #include "utils/ESP32_RTC_EEPROM.h"
  #else
    #include <EEPROM.h>
  #endif
#endif
```

This makes deep sleep possible on ESP32, **doesn't do anything for anyone else**, and if people on ESP32 want to write/erase a significant number of bytes to flash every LoRaWAN send like they've been doing so far they are welcome to, they just add `EEPROM.toNVS()` after each `saveSession()`. It comes with a good and clear example and the added code is well-documented in its own repository. Sure, it adds to a bit to pre-existing complexity , but it's in itself a simple solution and it works.

* I think this PR should go in, even while at the same time realizing this is not how things should be eventually. ESP32 users need deep sleep to work and have their devices last, this code does that and doesn't affect anyone else. The example works like a charm: deep sleeping, force re-joining when it recognizes it has lost some state and not saving to flash 99% of the sends. Yes, I recognize that longer term we need a better solution, and I understand you couldn't do this for 100 different processors this way.*

### 2.

I offer to help abstract this all away, with a somewhat concrete proposal, while also signaling to be very open to different approaches. The reaction, which I will paraphrase as "Let's just get rid of all this stuff, or you do it somewhere else" is not enough elaborated on for me to see how this would work if people want to actually use LoRaWAN with RadioLib. Not just ESP32 users but everyone needs to have these few hundred bytes persist. Sheesj people, it's a few hundred bytes, and there's not 100 different ways for people to store this data. But yes, there are a few different ways people will need to store this data on their platform. **Importantly, some of these methods need insight into what is being stored.**

> to explain: e.g., for ESP32 flash, they will want each separate item in the state as a separate key in NVS. that way each thing that changes changes a limited number of bytes. Still writes every time, but like I said before, this then starts out at ~15x better than what we have. Now it would not make sense (to me at least) to name these NVS flash keys anything other than the names of the entities in LoRaWAN. That means this code is connected to, and IHMO part of, the LoRaWAN code. It is also small and in its own file: 20 or so lines plus one per stored item if I had to guess.

My loose proposal would be to put this persistence interface and the small files per storage method in a directory under LoRaWAN, replace `this->devNonce` by (something along the lines of) `this->persist.devNonce` and have the main LoRaWAN code just `#include persistence/persist.h` and `PersistClass persist` and not worry about anything further. Clean.

This means persistent storage can come out of ArduinoHal (which now forces us to to do byte-level storage without any context, which is explicitly not flexible enough) , all the `#if !defined(RADIOLIB_EEPROM_UNSUPPORTED)` in LoRaWAN go away, as does `saveSession()` and most of `restore`. Instead the user interacts with `node.persist` to save when appropriate, in whatever way is appropriate. Or something similar; I am explicitly not attached to anything specific here, just describing how it *could* work. Users can have their own way of making the data in persist get stored and reappear, but by default the library picks what makes most sense for their platform to make things "just work".

None of the above is a lot of code. None of it is very complex. In fact, `LoRaWAN.cpp` will become *much* more readable once this stuff is gone from there.

Sure, all in all it's a fair bit of work, I recognize that. But I am not proposing something up in the air that other people then have to work on: I'm offering to write this, because I feel things will get better from it. I think doing this in a separate repository makes no sense for reasons outlined above.  

### 3.

We make some persistence interface, but for the reasons above will still need to do essentially the entire above proposal as far as the LoRaWAN.cpp side is concerned. Except now the directory with (guessing) 7 or 9 small files under LoRaWAN is in its own repo. Then what would compile for a system that has non-volatile RAM, or flash, or EEPROM? A version without support for persistence? And then instructions in the README to get this other repo, modify some #defines in a config file and compile again?

&nbsp;

To come back to the solution @jgromes said he preferred in #999:

> We offload this to the user. Keep the current implementation as-is, and provide methods (as @ropg proposed in the original post) to save the session state or even the keys in a persistent storage of their choosing, be it an external FRAM, EEPROM or an 8-inch floppy. In theory this can be done already if the user were to write their own HAL that just overrides the methods for persistent memory management and inherits the rest from Arduino HAL, but it's not very practical. This would allow the user to do their own tradeoff of persistent memory degradation against the risk of losing counters.

Now I would argue that *"provide methods (as @ropg proposed in the original post) to save the session state or even the keys"* is precisely what's implemented in the above, "allowing for 8-inch floppies" (or anything else) is very much part of the plan, except we also provide a few of these methods ourselves so that things just work for the majority of users. But they live in their own directory where they don't unnecessarily clutter up anything else. (And they're small! They store a few bytes!)

StevenCellist:
I've eaten full meals today and conferred on this topic. While the following is the perspective of @HeadBoffin and me, we hope it aligns with @jgromes.

We would suggest the following approach:
- In the LoRaWAN stack, supply two sets of handles:
  1. Setter and getter of a buffer for Keys & Nonces - this should, wherever possible & feasible, be saved in flash or other well usable persistent memory (ESP32, SAMD, AVR, STM32, ...)
  2. Setter and getter of a buffer for the session - this does only need to be used for ESP32 and maybe, I don't know, another platform that would need this (RP2040?). On 'better' platforms, these don't need to be used as RAM is retained through sleep.

  This is all aimed at very very minimal confusion, and allows users to gradually get acquainted with LW. As most users start off with (if we are lucky) the bare minimum, we can guide them from the simplest sketch onto full-fledged v1.1 with the right persistence plugin for their platform.

- Somewhere else than in RadioLib's source, we can collect a set of functions for each platform that can be plugged in by the user and set as callbacks to the handles exposed in the stack. RadioLib has no business being a source for persistence APIs, and even less providing support for anything related to that. So we'd need a supporting repository authored by (a group of) devs with more knowledge on one or more platforms.

There are more details to be discussed, but I don't want to get lost on those before we can find an agreement on the main topics.

---

You'll see that we are not so concerned with the urgence of the ESP32 flash wear as you are. I'm willing to bet that I am the most affected person in the world right now, with a number of active devices that requires more than one hand - on tens of thousands of uplinks. Why no hurry? Because the currently released 6.4.2 even includes an error in the FcntUp restore function that renders ESP32s using deepsleep pretty much unusable after a couple of hours. I've only seen two people here talking about that problem...

HeadBoffin:
> I think this PR should go in, even while at the same time realizing this is not how things should be eventually. 

As this is not yet out of beta and not recommended for production use, we are at a point where we do not need to accrue additional technical debt.

> The reaction, which I will paraphrase as "Let's just get rid of all this stuff, or you do it somewhere else" 

Ouch!

The response is "let's not complicate RadioLib with stuff the user can handle for one of the many protocols - it's important for LW, particularly for ESP32 deep-sleep, so let's provide some working examples".

> is not enough elaborated on for me to see how this would work if people want to actually use LoRaWAN with RadioLib.

Whilst not for production use, it is working just fine on a range of hardware and was working OK for EU868 before @StevenCellist took up the baton of making it more spec compliant.

> Not just ESP32 users but everyone needs to have these few hundred bytes persist. 

**I think this must be about the fourth or fifth time I've said this, but actually mostly not, it is only an ESP32 issue.** There are downsides to not having persistence for LW 1.0.4 and 1.1 but it's mostly an inconvenience upon battery change, which is a pretty rare event. Hundreds of thousands of uplinks later, my DIY #2 device has only had one battery change last summer and it took about five minutes to sort its join out.

But even then, **there is full support for providing persistence** because the end goal is to run the stack through the official compliance tests, it's just that there is little appetite for embedding it in to RadioLib specifically for ESP32.

If the getters & setters are available along with some examples on a separate repro, this covers the majority of use cases. If someone then wants to implement external storage or use an SD Card or punched tape, they can use the API to implement it themselves.

If persistence is embedded in to RadioLib, then people have to dig under the hood to implement their own storage, which is rather not the point of a library.

It is great that you, @ropg, drew our attention to the flash write system on ESP32. Why not run with creating a few devices with the fork you have, learn LoRaWAN in practise and help out with the design of a universal solution for persistence. And also consider that your own persistence is taking up time that could be used on actual progress.

ropg:
@StevenCellist @HeadBoffin @jgromes

# 👍

Looks very reasonable to me.  Using proposed persistence blobs would indeed be easy enough, assuming it would simply rejoin if it sees blob one but not blob two. 

First thoughts:

* You would probably want to exclude FCntUp from the blob and create a separate getter and setter for it, so that people that do write it all to EEPROM can stay agnostic to blob contents and either use a wear-levelling trick on this only value that needs it, or, say, write it to ESP32 NVS in a separate key so that only 32 bytes (the smallest chunk in NVS) are written each time. This would already, on ESP32, reduce the penalty for keeping everything in flash 15-fold.
* I think this is indeed all the granularity needed, assuming no other single parts in either of these blobs changes very frequently, which you said wasn't the case.

> Because the currently released 6.4.2 even includes an error in the FcntUp restore function that renders ESP32s using deepsleep pretty much unusable after a couple of hours.

Well, I will say you had me looking there for a bit, but as far as I can see that's either not true or it's so subtle I can't find it. Both are possible, but I ran `saveFcntUp()` and `restoreFcntUp()` on a counter even, works like a charm as fas as I can tell. What gives?

> Why not run with creating a few devices with the fork you have, learn LoRaWAN in practise and help out with the design of a universal solution for persistence.

Have two test devices running, sometimes being reset by me to see how/if they deal with that. I'll gladly help out with whatever is needed.

jgromes:
@StevenCellist I fully agree with the approach that you outlined. If I understand it correctly (unfortunately don't have the time to follow all the quite in-depth discussions), it would mean getting rid of persistent storage from RadioLib completely, which should be done anyway. I can now clearly see it was a mistake on my part to [add persistent storage at all](https://github.com/jgromes/RadioLib/commit/e486829b8f432ca12b2c91d8aee78f4d4270ab73). 

As to the where to put the platform- and/or memory-specific functionality, in my opinion it really should be in another repository. To keep it close to RadioLib (since we will probably be referencing it from LoRaWAN examples), I think it would be best to put it up in some organization. It would also be the better to transfer RadioLib to org as well, in due time. But first things first, let's resolve the persistent stuff and get the update for LoRaWAN examples for the 6.4.3 release.

StevenCellist:
No mistake! If there wasn't any form of persistence, I would've given up instantly because I'm purely ESP32 ;)

We've just speedran into the thick brick wall that would eventually result in the same.

jgromes:
Since LoRaWAN persistence handling was reworked in #1017, I will close this PR now. New examples on how to use LoRaWAN persistence will be added before the 6.5.0 update (though I'm sure intertested parties will be able to figure it out regardless).

Thanks @ropg for bringing our attention to this issue, albeit in a rather ... persistent manner ;)

PR #1014, opened 2024-03-13
Arduino string fixes

ropg:
* Arduino Strings can contain null chars, but Strings provided to uplink() were truncated at first null, and downlink messages were only copied into the provided String until the first null. This fixes that.
* A sendReceive function is provided for when the user doesn't care about user downlink messages. This provides the fitting Arduino String equivalent.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -1690,6 +1690,16 @@ int16_t LoRaWANNode::sendReceive(String& strUp, uint8_t port, String& strDown, b
   state = this->downlink(strDown, eventDown);
   return(state);
 }
+
+int16_t LoRaWANNode::sendReceive(String& strUp, uint8_t port, bool isConfirmed, LoRaWANEvent_t* eventUp, LoRaWANEvent_t* eventDown) {
+  // send the uplink
+  int16_t state = this->uplink(strUp, port, isConfirmed, eventUp);
+  RADIOLIB_ASSERT(state);
+
+  // wait for the downlink
+  state = this->downlink(eventDown);
+  return(state);
+}
 #endif
 
 int16_t LoRaWANNode::sendReceive(uint8_t* dataUp, size_t lenUp, uint8_t port, bool isConfirmed, LoRaWANEvent_t* eventUp, LoRaWANEvent_t* eventDown) {
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -1335,6 +1335,7 @@ int16_t LoRaWANNode::downlink(String& str, LoRaWANEvent_t* event) {
 
     // Copy data by appending so char(0) not lost. 
     str = ""; // Empty the String that was passed in
+    str.reserve(length);
     for (int i = 0; i < length; i++) {
       str += (char)data[i]; // Append each byte as a char
     }
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -926,8 +926,8 @@ void LoRaWANNode::saveFcntUp() {
 #endif  // RADIOLIB_EEPROM_UNSUPPORTED
 
 #if defined(RADIOLIB_BUILD_ARDUINO)
-int16_t LoRaWANNode::uplink(String& str, uint8_t port, bool isConfirmed, LoRaWANEvent_t* event) {
-  return(this->uplink(str.c_str(), port, isConfirmed, event));
+int16_t LoRaWANNode::uplink(String& str, uint8_t port, bool isConfirmed, LoRaWANEvent_t* event) {  
+  return(this->uplink((uint8_t*)str.c_str(), str.length(), port, isConfirmed, event));
 }
 #endif
 
@@ -1325,18 +1325,20 @@ int16_t LoRaWANNode::downlink(String& str, LoRaWANEvent_t* event) {
   int16_t state = RADIOLIB_ERR_NONE;
 
   // build a temporary buffer
-  // LoRaWAN downlinks can have 250 bytes at most with 1 extra byte for NULL
+  // LoRaWAN downlinks can have 250 bytes
   size_t length = 0;
-  uint8_t data[251];
+  uint8_t data[250];
 
   // wait for downlink
   state = this->downlink(data, &length, event);
   if(state == RADIOLIB_ERR_NONE) {
-    // add null terminator
-    data[length] = '\0';
 
-    // initialize Arduino String class
-    str = String((char*)data);
+    // Copy data by appending so char(0) not lost. 
+    str = ""; // Empty the String that was passed in
+    for (int i = 0; i < length; i++) {
+      str += (char)data[i]; // Append each byte as a char
+    }
+
   }
 
   return(state);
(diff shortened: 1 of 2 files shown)


ropg:
Arduino Strings have a `.length()` and are [internally stored as a buffer and a length](https://github.com/arduino/ArduinoCore-API/blob/0c853c5cded2768122fae258d42b2b4c06cdb3b1/api/String.h#L222). They are also null-terminated, but that's only so that they can be (read-only) provided as `char*` by calling `.c_str()`, with the caveat that whatever is handling that will terminate at the first null, naturally. I frequently use Strings to move binary data around, precisely because it has a length field as opposed to using char*.

```cpp
String test = "Hello";
test += '\x00';
test += "World";
Serial.println(test);
```
prints
```
HelloWorld
11
```
as it should.

Given the fact that LoRaWAN is meant to transfer short binary messages, it seemed relevant to be able to handle binary data wherever possible. I did check all the occurrences of Arduino Strings in LoRaWAN to see if there were any other issues, but uplink, downlink and sendReceive are the only functions touching String: 6 occurrences of "String" in the file in total.

As for the missing function: I'll make a similar case two PRs next time. They seemed both Arduino-String-sendReceive-related and the function seemed obviously forgotten/missing.

StevenCellist:
What about.. we just get rid of Strings in general? Enough folks that don't like 'em, LoRaWAN is also never meant to support Strings, all other functions already provide the length if any char array, and the compiler has enough (too much) trouble with the various overloads and argument default values already.
@jgromes and @HeadBoffin.

@ropg, please please, open up an Issue or Discussion on topics like these before submitting a PR out of the blue. I have a massive update pending to support your finding on the EEPROM issue, which is currently being tested. If you give a heads-up, we can schedule it on the roadmap instead of being forced to handle things in between & during development with resulting merging issues and what not. Thanks in advance!

jgromes:
> I frequently use Strings to move binary data around

Good Lord. I think you are profoundly misusing what was originally meant to be a "simple-to-use" string class for people who came from Java or Python and don't want to learn how strings and arrays work in C. Arduino `String` class has sprung many a memory leak over the years, and its use is explicitly prohibited in RadioLib except for public API (see the [contributing doc](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md)). For example, you have this in your PR:

```
  // Copy data by appending so char(0) not lost. 
    str = ""; // Empty the String that was passed in
    for (int i = 0; i < length; i++) {
      str += (char)data[i]; // Append each byte as a char
    }
```

Every single time you `+=` to the `String`, you call `realloc` just to increase the buffer length by one. That is horribly inefficient and will fragment your heap in no time. 

> Given the fact that LoRaWAN is meant to transfer short binary messages, it seemed relevant to be able to handle binary data wherever possible. 

I agree. That's why all string methods in RadioLib ultimately lead to `uint8_t* buff, size_t len` methods.

I think you are missing the fact that RadioLib is not an Arduino library. It's a common abstraction platform

ropg:
> What about.. we just get rid of Strings in general?

Fine by me. Given the desire to have LoRaWAN messages be binary and not JSON, you could even ditch char* as well and only support uint8_t* with length.

> @ropg, please please, open up an Issue or Discussion 

OK, will do that next time. It just seemed easy to fix.

(FWIW: I think this will merge fine either before or after what you're working on.)

ropg:
> ```
>   // Copy data by appending so char(0) not lost. 
>     str = ""; // Empty the String that was passed in
>     for (int i = 0; i < length; i++) {
>       str += (char)data[i]; // Append each byte as a char
>     }
> ```
> 
> Every single time you `+=` to the `String`, you call `realloc` just to increase the buffer length by one. That is horribly inefficient and will fragment your heap in no time.

Ah, you're right. I forgot to add what String has for that: `.reserve()`. That's in the PR now.

jgromes:
@ropg 

> Ah, you're right. I forgot to add what String has for that: .reserve(). That's in the PR now.

You're missing my point. It's incredibly easy to use `String` very incorrectly, as you just did. RadioLib only uses it in its public Arduino interface, because there's a lot of Arduino users who get very confused when a method that takes `char*` refuses to take `String`. Nevertheless, I remain unconvinced that Arduino Strings should contain `\0` bytes for any reasonable use case. It's not prohibited by the `String` class (I would argue it should be), and the header file does seem to suggest it was an intention to use `\0` as terminator. If it was meant to include `\0`, why keep the final terminating one in the internal buffer?

https://github.com/arduino/ArduinoCore-API/blob/0c853c5cded2768122fae258d42b2b4c06cdb3b1/api/String.h#L219-L222

Closing because I don't think we should be supporting what is at best unintentional side-effect.

@StevenCellist 

> What about.. we just get rid of Strings in general?

I'd love to, but unfortunatly that will most likely open a floodgate of Arduino users opening issues about missing `String` and `F()` support or claiming on other sites/forums that RadioLib doesn't support strings. I agree the endless overloads are not a great solution, but I have yet to find a better one.

HeadBoffin:
And in all of this back and forth, the examples are being updated, as evidenced in my current PR re the Starter.ino, to drop the Strings and use byte arrays.

As part of the examples update we have already put in place a sendReceive that effectively hides the downlink calls which has been used to simplify the Starter.ino

In the meanwhile we have been reworking the persistence using the internal functions of the Arduino-ESP32, thank you for your prior observations.

Hopefully you can find some patience & restraint to let us complete this phase, at which point we can then use discussions to see what would help RadioLib users implementing LoRaWAN next. 

I'm sure you will agree that prior experience of implementing LoRaWAN & supporting those implementing it will be of benefit in providing insightful input. If you want to get some flavour of where people come unstuck, searching the TTN forum for LMIC will provide you with plenty of examples.

HeadBoffin:
> They seemed both Arduino-String-sendReceive-related and the function seemed obviously forgotten/missing.

Ouch! Please stop talking like you know what's inside our heads.

@jgromes did much of the groundwork on the protocol and @StevenCellist has done the detail work on the MAC commands and the subsequent issues arising to get it as close to compliance as reasonably possible in short order. As we can see the gaps that aren't appropriate to let run out in the wild, we elected to mark the protocol as beta, which it still is, mostly because of ESP32 persistence as @StevenCellist has a huge pile of Air Quality Monitors based on Heltec ePaper modules for his students that will benefit from RadioLib. So the things that you see missing may well be work in progress, if you just ask first.

> (FWIW: I think this will merge fine either before or after what you're working on.)

How could you possibly know that? How? I mean how could you possibly come to such a conclusion? If you looked at my PR you'd see that can't be the case. You need to look around you to see what others are doing rather than pressing on at speed.

ropg:
It seemed obvious to me that if the user gets a 20 byte message, he/she wants strDown.length() to be 20, not some number up until the first zero byte in the message. And in the event the user supplies a strUp with a zero byte in it, the user probably wants that byte sent. But I do not want to impose, and if y'all think it's better the other way, then please do it the other way.

But whatever you do, can you please assume I mean well? I said "FWIW" and "I think", naturally I cannot know, nor did I imply to know, what you plan to do. That doesn't mean I'm "pressing" for anything, I merely (and ONLY) tried to be helpful and fix what I thought was a bug.

As requested, I shall not be filing PRs and (at most) file an issue or discuss if I see anything further.

HeadBoffin:
FWIW is code for "in my opinion" - the fact is that it wouldn't.

Once more you try to extend your position after having the 'why not' explained, this is wasting everyone's time, yours included. Even if you had several years experience with different LW stacks this would still be irritating. Please consider that we are working on a release that has been badly delayed with all this to'ing & fro'ing. Once it is done we can then form a huddle to figure out next steps, but until then give us some space to get there.

PR #1017, opened 2024-03-16
[LoRaWAN] Change and upgrade persistence handling

StevenCellist:
Description following soon... yes I know about ```BuildOpt.h```, will resolve.
Ready for review, tested EU868 with OTAA and ABP, tested US915 with OTAA.


StevenCellist:
Due to the problem reported by @ropg (thank you!), extensive work has been put into remodelling the handling of session persistence. This PR aims to remove any and all support of platform-specific code for handling the required persistence. This will be handled in https://github.com/radiolib-org/radiolib-persistence.

The solution in this PR is as follows - the stack exposes two buffers, namely:
1. a 'Nonces' buffer (size: 16 bytes) that includes the Nonces of the OTAA sessions, as well as integrity values that keep track of whether these Nonces match the current configuration of the device, and
2. a 'Session' buffer (size: 396 bytes) that contains all necessary details of the currently active session, such as all counters and the complete MAC state.

Both buffers include checksums, and the session buffer includes the checksum of the Nonces buffer and requires that these two match each other. This makes sure that the buffers cannot easily be (accidentally) tempered with.

@HeadBoffin is working on examples that showcase how these buffers can be used for nicely working persistence of ESP32. 

**Converting this into some sort of persistence library will come at a later point in time in the other repository and is not the aim of this PR.** It can be handled independently in the other mentioned repository.

A relevant changelog will follow later.

_This PR also deletes all references to anything related to EEPROM throughout the source files._

HeadBoffin:
> Converting this into some sort of persistence library will come at a later point in time 

If it transpires a persistence library is required - the ESP32 example we are using for testing uses native Arduino-ESP32 APIs with the aim that examples become available for AVR, SAMD, STM32 etc that uses the BSP APIs.

I'm not adverse to a library, but native APIs are looked after by a very large number of eyeballs - if we start creating additional libraries we end up with update work.

I'm aiming to put the ESP32 persistence example in to my WIP PR Sun 16th so we can join up everything.

ropg:
Great work!  👍

StevenCellist:
* [LoRaWAN] Change session persistence from integrated EEPROM to exposed buffers
* [LoRaWAN] Make the return codes from `beginOTAA()` and `beginABP()` zero again (ERR_NONE)
* [LoRaWAN] Swap the arguments to `beginOTAA()`
* [LoRaWAN] Include frequency plan and device class into integrity checks
* [LoRaWAN] Fix bugs in MAC state restoring
* [LoRaWAN] Fix #1018 incorrect loop
* [LoRaWAN] Improve checksum calculation by working on pointers instead of creating new buffers
* [LoRaWAN] Decrease debug output significantly (full restore + uplink now consumes about 100 lines on EU868/OTAA instead of closer to 200)
* [BuildOpt] Remove EEPROM support code
* [HAL] Remove EEPROM support code
* [ArduinoHAL] Remove EEPROM support code
* [TypeDef] Add checksum mismatch error

StevenCellist:
@jgromes not sure about your CI checks, but with the removal of EEPROM, it may be the case that you can add LoRaWAN back to some of the platforms for which it was disabled?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -48,172 +44,208 @@ void LoRaWANNode::setCSMA(uint8_t backoffMax, uint8_t difsSlots, bool enableCSMA
...
   if(this->activeMode != RADIOLIB_LORAWAN_MODE_NONE) {
-    return(this->activeMode);
+    return(RADIOLIB_ERR_NONE);
   }
 
-  Module* mod = this->phyLayer->getMod();
+  bool isSameKeys = LoRaWANNode::ntoh<uint16_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_CHECKSUM]) == checkSum;
+  bool isSameMode = LoRaWANNode::ntoh<uint16_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_MODE]) == lwMode;
+  bool isSameClass = LoRaWANNode::ntoh<uint8_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_CLASS]) == lwClass;
+  bool isSamePlan  = LoRaWANNode::ntoh<uint8_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_PLAN]) == freqPlan;
```
jgromes, line 136: Appreciate the consistency, but there's no need to convert single byte from network endianity to host endianity ;)

EDIT: this is also present at a few more places.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -48,172 +44,208 @@ void LoRaWANNode::setCSMA(uint8_t backoffMax, uint8_t difsSlots, bool enableCSMA
...
+  this->aFcntDown    = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_A_FCNT_DOWN]);
+  this->nFcntDown    = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_N_FCNT_DOWN]);
+  this->confFcntUp   = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_CONF_FCNT_UP]);
+  this->confFcntDown = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_CONF_FCNT_DOWN]);
+  this->adrFcnt      = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_ADR_FCNT]);
+  this->fcntUp       = LoRaWANNode::ntoh<uint32_t>(&this->bufferSession[RADIOLIB_LORAWAN_SESSION_FCNT_UP]);
   
-  // fcntUp is stored in highly efficient wear-leveling system, so parse it
-  this->restoreFcntUp();
+  int16_t state = RADIOLIB_ERR_NONE;
```
jgromes, line 184: General comment - in most cases, when initializing `state`, I set it to some error code (e.g. `RADIOLIB_ERR_UNKNOWN`) so that the next assert will fail if we do not assign it some new value (e.g. by calling a function that returns an OK code).

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -48,172 +44,208 @@ void LoRaWANNode::setCSMA(uint8_t backoffMax, uint8_t difsSlots, bool enableCSMA
...
-    }
-  }
-  uint32_t bits_7_0 = (uint32_t)fcntBuff[idx-1] & 0x7F;
-  #if !RADIOLIB_STATIC_ONLY
-  delete[] fcntBuff;
-  #endif
+  this->activeMode = LoRaWANNode::ntoh<uint16_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_MODE]);
 
-  this->fcntUp = (bits_30_22 << 22) | (bits_22_14 << 14) | (bits_14_7 << 7) | bits_7_0;
+  return(RADIOLIB_ERR_NONE);
```
jgromes, line 248: This should return `state` - I do realize that the only way for the program to get to here is if `state == RADIOLIB_ERR_NONE`, however, that could change in the future, even if by mistake.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -48,172 +44,208 @@ void LoRaWANNode::setCSMA(uint8_t backoffMax, uint8_t difsSlots, bool enableCSMA
...
   LoRaWANMacCommand_t cmd = {
-    .cid = RADIOLIB_LORAWAN_MAC_LINK_ADR,
+    .cid = RADIOLIB_LORAWAN_MAC_TX_PARAM_SETUP,
     .payload = { 0 }, 
-    .len = MacTable[RADIOLIB_LORAWAN_MAC_LINK_ADR].lenDn,
+    .len = MacTable[RADIOLIB_LORAWAN_MAC_TX_PARAM_SETUP].lenDn,
     .repeat = 0,
   };
+  memcpy(cmd.payload, &this->bufferSession[RADIOLIB_LORAWAN_SESSION_TX_PARAM_SETUP], cmd.len);
+  execMacCommand(&cmd);
```
jgromes, line 201: Pedantic: If we are not using the return value of this method, it should be explicitly cast to `void`, e.g.: `(void)execMacCommand(&cmd);` - see [Power of ten rule 7](https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code) or the [NASA JPL Coding standard](https://yurichev.com/mirrors/C/JPL_Coding_Standard_C.pdf). Even if these are for C and not C++, most of the rules can be applied in general, it is still a great guideline and extremely brief as far as space standards go ;).

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -710,28 +673,31 @@ int16_t LoRaWANNode::beginOTAA(uint64_t joinEUI, uint64_t devEUI, uint8_t* nwkKe
...
+  LoRaWANNode::hton<uint16_t>(&this->bufferNonces[RADIOLIB_LORAWAN_NONCES_SIGNATURE], signature);
+
   return(RADIOLIB_ERR_NONE);
 }
 
 int16_t LoRaWANNode::beginABP(uint32_t addr, uint8_t* nwkSKey, uint8_t* appSKey, uint8_t* fNwkSIntKey, uint8_t* sNwkSIntKey, bool force) {
   // if not forced and already joined, don't do anything
   if(!force && this->isJoined()) {
-    return(this->activeMode);
+    return(RADIOLIB_ERR_NONE);
```
jgromes, line 697: Since this is not distinguishable by the return code, I would suggest to put a debug print here to show that we are joined already.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -2504,7 +2351,7 @@ bool LoRaWANNode::execMacCommand(LoRaWANMacCommand_t* cmd, bool saveToEeprom) {
       float freq = (float)freqRaw/10000.0;
       uint8_t maxDr = (cmd->payload[4] & 0xF0) >> 4;
       uint8_t minDr = cmd->payload[4] & 0x0F;
-      RADIOLIB_DEBUG_PROTOCOL_PRINTLN("New channel: index = %d, freq = %f MHz, maxDr = %d, minDr = %d", chIndex, freq, maxDr, minDr);
+      // RADIOLIB_DEBUG_PROTOCOL_PRINTLN("NewChannelReq: index = %d, freq = %f MHz, maxDr = %d, minDr = %d", chIndex, freq, maxDr, minDr);
```
jgromes, line 2354: Not sure why is this commented out?

jgromes (commented):
@StevenCellist great work as always! I left some comments but none of them are major, more suggestions on improvements ;)

jgromes:
@StevenCellist 

> with the removal of EEPROM, it may be the case that you can add LoRaWAN back to some of the platforms for which it was disabled?

Definitely, some of them blocked only the persistent examples. I will try that manually after this is merged. The otehr ones are disabled due to program/data memory limitations (e.g. Arduino Uno), I don't think those will fit unfortuntely, even if we could have saved some memory with this PR.

jgromes:
@HeadBoffin 

> native APIs are looked after by a very large number of eyeballs - if we start creating additional libraries we end up with update work.

I agree, perhaps we will end up with a repository of examples for different platforms (instead of a persistence library) - we'll see in due time. In my opinion that is OK as well, since the other option is having those examples in RadioLib, which is probably not the way this should be handled.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Mostly a matter of preference.. speed vs consistency indeed. But I interpret you have a preference for speed in this case ;)

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Mostly because a few lines down, the resulting new channel is printed, and the output then is almost exactly duplicated.. so it's just saving some clutter. But I did not completely remove that line.. any preference?

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Actually, I'd say we leave it like this. If there is future reason to convert this into e.g. uint16, it is much more likely that one forgets to convert it properly. Unless you have reasons to do change it.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: If the information is captured somewhere, I would prefer to remove this line. Otherwise it creates confusion why is that commented-out line there. If it is needed, why is it commented out? And if it is not needed, why does it stay there? It's this weird middle-ground that creates room for interpretation.

StevenCellist:
Warning: untested yet - am not at my desk.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: That's a valid point, let's leave it as-is nad hopefully more future-proof.

StevenCellist:
Bugfixed and tested thanks to @HeadBoffin.

jgromes:
All looks good to me - good job, big task off the todo list @StevenCellist and @HeadBoffin!

PR #1021, opened 2024-03-18
[LoRaWAN] Resolve warnings, fix bugs for fixed bands

jgromes:
Found when manually checking LoRaWAN builds for platforms that were previously disabled in CI. These were mostly unused variables, however, random selection of join DR was also unused, so maybe that was a mistake?

cc @StevenCellist


StevenCellist:
Ah, the `drJR` is now incorrectly 'forgotten' indeed. That function was using internal MAC commands which was a mistake (it overwrote the active state with the default channels), but apparently I also removed the MAC command to configure the required join-datarate. Will look at a fix.

jgromes:
@StevenCellist thanks! You should be able to push to this branch, so feel free to add the fix here.

StevenCellist:
Turns out that there was some more stuff to do: there was a bug in processing the CFList that caused the Tx power to drop by 14dBm upon a join, and not all scenarios w.r.t. selecting an initial datarate were handled. Changes tested on EU868/ABP and US915/OTAA.

StevenCellist:
Hold on a minute and I got one small thing as well :)

StevenCellist:
Good to go now from my perspective

jgromes:
Perfect, thanks for the help!

PR #1033, opened 2024-03-24
PQT = 0 in setPromiscuous

Crsarmv7l:
Sets PQT to 0 in setPromiscuous(true)

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -826,7 +826,7 @@ int16_t CC1101::setCrcFiltering(bool enable) {
   }
 }
 
-int16_t CC1101::setPromiscuousMode(bool enable) {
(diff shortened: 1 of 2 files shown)


Crsarmv7l:
I included the bool for the two carriersense options for promiscuous that I mentioned in the discussion. If you are open to it I can submit both this and that as one pull request.

jgromes:
Thanks for the contribution! I think the option for carrier sense can be added directly in this PR. Please make sure it is set to to `false` when calling `setPromiscuousMode` without arguments.

Crsarmv7l:
> Thanks for the contribution! I think the option for carrier sense can be added directly in this PR. Please make sure it is set to to `false` when calling `setPromiscuousMode` without arguments.

Should be good to go, let me know if there are any issues. Thank you for your patience, willingness to help and acceptance of some small but nice to have changes!

jgromes:

PR #1034, opened 2024-03-25
LW examples

HeadBoffin:
Just the Starter, ABP and Reference.

There is a winkle with ABP nomclementure & the order of parameters between the TTN console and beginABP. The config.h matches the Overview on console but the parameter names don't currently match under the hood. This will only be spotted by someone looking at the code before @StevenCellist corrects it.

There is also the possibility that we can drop the force=true on beginOTAA if there is no session restored from persistence.

Despite the above, this moves things forward and more importantly, some actual proper device testing can be done on other platforms (AVR, SAMD, STM32) and, for the other repro, persistence as well as working on the wiki.

Feel free to enact any changes to get these over the wire please.


StevenCellist:
@HeadBoffin @jgromes looks all good to me. I prepared a README for the `examples/LoRaWAN` folder that I will add as soon as the PR is merged.

examples/LoRaWAN/LoRaWAN_Starter/notes.md
```diff
@@ -0,0 +1,202 @@
...
+
+Additionally, as The Things Stack Sandbox aka TTN is an array of servers in three locations around the world paid for by The Things Industries, there is a Fair Use Policy so that those learning LoRaWAN, communities, hobbyists & makers are guided on how much of the resource any one device can use. In short, it's 30 seconds of airtime a day and 10 downlinks. When a gateway is transmitting a downlink it can not hear any uplinks (contributing to the potential uplink loss outlined above). The community concensus is that 1 downlink a fortnight to update or adjust settings is appropriate. See https://www.thethingsnetwork.org/docs/lorawan/duty-cycle/#fair-use-policy for more information.
+
+You can see what intervals can be used with this interactive calculator: https://avbentem.github.io/airtime-calculator/ttn/. Devices further away from gateways will have to use a higher Spread Factor to be heard - do not assume everything will happen at SF7. An uplink takes a minimum of 6 seconds from start to end, sometimes longer if the device is further away from the gateway, so
```
StevenCellist, line 110: Drumroll... 🥁 cliff hanger...

jgromes (approved):
Everything looks good to me. If there are no further changes then I think we can merge this and finally get 6.5.0 out!

StevenCellist:
That wasn't a nice squashed merge, oops, sorry, but it's in :)

PR #1035, opened 2024-03-25
[CI] Fix for new LoraWAN examples

jgromes:
Added support for new LoRaWAN examples in CI. It now builds as can be seen e.g. here: https://github.com/jgromes/RadioLib/actions/runs/8425458582/job/23071608049

All config.h-defined `RADIOLIB_LORAWAN_*` macros are set to 0 during build in CI.

Heads up - I had to do two small changes in the config.h files:

1. Uncomment SX127x in case pinout detection fails, so that `radio` is not undefined in CI - I guess the intention there was to force the user to select a module/pinout appropriate for them. But defining this from CI would require additional ifdefs, I don't think it's wort the extra trouble.
2. Change the `arrayDump` so that it doesn't use `HardwareSerial::printf` which may or may not be available

I think both of these are quite minor, @StevenCellist and/or @HeadBoffin please let me know if those changes are OK; asking since the first one slightly changes the default behavior of the example.


StevenCellist:
I suggest acting as if a board was defined, e.g. a Heltec or LilyGo board. 
E.g. `-D ARDUINO_heltec_wifi_kit_32_V3`

HeadBoffin:
All good, having CI run is important - we can flag a radio not initialised error in conjunction with the "pinmap 101". The arrayDump isn't called from any examples at this point, I typically use it to run the firmware without the radio and then copy & paste the printed payload in to a spreadsheet that can verify the payload, or the console payload formatter or in to the back end etc etc. I guess it could actually be removed for now if considered redundant.

HeadBoffin:
> I suggest acting as if a board was defined, e.g. a Heltec or LilyGo board. E.g. `-D ARDUINO_heltec_wifi_kit_32_V3`

"Bloody hell," said Majikthise, "now that is what I call thinking. Here Vroomfondel, why do we never think of things like that?" - "Dunno," said Vroomfondel in an awed whisper, "think our brains must be too highly trained Majikthise."

Issue #622, opened 2022-11-25
Getter methods for all setters


jgromes:
### Discussed in https://github.com/jgromes/RadioLib/discussions/616

<div type='discussions-op-text'>

<sup>Originally posted by **MrSniffer** November 21, 2022</sup>
Perhaps I missed it, but in searching the headers for the SX1262, I do see a Set and Get for CurrentLimit, but only a Set for OutputPower. 

Is there a way to return the OutputPower setting that is in effect?</div>

jgromes:
Seeing as this has been open for several years and there has not been much requests for it, I will close it as wontfix.

PR #1038, opened 2024-03-27
3 New CC1101 Functions

Crsarmv7l:
Not sure if you are interested in merging these, but figured I would offer.

There are 3 new functions:
getFrequency();
getBitRate();
CalcRxBandwidth();

Nothing spicy for the first two, they basically just return the variables for Freq and Bitrate. I can see them being useful to others though, if someone want to scan a freq range, and/or saving a signal you would need those two.

I really only put them in for the third one. CalcRxBandwidth() is an implementation of atlas0fd00m's calculatePktChanBW function in RfCat which works fairly well, based on the freq, baud and ppm. I tweaked it to work for this application and the CC1101 tolerances.

You can find his here: https://github.com/atlas0fd00m/rfcat/blob/07f99bbb9e2cdbde8658c9446fa0923d3a744d79/rflib/chipcon_nic.py#L973


src/modules/CC1101/CC1101.cpp
```diff
@@ -489,6 +499,25 @@ int16_t CC1101::setRxBandwidth(float rxBw) {
...
+    uncertainty = (uncertainty/1000); //Since bitrate is in kBit
+    float minbw = (getBitRate() + uncertainty);
+    
+    int possibles[16] = {58, 68, 81, 102, 116, 135, 162, 203, 232, 270, 325, 406, 464, 541, 650, 812};
+    
+    for (int i = 0; i < 16; i++) {
+      if (possibles[i] > minbw) {
+        int16_t state = setRxBandwidth(possibles[i]);
+        return(state);
+        break;
```
jgromes, line 515: This break statement is unreachable and should be removed.

src/modules/CC1101/CC1101.cpp
```diff
@@ -489,6 +499,25 @@ int16_t CC1101::setRxBandwidth(float rxBw) {
...
+      }
+    }
+
```
jgromes, line 518: Return value from this method is missing, it must return something (e.g. `RADIOLIB_ERR_UNKNOWN` or some "invalid bandwidth" error code if it never breaks out of the loop.

src/modules/CC1101/CC1101.h
```diff
@@ -745,6 +758,14 @@ class CC1101: public PhysicalLayer {
...
     int16_t setRxBandwidth(float rxBw);
 
+     /*!
+      \brief calculates and sets Rx bandwidth based on the freq, baud and freq uncertainty.
+      \Reimplement of atlas0fd00m (RfCat) CalculatePktChanBw function. 
+      \Modified for worse ppm with the CC1101, and adjusted for the supportted CC1101 bw.
+      \returns \ref status_codes
+    */
+    int16_t CalcRxBandwidth();
+
```
jgromes, line 768: The name of this method seems misleading - I would expect a method that is called "calc" to calculate something, not calculate and set it. I propose to change this to `autoSetRxBandwidth`.

Also, the leading `\` is not supposed to be on every line - it denotes commands for Doxygen, which generates documentation from these comments. So the leading `\` should be removed from the second and third line of the comment.

Finally, the capitalization is incorrect, RadioLib uses `camelCase` in public methods.

src/modules/CC1101/CC1101.h
```diff
@@ -730,13 +730,26 @@ class CC1101: public PhysicalLayer {
...
     */
     int16_t setBitRate(float br);
 
+    /*!
+      \brief gets the currently set bit rate.
+      \returns the bit rate in kbps
+    */
+    float getBitRate();
+
+
```
jgromes, line 752: Both of these methods fall under https://github.com/jgromes/RadioLib/issues/622, I'd rather solve that more globally for all modules. In the meantime, please remove them and use `this->bitRate` and `this->frequency` in the Rx BW method.

src/modules/CC1101/CC1101.h
jgromes: Also, public methods must be added to keywords.txt - be careful that the file uses true tabs and not spaces!

jgromes (changes requested):
Thank you, I left some comments, though nothing major. If those can be addressed, I think we can merge.

src/modules/CC1101/CC1101.cpp
Crsarmv7l: Done

src/modules/CC1101/CC1101.cpp
Crsarmv7l: I added a return value, but that return value will never be reached (why I didn't add it initially). You have checks for both Freq and bitrate and the only way to get through the loop without a bw being selected is if you are well outside the hw scope. 

I made that mistake years ago with my Yardstick one, accidentally set it to 43.392 instead of 433.92

src/modules/CC1101/CC1101.h
Crsarmv7l: All 3 complete. 

I am editing on github itself, so hopefully when setting the Tab drop down it is an actual tab.

src/modules/CC1101/CC1101.h
Crsarmv7l: Removed

src/modules/CC1101/CC1101.cpp
jgromes: Yes, if everything works correctly, the for loop will return. However, if for whatever reason the configured bandwidth is nonsense, the loop may not return, ehnce why the final return is actually needed, so thank you for adding it.

ESP32 build has enabled -Werror, the missing return creates a warning that a non-void function may not return a value, and that gets treated like an error due to the -Werror switch - see the CI run: https://github.com/jgromes/RadioLib/actions/runs/8451714866/job/23150617366#step:6:1237

src/modules/CC1101/CC1101.cpp
```diff
@@ -489,6 +489,24 @@ int16_t CC1101::setRxBandwidth(float rxBw) {
   return(RADIOLIB_ERR_INVALID_RX_BANDWIDTH);
 }
 
+int16_t CC1101::autoSetRxBandwidth() {
+    // Uncertainty ~ +/- 40ppm for a cheap CC1101
+    // Uncertainty * 2 for both transmitter and receiver
+    float uncertainty = (this->frequency) * 40 * 2);
+    uncertainty = (uncertainty/1000); //Since bitrate is in kBit
+    float minbw = (this->bitRate) + uncertainty);
```
jgromes, line 497: Looks like there's an extra parenthesis here which prevents this from being compiled - see the most recent CI run:  https://github.com/jgromes/RadioLib/actions/runs/8458982857/job/23174368214#step:6:1235

jgromes:
@Crsarmv7l it looks like your branch has a conflict with the master due to change in keywords - could you please merge the latest master to your branch so that I can merge this? Thanks!

src/modules/CC1101/CC1101.cpp
Crsarmv7l: Fixed

Crsarmv7l:
> @Crsarmv7l it looks like your branch has a conflict with the master due to change in keywords - could you please merge the latest master to your branch so that I can merge this? Thanks!

You guys move fast. Manually merged the change. Looks like it worked

jgromes (approved):

jgromes:
All good now - merged, thank you for the contirbution!

PR #1044, opened 2024-03-31
Update ArduinoHal.h to make spi and friends protected

jp-bennett:
Most of the "override" functions here can't actually be overridden in a useful way when spi, spiSettings, and everything else is marked private. If everything is override, then nothing should be private.

### DIFF
--- a/src/ArduinoHal.h
+++ b/src/ArduinoHal.h
@@ -60,7 +60,7 @@ class ArduinoHal : public RadioLibHal {
     uint32_t pinToInterrupt(uint32_t pin) override;
 
 #if !RADIOLIB_GODMODE
-  private:
+  protected:
 #endif
     SPIClass* spi = NULL;
     SPISettings spiSettings = RADIOLIB_DEFAULT_SPI_SETTINGS;


HeadBoffin:
What about the God Mode option?

jp-bennett:
> What about the God Mode option?

Yes, that's how we'll manage this for now, but seems like a sub-par solution. Has too many side-effects, and isn't designed for a long term solution.

jgromes:
I'm not sure I understand why this is necessary. Methods of `ArduinoHal` are not supposed to be overriden themselves, they are overriding the virtual methods from `RadioLibHal`, which is the base class of  `ArduinoHal`. Though perhaps I'm missing something  in your use-case.

HeadBoffin:
> > What about the God Mode option?
> 
> Yes, that's how we'll manage this for now, but seems like a sub-par solution. Has too many side-effects, and isn't designed for a long term solution.

I mean't will God Mode continue to be available if this is merged?

jp-bennett:
> I'm not sure I understand why this is necessary. Methods of `ArduinoHal` are not supposed to be overriden themselves, they are overriding the virtual methods from `RadioLibHal`, which is the base class of `ArduinoHal`. Though perhaps I'm missing something in your use-case.

We already use ArduinoHal as a parent class, and override spiBeginTransaction and spiEndTransaction to do some extra locking. It would be very useful to override spiTransfer to make use of the multi-byte transfer() methods available on some platforms.



> I mean't will God Mode continue to be available if this is merged?

Yes, should continue to work the exact same way. This change would just make these objects accessible for inherited classes, when god mode is turned off.

jgromes:
@jp-bennett thank you for the extra context, that makes the request quite clear. I think we can have the members protected to allow easier customization of the HAL.

PR #1047, opened 2024-04-01
Added pin maps

matthias-bs:


matthias-bs:
Those pin maps are tried and tested with https://github.com/matthias-bs/BresserWeatherSensorTTN (based on Arduino LMIC - therefore DIO1 was needed. I thought RadioLib would need it, too.

jgromes:
> based on Arduino LMIC - therefore DIO1 was needed

I don't think so, RadioLib needs just one of the DIO pins for SX127x/6x to detect packet being received. We're not using CAD to detect when the LoRaWAN gateway is sending a message, just a simple delay and receive with timeout. So I think that note should be removed and `RADIOLIB_NC` should be used instead of GPIO33.

matthias-bs:
Modified as suggested.

Issue #276, opened 2021-03-30
Problem with 'receiveDirect()' on SX1276


gef30441:
I tried to read the incoming data continuously via the receiveDirect() function on the following board: [Heltec WiFi LoRa ESP32 V2](https://heltec.org/project/wifi-lora-32/).
The board utilizes the SX1276, which is wired like this to the ESP32:
- CS: GPIO18
- DIO0: GPIO26
- DIO1: GPI35
- DIO2: GPI34

I basically used the 'SX127x FSK Modem Example' and extracted the code i figured was enough to fit my needs.
You can see my code here:
![code](https://user-images.githubusercontent.com/81627000/112988331-031e3300-9164-11eb-880d-d80dcb92eb48.PNG)

Unfortunately when flashing the software on my board, I get the following error message:
![image](https://user-images.githubusercontent.com/81627000/112988602-52fcfa00-9164-11eb-8590-f3eb7017ca9a.png)

Error code 16 translates to: [ERR_SPI_WRITE_FAILED](https://jgromes.github.io/RadioLib/group__status__codes.html#ga4e05af49324b9f065ae93f9664d51c2f).
The wiring can't be faulty (I hope at least) and I haven't found anything in the forums/RadioLib wiki that could help me.

I hope someone reading this is able to tell me what I did wrong here.

gef30441:
I forgot to mention that I already looked at the signals on DIO2 (DATA) and DIO1 (CLK), but nothing meaningful was transferred.

jgromes:
Double check the wiring, in your post it says that both DIO1 and DIO2 are connected to the same pin (GPI35), that seems wrong. Also, I suggest checking whether DIO2 is actually connected on that board, since on most LoRa/ESP32 boards, only DIO0 is always connected, and DIO1 is connected only sometimes. DIO2 is almost always unused.

Regarding your sketch, you're calling `receiveDirect` in the loop - you're basically spamming the radio with requests to set itself to Rx mode. Call `receiveDirect` once from setup, the radio will then remain in receive mode.

Finally, if neither of the above helps, please post debug mode output.

gef30441:
I double checked the wiring and realized I made a typo.
DIO2 is on GPI34 and DIO1 is on GPI35. 

Thanks for your response, I will try your suggestion asap and report if it worked or not.

gef30441:
So i just retried running the program with the following things changed:
- RadioLib_Debug active (RadioLib_Verbose inactive)
[...]

PR #1052, opened 2024-04-03
Updated board definitions for the LoRaWAN Examples

Velocet:
Since i don't trust HelTec to get their board configurations right (or up-to-date) i used [Arduino core for the ESP32 variants](https://github.com/espressif/arduino-esp32/blob/master/variants) as source.


README.md
jgromes, line 1: I'm guessing this is a mistake, but the library readme definitely needs to stay ;)

examples/LoRaWAN/LoRaWAN_ABP/configABP.h
```diff
@@ -64,32 +64,41 @@ const uint8_t subBand = 0;  // For US915, change this to 2, otherwise leave on 0
...
-#elif defined (ARDUINO_heltec_wireless_stick)
-  #pragma message ("Using Heltec Wireless Stick")
-  SX1278 radio = new Module(14, 4, 12, 16);
+  #pragma message ("Using Heltec WiFi LoRa32")
+  SX1276 radio = new Module(18, 26, 14, 33);
 
-#elif defined(ARDUINO_heltec_wifi_lora_32_V2)
+#elif defined(ARDUINO_HELTEC_WIFI_LORA_32_V2)
   #pragma message ("Using Heltec WiFi LoRa32 v2")
-  SX1278 radio = new Module(14, 4, 12, 16);
```
jgromes, line 81: This was contributed in #1047 and according to @matthias-bs it was tested. So which pin map (and module type)¨is correct then?

jgromes (changes requested):
Thank you for the contribution! I left some comments, some things are not totally clear to me.

README.md
Velocet: And what a mistake ... thought it was excluded in the PR cause of the separate branch.

examples/LoRaWAN/LoRaWAN_ABP/configABP.h
Velocet: Oh! From @matthias-bs! I know his Bresser projects which are really cool!

I just tested this with my HelTec LoRa32v2: Using pin name definitions (from arduino-esp32) and plain numbers work:
`SX1276 radio = new Module(18, 26, 14, 35);`
`SX1276 radio = new Module(SS, DIO0, RST_LoRa, DIO1);`

Transmit/Receive (FSK) works. As do the interrupts via `fifoGet`.

When looking at the [Bresser board definitions](https://github.com/matthias-bs/BresserWeatherSensorReceiver/blob/main/src/WeatherSensorCfg.h#L454) you can see that he is also using the _arduino-esp32_ names instead of numbers. I can't see any error with my PIN definitions?

When compiling (ArduinoIDE 2.x) the board flag gets set in all caps: `-DARDUINO_HELTEC_WIFI_LORA_32_V2`

Regarding module type: Most LoRa32 boards use the SX1276 cause of the wider freq range.
But when poking the version register it identifies as a SX1278. I can cross check with a _STM32L151CCU6_ board (_HelTec LoRa 151_) if you want?

PS: Thanks for this wonderful lib! Great work!

HeadBoffin:
@Velocet, as we've only just stabilised LoRaWAN and I can't see how it fits in to your home automation plans, would it be better to make RadioLib a dependency on your wrapper rather than fork the code?

Velocet:
@HeadBoffin Let me try to enlighten you! xD 

The "official" short board name for the V3 is [`WIFI_LoRa_32_V3`](https://github.com/espressif/arduino-esp32/blob/dcc307665f56d4b92a773f5ef1cde0160bbfbaa5/variants/heltec_wifi_lora_32_V3/pins_arduino.h#L6C9-L6C24) but what gets used in the end is the folder name:

```INI
-DARDUINO_HELTEC_WIFI_LORA_32_V2            // ALL CAPS
-DARDUINO_BOARD="HELTEC_WIFI_LORA_32_V2"    // ALL CAPS
-DARDUINO_VARIANT="heltec_wifi_lora_32_V2"  // FolDer namE (watch out for the V2 at the end!)
```

The mentioned `heltec_wifi_kit_32_V3` is the same board without the LoRa module. Since the pins are the same and you are not using the board defined pin names this works! But don't try to understand any of these names. I never had any failures with the official arduino-esp32 pin definitions as those get updated and reviewed more often. Maybe a hint to only use those would be good as most manufacturer implementations suck (or are wrong.. even in arduino-esp32: the SX1262 has no DIO0 but it is defined^^).

I only forked RadioLib so i can test more easily and maybe make some contributions as i get along. In the end it will just be a protocol addition.

Velocet (commented):
Cross checked again with official definitions. No errors found.

I can't remove the readme from the PR?

HeadBoffin:
> Cross checked again with official definitions.

Which ones?

StevenCellist:
> I can't remove the readme from the PR?

Well, then re-upload it!

README.md
jgromes: This is not resolved, the readme is still missing. @Velocet since you deleted it in one commit yit should be relatively straightforward to revert that commit. All commits in the PR will get squashed before merging anyway.

examples/LoRaWAN/LoRaWAN_ABP/configABP.h
jgromes: To me this is still not resolved. @matthias-bs can you comment on this? Is your set of pins correct, or is @Velocet's? It seems to me like it should be the same ahrdware, so I don't understand why there are two different sets of pins.

If there is some discrepancy then I will have to err on the side of caution and just remove the pin maps.

StevenCellist: Based on [pinmap] (https://resource.heltec.cn/download/WiFi_LoRa_32/WIFI_LoRa_32_V2.pdf) and [schematic](https://resource.heltec.cn/download/WiFi_LoRa_32/V2/WIFI_LoRa_32_V2(868-915).PDF) I'd say the SX1276 is correct. And it is an SX1276 as can be seen from the schematic. But I don't have one, so can't test for 100% certainty.

Velocet:
Will do so when i'm back home. Thanks for the hint. Seems the app doesn't let me do it.

examples/LoRaWAN/LoRaWAN_ABP/configABP.h
HeadBoffin: > If there is some discrepancy then I will have to err on the side of caution and just remove the pin maps.

I've got all the files open on the sofa laptop ready for some wine based PR's tonight, including correcting a couple of my own errors. Should I compile before pushing?

Velocet: > To me this is still not resolved. @matthias-bs can you comment on this? Is your set of pins correct, or is @Velocet's? It seems to me like it should be the same ahrdware, so I don't understand why there are two different sets of pins.

I tested this with my HelTec WiFi LoRa32 V2 and other projects use the same PINs as do other people who use RadioLib:
- (SX1276 unable to receive packets)[https://github.com/jgromes/RadioLib/discussions/371]
- [Problem with 'receiveDirect()' on SX1276](https://github.com/jgromes/RadioLib/issues/276)

The values are the same for the [HelTec](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/tree/master/variants) and [arduino-esp32](https://github.com/espressif/arduino-esp32/tree/master/variants) PIN definitions. One thing to mention: DIO1 and DIO2 on the HelTec are connected to Input only PINs which could result in errors handling interrupts.

> If there is some discrepancy then I will have to err on the side of caution and just remove the pin maps.

The PIN maps are a great value for beginners and should definitely stay. I also see no discrepancy: Use the "official" PIN definitions from arduino-esp32 and you are good to go without any headaches.

PS: only have my mobile with me. Can't do the readme commit deletion before sunday 🫠

matthias-bs: > To me this is still not resolved. @matthias-bs can you comment on this? Is your set of pins correct, or is @Velocet's? It seems to me like it should be the same ahrdware, so I don't understand why there are two different sets of pins.
> 
> If there is some discrepancy then I will have to err on the side of caution and just remove the pin maps.

I could not test any Heltec board myself - I don't have any and I won't by one anytime soon. The Heltec definitions are based on documentation and feedback by users of my project.

HeadBoffin: I have a V2, which is where the original pins came from so I'm very sure that it does work but as I selected the wifi_kit rather than wifi_lora when originally fishing out the defines. The wifi_lora has been added, I'm not sure where those pins have come from, possibly from the wireless stick.

HeadBoffin:
> Another interesting issue regarding board definitions: [espressif/arduino-esp32#9423](https://github.com/espressif/arduino-esp32/issues/9423)

Like standing on quick sand some days!

Velocet:
Finally readded the README.md. Now everything should be resolved.

jgromes (approved):

jgromes:
@Velocet looks good now, merged - thanks!

However, with the caveat that I am not sure if these configuration files and pinmaps will remain long-term. So far it seems more trouble than it's worth (it is prone to breaking if the platform provider makes a brekign change and the CI for it is a nightmare). Guess we'll see ...

matthias-bs:
I'm still wondering why you get all-capitalized defines an I don't... I double-checked my setup - it's all up-to-date.

StevenCellist:
I see all caps for both [Arduino](https://github.com/espressif/arduino-esp32/blob/49b94644c693df22f0a15ffb9454e2d7c3e6c679/boards.txt#L21658) and [PlatformIO](https://github.com/platformio/platform-espressif32/blob/c9c9bbeceef04fbe756d9540e23f47681a8d1000/boards/heltec_wifi_lora_32_V2.json#L8).

PR #1054, opened 2024-04-05
Make PA ramp up time configurable for SX1262

S5NC:
Default behaviour should not be changed.
Backwards-compatible (I would have liked to put `rampTime` next to `power` in `begin()` but this would have been breaking)
If you are okay with this, I can makes the same changes to the SX1261 and SX1268.

### DIFF
--- a/src/modules/LLCC68/LLCC68.cpp
+++ b/src/modules/LLCC68/LLCC68.cpp
@@ -6,7 +6,7 @@ LLCC68::LLCC68(Module* mod) : SX1262(mod) {
   this->XTAL = true;
 }
 
-int16_t LLCC68::begin(float freq, float bw, uint8_t sf, uint8_t cr, uint8_t syncWord, int8_t pwr, uint16_t preambleLength, float tcxoVoltage, bool useRegulatorLDO) {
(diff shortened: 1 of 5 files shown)


jgromes:
> The build fails because ...

I have this one weird trick you can try - maybe try building it yourself before pushing to the pull request? You are the one who opened it, it's up to you to make sure it doesn't break the rest of the repository.

Also, despite your original comment, you did add the ramp time to `begin` methods, which could be considered a breaking change. Why? A simple method e.g. `setPaRampTime` alongside an additional argument for `setOutputPower` would have been so much easier.

GitHub sends me an email every time you push and/or comment, and because you both comment AND commit in a stream-of-consciousness style, most of those emails are completely useless to me because by the time I have a chance to reply, everything is changed already. It's incredibly rude to use git commits or issue comments as chat.

jgromes:
> I am still getting error: 'virtual int16_t PhysicalLayer::setOutputPower(int8_t)' was hidden [-Werror=overloaded-virtual=],

It's because `PhysicalLayer` expects `setOutputPower` to have the signature `int16_t setOutputPower(int8_t)`, while you changed it to  `int16_t setOutputPower(int8_t, uint8_t)`.

But why are you opening a pull request which doesn't compile and which you cannot fix? If you have a suggestion for a new feature that you can't implement yourself, either learn how to it, or open the appropriate issue.

Closing as implementing this from the scratch myself is easier than fixing your PR.

PR #1075, opened 2024-04-24
Use RadioLibTime_t (aka unsigned long) when dealing with millis() and micros()

Elizafox:
Although `sizeof(uint32_t) == sizeof(unsigned long)` on Arduino, this is not the case on 64-bit Linux, where `sizeof(unsigned long) == sizeof(uint64_t)`.

Most timestamp arithmetic and comparisons have been left alone, to reduce code churn. This is fine, as `uint32_t` is perfectly wide to store most timestamp deltas this library will deal with, and C will promote the integer rather than do a narrowing conversion. The real problem arises with narrowing conversions being done by assuming timestamps are 32-bit.

No functional changes intended for platforms where `sizeof(uint32_t) == sizeof(unsigned long)` (so most 8/16/32-bit platforms).

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -118,7 +118,7 @@ int16_t LR11x0::reset() {
   this->mod->hal->delay(300);
   
   // wait for BUSY to go low
-  unsigned long start = this->mod->hal->millis();
+  RadioLibTime_t start = this->mod->hal->millis();
   while(this->mod->hal->digitalRead(this->mod->getGpio())) {
     this->mod->hal->yield();
     if(this->mod->hal->millis() - start >= 3000) {
@@ -143,7 +143,7 @@ int16_t LR11x0::transmit(uint8_t* data, size_t len, uint8_t addr) {
   // get currently active modem
   uint8_t modem = RADIOLIB_LR11X0_PACKET_TYPE_NONE;
   state = getPacketType(&modem);
-  uint32_t timeout = getTimeOnAir(len);
+  RadioLibTime_t timeout = getTimeOnAir(len);
   if(modem == RADIOLIB_LR11X0_PACKET_TYPE_LORA) {
     // calculate timeout (150% of expected time-on-air)
     timeout = (timeout * 3) / 2;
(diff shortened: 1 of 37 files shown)


jgromes:
Thank you for the contribution!

One question/suggestion, would it be possible to use a custom type for this, to make the intent clear? For example `typedef RadioLibTimestamp_t unsigned long` somewhere in TypeDef.h. This would also have the added benefit of making things a bit more future-proof (since it would be a lot easier to keep this in mind when writing more code/checking pull requests).

Elizafox:
That's a great idea. I thought about doing it like this in fact. I will do it today after work.

Elizafox:
> That's a great idea. I thought about doing it like this in fact. I will do it today after work.

In that case it may also be beneficial to change more uses where timestamps are involved to use this type. I already had to update some of the API, it might be beneficial to just rip the bandaid off and just change it all for clarity.

jgromes:
> In that case it may also be beneficial to change more uses where timestamps are involved to use this type

Definitely, that would be better to keep everything consistent.

Elizafox:
> > In that case it may also be beneficial to change more uses where timestamps are involved to use this type
> 
> Definitely, that would be better to keep everything consistent.

I have mostly done so. I have kept some timestamps 32-bit, mostly stuff dealing with chips and protocol stuff. This is to make it clear what values are expected.

src/modules/LR11x0/LR11x0.cpp
```diff
@@ -1282,14 +1282,13 @@ uint32_t LR11x0::getTimeOnAir(size_t len) {
...
+    return(((RadioLibTime_t(1) << this->spreadingFactor) / this->bandwidthKhz) * N_symbol * 1000.0);
 
   } else if(type == RADIOLIB_LR11X0_PACKET_TYPE_GFSK) {
-    return(((uint32_t)len * 8 * 1000000UL) / this->bitRate);
+    return(((RadioLibTime_t)len * 8 * 1000000UL) / this->bitRate);
   
   } else if(type == RADIOLIB_LR11X0_PACKET_TYPE_LR_FHSS) {
-    return(((uint32_t)len * 8 * 1000000UL) / RADIOLIB_LR11X0_LR_FHSS_BIT_RATE);
-
+    return(((RadioLibTime_t)len * 8 * 1000000UL) / RADIOLIB_LR11X0_LR_FHSS_BIT_RATE);
```
jgromes, line 1291: I don't think these casts should be to `RadioLibTime_t`, the point here is to not overflow `int` with the bitshifts and multiplications. HAL millis()/micros() are not used here, so `uint32_t` should stay internally in this method. The return type is correct, as outside it needs to be trated as time.

This is also the case for `SX127x::getTimeOnAir` and `SX128x::getTimeOnAir`.

jgromes (changes requested):
Thank you, I left one minor(/bit nitpicky) comment.

src/modules/LR11x0/LR11x0.cpp
Elizafox: Done in e35e03

jgromes (approved):

jgromes:
All looks good now - merged, thank you for the contribution!

PR #1081, opened 2024-04-29
Buffer Changes: write to a single byte buffer, then write to main buffer, prevent buffer overflow

Crsarmv7l:
Based on the discussion here: https://github.com/jgromes/RadioLib/discussions/1053#discussioncomment-9076154 this pull writes each byte to a single byte buffer and checks it before writing to the main buffer.

Please let me know what else you are thinking so we can access the buffer directly without god mode.


Crsarmv7l:
As far as I am aware I still need to pull the buffer out of God mode, add the byteBuff to God Mode, and whatever else I missed (please let me know what those things are :p)

jgromes:
Thank you for the contribution! The implementation of the `byteBuff` is pretty much what I had in mind in the discussion linked above. However, I'm now less sure about `PhysicalLayer::getDirectBuffer` as that pretty much just makes the direct mode buffer `public`.

I think the proper solution would be to internally use a circular buffer, as the current implementation has no protections against overflows. The user would then only be allowed access to the data via `PhysicalLayer::read`. If you'd be willing to implement that, it would be a significant improvement.

Crsarmv7l:
> Thank you for the contribution! The implementation of the `byteBuff` is pretty much what I had in mind in the discussion linked above. However, I'm now less sure about `PhysicalLayer::getDirectBuffer` as that pretty much just makes the direct mode buffer `public`.
> 
> I think the proper solution would be to internally use a circular buffer, as the current implementation has no protections against overflows. The user would then only be allowed access to the data via `PhysicalLayer::read`. If you'd be willing to implement that, it would be a significant improvement.

I had thought of a circular buffer back when we talked but a few things made me not suggest it:

1. A circular buffer should probably be zeroed out when you loop around, are you ok with memset to zero it out?
2. The buffer size should probably be tweaked with a circular buffer, STATIC_ARRAY_SIZE is quite large.
3. It would solve overflow, but the end user still would have to create a copy of the buffer in order to process it. 

Let me know your thoughts, and I can probably take a whack at it

jgromes:
> 1. A circular buffer should probably be zeroed out when you loop around, are you ok with memset to zero it out?

Not really, the buffer is just a storage space, we can overwrite the old data when looping around.

> 2. The buffer size should probably be tweaked with a circular buffer, STATIC_ARRAY_SIZE is quite large.

Possibly, but that seems to be just an optimization. I don't see 256 bytes as particulary large buffer ;)

> 3. It would solve overflow, but the end user still would have to create a copy of the buffer in order to process it.

In this case to copy the data on `read` is valid in my opinion. We want the user to have access to data that won't change/get overwritten, that's why the access must be done through the dedicated `read` method.

Crsarmv7l:
>> Not really, the buffer is just a storage space, we can overwrite the old data when looping around.

If I use the single byte buffer in this pull, yeah that always resets/starts from zero so we get the correct byte then can overwrite the buffer position. Seems risky to be doing bit operations on a non-0 value (the way the buffer originally was it wrote straight to the buffer) and expecting it to come out correctly

>> Possibly, but that seems to be just an optimization. I don't see 256 bytes as particulary large buffer ;)

Fair, that is not an urgent thing

>> In this case to copy the data on `read` is valid in my opinion. We want the user to have access to data that won't change/get overwritten, that's why the access must be done through the dedicated `read` method.

What I am getting at is only having read is highly inefficient for decoders, having to loop through read and only getting a single byte each time to copy to a temp buffer, that you then have to iterate through again to process vs once the buffer is full enough, iterating directly through the buffer or a copy once and decoding.

But that may be something I do locally. I am more leaning towards creating a function to return a copy of a buffer for 0 to whatever user specified position.

Crsarmv7l:
Changes:
 - Removed getDirectBuffer
 - Added check of bitBufferWritePos and set to 0 if >= RADIOLIB_STATIC_ARRAY_SIZE before writing the single byte buffer to the main buffer
 - Since Available simply returns bitBufferWritePos, this is a simple way to prevent buffer overflow

PR #1082, opened 2024-05-01
examples/NonArduino/Tock: Support RISC-V and bump libtock-c

alistair23:


jgromes:
I took a look in the CI runs and it looks like it's missing pretty basic stuff, like stdio and stdint:

https://github.com/jgromes/RadioLib/actions/runs/8907994652/job/24462852153#step:5:8662

To me that makes it look like the environment is not set up correctly. So I'm not sure this should be merged before that gets resovled. Disabling RISC-V build in CI is a rather strange choice considering that this PR adds RISC-V support.

alistair23:
What is happening is that the RadioLib cmake build environment in the CI isn't setup correctly for RISC-V, hence the missing stdio and stdint.

I can't reproduce the failures locally in Arch, as Arch has better RISC-V packages.

libtock-c also builds fine as it [builds it's own standard libraries](https://github.com/tock/libtock-c/blob/master/Configuration.mk#L242).

The problem is that I can't easily get the Make information into CMake to get it to build successfully in the CI. I can use [RadioLib as a library with libtock-c](https://github.com/tock/libtock-c/tree/master/wip/lora) (with a not yet submitted patch) to build a RISC-V application and I can build this PR successfully locally for RISC-V. The CI happily builds libtock-c for RISC-V as well. The stdio and stdint headers are provided as well as the libraries, so there is just something slightly wrong in the CMake setup that fails in the CI, but I'm a little stuck reproducing it.

jgromes:
> What is happening is that the RadioLib cmake build environment in the CI isn't setup correctly for RISC-V, hence the missing stdio and stdint.

But you have control over that environment, the `tock-build` job does nothing else than build tock. So you should be able to set it up in such a way that it can compile both ARM and RISC-V, otherwise it's a bit of a "it works on my machine" kind of situation ... ;)

alistair23:
That's a fair point. I managed to fix it!

So now RISC-V is built in the CI. The script still provides an option to skip the RISC-V RadioLib build in case others have an issue.

Also the `abs()` function isn't officially include in `math.h`: https://www.techonthenet.com/c_language/standard_library_functions/stdlib_h/abs.php

jgromes:
Logging good, thank you very much!

> Also the abs() function isn't officially include in math.h:

Thanks, RadioLib should be using `fabs` exclusively. Though there do seem to be two of them I forgot in FSK4.cpp, I will fix that right away.

Issue #729, opened 2023-04-12
Raspberry Pico + SX 1262 


Steffen54321:
Hi,

I am new to the LoRa/ microcontroller business and try to make 2 raspberry pico to communicate via LoRa or FSK. Therfore i bought a Waveshare SX1262 Shield (https://www.waveshare.com/pico-lora-sx1262-868m.htm) and use the Arduino IDE with Arduino MBED RP2040 Board chosen. 
When compiling the SX126X examples of RadioLib, I get the failed, code -2. 
I changed the Ports as shown in the waveshare link (SX1262 radio = new Module(5, 26, 20, 4);)
What am I doing wrong? Thanks alot in advance!

jgromes:
I don't have either the RP2040 or the waveshare module, but looking at the pinout diagram, I think you may have used the physical device pin numbers (1 - 40) as ooposed to the actual GPIO pin number (GPxy). Can you try `Module(3, 20, 15, 2);`?

Steffen54321:
Thanks for the fast response! But i still get: 
failed, code -2

jgromes:
In that case please fill in the "module not working" template, especially post the debug output.

Steffen54321:
**Sketch that is causing the module fail**

```c++
//SX126x_Transmit Example Code
// include the library
#include <RadioLib.h>

// SX1262 has the following connections:
// NSS pin:   10      3
// DIO1 pin:  2       20
// NRST pin:  3       15
// BUSY pin:  9       2  
SX1262 radio = new Module(3, 20, 15, 2);



// or using RadioShield
// https://github.com/jgromes/RadioShield
//SX1262 radio = RadioShield.ModuleA;

// or using CubeCell
//SX1262 radio = new Module(RADIOLIB_BUILTIN_MODULE);

void setup() {
  Serial.begin(9600);

  

  // initialize SX1262 with default settings
  Serial.print(F("[SX1262] Initializing ... "));
  int state = radio.begin();
  if (state == RADIOLIB_ERR_NONE) {
    Serial.println(F("success!"));
  } else {
    Serial.print(F("failed, code "));
    Serial.println(state);
    while (true);
  }

  // some modules have an external RF switch
  // controlled via two pins (RX enable, TX enable)
  // to enable automatic control of the switch,
  // call the following method
  // RX enable:   4
  // TX enable:   5
  /*
    radio.setRfSwitchPins(4, 5);
  */
}

void loop() {
[...]

PR #1086, opened 2024-05-04
add support for Pi Pico + sx1262 module to LoRaWAN examples

K4KDR:
Hello!   After finding success w/ the LoRaWAN examples using my Arduino UNO + sx1278 hardware, I wanted to be able to utilize the excellent **Waveshare SX1262 LoRa Node Module for Raspberry Pi Pico** (https://www.waveshare.com/pico-lora-sx1262-868m.htm) that is a direct header plug-in to the Pi Pico microcontroller.

pico+sx1262

A previous issue (https://github.com/jgromes/RadioLib/issues/729) identified the syntax required to use the sx1262 module.

After modifying my local code to successfully uplink a LoRaWAN payload using the Pi Pico / sx1262 combination, I wanted to share the code if by chance you would like to integrate this hardware option into the list of choices in the LoRaWAN examples.

Thanks!

-Scott,  K4KDR

### DIFF
--- a/examples/LoRaWAN/LoRaWAN_ABP/configABP.h
+++ b/examples/LoRaWAN/LoRaWAN_ABP/configABP.h
@@ -3,6 +3,8 @@
 
 #include <RadioLib.h>
 
+#define Pi_Pico_1262_LoRa   // set to match hardware in-use (see pin maps below!)
+
 // how often to send an uplink - consider legal & FUP constraints - see notes
 const uint32_t uplinkIntervalSeconds = 5UL * 60UL;    // minutes x seconds
 
@@ -46,6 +48,11 @@ const uint8_t subBand = 0;  // For US915, change this to 2, otherwise leave on 0
     #pragma message ("Link required on board")
     SX1276 radio = new Module(8, 3, 4, 6);
 
+// Pi_Pico_1262_LoRa
+#elif defined(Pi_Pico_1262_LoRa)
+  #pragma message ("Pi_Pico_1262_LoRa")
+     // fix from https://github.com/jgromes/RadioLib/issues/729
+  SX1262 radio = new Module(3, 20, 15, 2, SPI1, RADIOLIB_DEFAULT_SPI_SETTINGS);
 
 // LilyGo 
 #elif defined(ARDUINO_TTGO_LORA32_V1)
--- a/examples/LoRaWAN/LoRaWAN_ABP/LoRaWAN_ABP.ino
+++ b/examples/LoRaWAN/LoRaWAN_ABP/LoRaWAN_ABP.ino
@@ -37,14 +37,14 @@ void setup() {
   delay(5000);  // Give time to switch to the serial monitor
   Serial.println(F("\nSetup ... "));
   
-#if defined(Pi_Pico_1262_LoRa)
-  SPI1.setSCK(10);            // fix from https://github.com/jgromes/RadioLib/issues/729
-  SPI1.setTX(11);             // ~
-  SPI1.setRX(12);             // ~
-  pinMode(3, OUTPUT);         // ~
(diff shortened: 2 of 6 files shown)


HeadBoffin:
I've not put comments on the other files, but they translate over.

HeadBoffin:
Sorry, rude of me not to say, thanks for this, very useful consolidation of the prior issues.

examples/LoRaWAN/LoRaWAN_ABP/configABP.h
```diff
@@ -46,6 +48,11 @@ const uint8_t subBand = 0;  // For US915, change this to 2, otherwise leave on 0
     #pragma message ("Link required on board")
     SX1276 radio = new Module(8, 3, 4, 6);
 
+// Pi_Pico_1262_LoRa
+#elif defined(Pi_Pico_1262_LoRa)
+  #pragma message ("Pi_Pico_1262_LoRa")
+     // fix from https://github.com/jgromes/RadioLib/issues/729
+  SX1262 radio = new Module(3, 20, 15, 2, SPI1, RADIOLIB_DEFAULT_SPI_SETTINGS);
```
jgromes, line 55: I don't think it can be called a "fix" when you are using the library as intended - what exactly is being fixed here?

Having to specify which SPI bus is used to connect to the module is a very basic information that even has its own section on the Wiki and in the troubleshooting guide. It's not something I want to hide from the user, if they leave everything at the default value and then *insert surprised Pikachu face* when it doesn't work then that suggests they know very little about the hardware they are using.

jgromes (commented):
Thank you for the contribution Scott! However, I'm asking myself whether we need to be adding more example LoRa boards to the LoRaWAN examples ... surely there are still many more we are missing and new ones will be released in the future. By including these, we are basically building a database of pin maps. Fair enough, but why should that be in the RadioLib examples - and copied three times over?

My point is, this seems very similar to what @Velocet proposed in #1065. So maybe instead of kicking the can down the road we should move these pin maps somewhere else (e.g. a new repository in https://github.com/radiolib-org)?

K4KDR:
Great points - I learn **so much** every time you answer a question or add code to a project... THANK YOU!

One example is that I had **no idea** that these examples were doing any kind of hardware auto-detect!   When I saw all the `#if defined` statements, I just assumed that the code was expecting the user to "just know" that the correct way to utilize that was to add the appropriate `#define` statement for their hardware.   There is line just before that section that mentions `Auto select MCU <-> radio connections` , but that meant nothing to me since I had never heard of any such thing.

Once I found the correct syntax to use the Pi Pico + sx1262 combination, I just wanted to share that for your review.   But it makes complete sense to leave references to work-arounds for specific hardware to **Issues** or **Discussions** in the repository where they can easily be found by keyword searches.

Thanks for taking the time to look at what I am doing and providing YET MORE learnings.   Much appreciated!

HeadBoffin:
> Once I found the correct syntax to use the Pi Pico + sx1262 combination, I just wanted to share that for your review. But it makes complete sense to leave references to work-arounds for specific hardware to Issues or Discussions in the repository where they can easily be found by keyword searches.

There will be a Pico worked example, it's just that someone else also submitted one at the same time and we need to get the auto-pinmap thing figured. So you work will emerge as a full example, just not today!

StevenCellist:
@K4KDR while this PR is pending a pin-table or other way to get the pinmaps in a better state than keeping multiple `config.h` files synchronized, I already want to thank you for your code here to get the Pico + SX1262 hat going for https://github.com/radiolib-org/radiolib-persistence/pull/4. Thanks! :)

jgromes:
Closing as https://github.com/radiolib-org/RadioBoards/pull/2 was merged.

PR #1093, opened 2024-05-12
[LoRaWAN] Change session activation

StevenCellist:
Due to a slight bug introduced in one of my recent patches, I decided after conferring with @HeadBoffin to shift around some of the logic around session restoring / activation.

Restoring the Nonces and Session buffer now immediately restores all the contents of the buffer, instead of waiting for this to happen in begin(). Should make it easier to read and maintain. Now if begin() is called with the `force` argument set to false (default), it will check if a restored session is available for the current activitation information (keys, plan and the like) and return `SESSION_RESTORED` if this is the case - if there was none or not a valid session, it will immediately do a join and return (if succeeded) `NEW_SESSION`. Hence there is no need to set `force` to true anymore, simplifying logic and being more intuitive. It is still available for rejoining while a session is active, however.

For the examples in this repository there is not much difference (just begin()'s return code and dropping the argument to begin()), but the persistence examples should see nice improvements. 

Also, a slight bug has been fixed regarding Nonces not increasing on a failed join (buffer checksum was not recalculated and could result in restoring rejection).

Pending confirmation by @HeadBoffin (already tested to run fine, but slight misuse-testing also welcome), and ready for review.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.h
+++ b/src/protocols/LoRaWAN/LoRaWAN.h
@@ -528,6 +528,11 @@ class LoRaWANNode {
     LoRaWANNode(PhysicalLayer* phy, const LoRaWANBand_t* band, uint8_t subBand = 0);
 
     /*!
+      \brief Clear an active session, so that the device will have to rejoin the network.
+    */
+    void clearSession();
+
+    /*!
       \brief Returns the pointer to the internal buffer that holds the LW base parameters
       \returns Pointer to uint8_t array of size RADIOLIB_LW_NONCES_BUF_SIZE
     */
@@ -565,11 +570,10 @@ class LoRaWANNode {
     /*!
       \brief Join network by restoring OTAA session or performing over-the-air activation. By this procedure,
       the device will perform an exchange with the network server and set all necessary configuration. 
-      \param force Set to true to force joining even if previously joined.
       \param joinDr The datarate at which to send the join-request and any subsequent uplinks (unless ADR is enabled)
       \returns \ref status_codes
     */
-    int16_t activateOTAA(bool force = false, uint8_t initialDr = RADIOLIB_LW_DATA_RATE_UNUSED, LoRaWANJoinEvent_t *joinEvent = NULL);
+    int16_t activateOTAA(uint8_t initialDr = RADIOLIB_LW_DATA_RATE_UNUSED, LoRaWANJoinEvent_t *joinEvent = NULL);
 
     /*!
       \brief Set the device credentials and activation configuration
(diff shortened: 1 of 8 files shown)


jgromes (approved):
Thanks again @StevenCellist  - looks good from my end! Admittedly I can't really check the complete join/restore logic, but from the POV of code clarity, style, memory handling etc. all seems to be in order :)

StevenCellist:
The latest commits split the `begin()` out into `begin()` and `activate()`.
The reasoning is as follows: to restore the nonces and session, the code should already know what it should expect w.r.t. keys, mode, frequency plan and class. If you first restore and then call `begin()` with a different key, it must throw away whatever it restored and it is hard to make it clear from return codes that something like that happened.
It is much easier if we first call `begin()` so the code knows what to expect, then restore the buffers, and finally call `activate()` - this either finalizes the restored buffers, or does a fresh join.

- `begin()` is now a `void`-function as it just takes the keys, generates their checksums and sets mode and class (latter always A)
- `setBufferNonces()` returns `ERR_NONE` or `NONCES_DISCARDED` (or `CHECKSUM_MISMATCH`)
- `setBufferSession()` returns `ERR_NONE` or `SESSION_DISCARDED` (or `CHECKSUM_MISMATCH`)
- `activate()` returns `SESSION_RESTORED` or `NEW_SESSION` (or `NO_DOWNLINK` or any hardware-related error-code)

Additionally, an optional event is added to `activateOTAA()` that returns the DevNonce and JoinNonce.

Checked to run fine on deepsleeping ESP32s for both OTAA and ABP on EU868, but waiting to get verified by an independent tester (@HeadBoffin).

StevenCellist:
Additionally, the user now has access to clearing a session, after which calling `activate()` results in starting a new session. Thus, the `force` argument in the `activate()` functions is now dropped to make the flow more explicit if one wishes to start a new session while one is active. 
After calling `clearSession()`, the function `isActivated()` evaluates to `false` which is an easy indication to call `activate()`.

HeadBoffin:
Abbrviating th function nams and th rturn cods dosn't mak things straightforward. Fundamntally works but thr is somthing in th saving of th join on boot == 1 that causs a join on boot == 2. Thraftr it sms to procd normally.

StevenCellist:
@HeadBoffin I think your 'e'-key needs replacement ;)

src/protocols/LoRaWAN/LoRaWAN.h
```diff
@@ -545,47 +559,43 @@ class LoRaWANNode {
...
-      the device will perform an exchange with the network server and set all necessary configuration. 
+      \brief Set the device credentials and activation configuration
       \param joinEUI 8-byte application identifier.
       \param devEUI 8-byte device identifier.
       \param nwkKey Pointer to the network AES-128 key.
       \param appKey Pointer to the application AES-128 key.
-      \param force Set to true to force joining even if previously joined.
+    */
+    void beginOTAA(uint64_t joinEUI, uint64_t devEUI, uint8_t* nwkKey, uint8_t* appKey);
+
```
jgromes, line 569: Since `begin` methods are now `void`, and new `activate` methods were added, we need to update the examples to reflect that.

HeadBoffin: A product of the testing!

StevenCellist:
Examples rework in progress, via separate PR.

PR #1094, opened 2024-05-12
LR11x0: fix setDioAsRfSwitch argument count

bkleiner:
All 3 LR chip variants actually take 8 bytes as argument to `setDioAsRfSwitch` with different bytes being RFU depending on the supported feature-set.
Seeing that this is really an internal-ish function, i propose we expose all 8 bytes to the calling side and let that figure out which ones are actually needed.


jgromes:
Thank you for the contribution! Admittedly I don't have the RF switching on LR1110 fully worked out yet, though it is being worked on.

PR #1102, opened 2024-05-20
fix payload of DevStatusAns mac-command

pnxs:
Lorawan specs say (e.g. 1.0.4 or 1.1), that the first byte is the battery-level and the second byte is the margin.


StevenCellist:
Ah.. that explains! Thank you for fixing and contributing!

PR #1105, opened 2024-05-25
Fixed SX128X checkOutputPower

lewisxhe:
The checkOutputPower parameter name is the same as the class member variable. As a result, when checking the set power, the value in the member variable is obtained. Therefore, as long as the set power value is greater than 0, an error will be returned.


jgromes:
Thank you for the contribution! I've been cleaning up these shadowed arguments/variables but clearly I missed at least one.

PR #1108, opened 2024-05-25
[LR11x0] Make config options public, in line with SX126x

caveman99:
Meshtastic source code requires these functions to be public, same as on SX126x. We'd like to avoid using godmode :-)


jgromes:
Thanks for letting me know about this - I'm currently trying to figure out how to handle RF switching for the LR11x0 since the options for it are simultaneously rather extensive, and also completely out of control for the typical user ...

I think there will be some changes to these methods yet, most likely to `setDioAsRfSwitch`, so I propse we close this. I also suggest to wait with implementing Meshtastic support for LR11x0 until the next RadioLib release ;)

caveman99:
Fair enough :-) - we're still in a draft branch on LR11x0 support, and just building along ... I am also not sure we are using the right irq mask register setup yet.

PR #1115, opened 2024-06-04
[LoRaWAN] To change CR to 4/5 for all frequency plans

IoTThinks:
Hi team,
I'm testing RadioLib in AS923 frequency plan.
I realize RadioLib uses CR 4/7 for AS923 and other Asian frequency plans such as CN, AS923, EU433, KR and IN.

I read the LoRaWAN regional parameter.  It never mentions which CR to use for CN, AS923, EU433, KR and IN
https://resources.lora-alliance.org/technical-specifications/rp002-1-0-4-regional-parameters

image

However, the default CR in use for LoRaWAN server and gateways and devices is CR 4/5 for AS923_1/2/3/4.
This is CR 4/5 in ChirpStack LoRaWAN Server in AS923. All frequency plans use CR 4/5.
https://github.com/chirpstack/chirpstack/blob/master/lrwn/src/region/as923.rs#L45
image

The Things Stack also uses CR 4/5 for AS923.
https://github.com/TheThingsNetwork/lorawan-stack/blob/v3.31/pkg/band/as_923_rp2_v1_0_4.go#L106
image

All Dragino devices and RAK 3172 Evaluation Board use CR 4/5 for AS923.

In CR 4/7, gateways receives the packets but ChirpStack seems to drop the packets.
After changing to CR 4/5, everything works perfect.
image

I fix to all frequency plans to use CR 4/5 too.
Thanks a lot.


StevenCellist:
Hi @IoTThinks, thank you for checking out RadioLib! I was already aware of your activity due to the ❤️ on TTN forum ;)
The coding rate in LoRaWAN is a bit of an oddity - you never hear about it, gateways don't care, TTN doesn't care but indeed Chirpstack does. The regional parameters (heads up: RadioLib uses Regional Parameters 1.1, not 1.0.4!) fix the coding rate for some of the bands but not all, for which we assumed the older default 4/7. 
TTN also turns out to use 4/5 for RP 1.1 (e.g. [here](https://github.com/TheThingsNetwork/lorawan-stack/blob/3b30399d4ef9e4cb2e2a7aeb68e8c0b4d54bfa26/pkg/band/au_915_928_rp1_v1_1_rev_a.go#L42)), so this looks all good to me.
Thanks for investigating and contributing!

IoTThinks:
Yah, thanks a lot.
I will add a few codes to support AS923_2, _3 and _4 too.

Thanks a lot for your awesome LoRaWAN library.

StevenCellist:
There are no AS923_2, _3 and _4 for RP 1.1, unless you have a different source than me?

PR #1121, opened 2024-06-12
[CC1101] Fix getRSSI data source

Shensd:
Class variable `directModeEnabled` is set to true by default and is only set to false when calling `CC1101::receiveDirectAsync`. This has the following effects:

- When using asynchronous direct receive mode, -74.00 is always returned as RSSI data is not pulled from the RSSI register and class variable `rawRSSI` has never been changed from the default of 0.
- When using the non-direct receive mode, RSSI is pulled from the RSSI register rather than being from the last packet's status.

I am not an expert in how the CC1101 works, but under my current understanding data should be pulled from the RSSI register when using either of the direct modes and from the status of the previous packet when using the non-direct mode. 

Thank you for the wonderful library!

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -1023,7 +1023,7 @@ int16_t CC1101::directMode(bool sync) {
   SPIsendCommand(RADIOLIB_CC1101_CMD_IDLE);
 
   int16_t state = 0;
-  this->directModeEnabled = sync;
+  this->directModeEnabled = true;
   if(sync) {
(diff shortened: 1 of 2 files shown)


jgromes:
You're right, this is indeed not correct. However, to completely fix this, we should also have `this->directModeEnabled = false;` in `CC1101::setPacketMode`. Could you add that so that the entire fix is in this PR? Thank you!

Shensd:
Good point, I've added the requested changes.

jgromes:
All looks good now - merged, thank you very much for the contribution!

PR #1128, opened 2024-06-21
Fixed LR1121 initialization and power settings

lewisxhe:
1. Fixed the problem of LR1121 initialization failure. Because LR1121 does not have WiFi and GNSS, calling gnssReadVersion and wifiReadVersion will return failure. The solution is to determine whether it is LR1121. If so, do not read these two information.
2. Fix the problem of no output power when using the 2.4G band. When using 2.4G, you need to switch the PA to the High-frequency Power Amplifier, otherwise the output power is very weak and cannot be used normally.

It is not very good to write a setOutputPower method alone. I don't know the parameters of the other two models (LR1110, LR1120), but I used this method to quickly fix it.
Submitting this PR is just to tell you the current solution. If there is a better way, please close this PR

### DIFF
--- a/src/modules/LR11x0/LR1121.cpp
+++ b/src/modules/LR11x0/LR1121.cpp
@@ -5,4 +5,44 @@ LR1121::LR1121(Module* mod) : LR1120(mod) {
   chipType = RADIOLIB_LR11X0_DEVICE_LR1121;
 }
 
+int16_t LR1121::setOutputPower(int8_t power, bool useHighFreqPa)
+{
+
+    int16_t state = RADIOLIB_ERR_INVALID_OUTPUT_POWER;
+
+    if (useHighFreqPa) {
+        if ((-18 <= power ) && ( power <= 13 )) {
+            state = setPaConfig(0x02, //  High-frequency Power Amplifier
+                                0x00, //  Power amplifier supplied by the main regulator
+                                0x04, //  Power Amplifier duty cycle (Default 0x04)
+                                0x07  //  Number of slices for HPA (Default 0x07)
+                               );
+        } else {
+            return RADIOLIB_ERR_INVALID_OUTPUT_POWER;
+        }
+    } else {
+        if (( -17 <= power ) && (power <= 22 )) {
+            if (power == 22) {
+                state = setPaConfig(0x01, //  High-power Power Amplifier
+                                    0x01, //  Power amplifier supplied by the battery
+                                    0x04, //  Power Amplifier duty cycle (Default 0x04)
+                                    0x07  //  Number of slices for HPA (Default 0x07)
+                                   );
+            } else {
+                state = setPaConfig(0x00, //  Low-power Power Amplifier
+                                    0x00, //  Power amplifier supplied by the main regulator
+                                    0x04, //  Power Amplifier duty cycle (Default 0x04)
+                                    0x07  //  Number of slices for HPA (Default 0x07)
+                                   );
+            }
+        } else {
+            return RADIOLIB_ERR_INVALID_OUTPUT_POWER;
+        }
+    }
+    RADIOLIB_ASSERT(state);
+
+    // set output power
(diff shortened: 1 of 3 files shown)


jgromes:
Thank you for letting me know about this - I only have LR1110, so I was not able to test the 2.4 GHz band. Unfortunately I cannot merge this as-is, because it doesn't build on ESP-IDF due to the new method. However, knowing about the issue is enough, I will push fixes to this separately and the close this PR.

K4KDR:
VERY interested in this thread, as I have never been clear on how to activate the high-power PA.    My LR1110  (LR1110MB1LCKS from SemTech) does have 2.4 GHz, but my interest is in increasing my TX power on the 915 MHz radio.   

Unless there is a better way, I will look to manually test these changes in the appropriate files for the LR1110 on my system to see if I can determine (absent a 915 MHz capable watt meter) if I am in fact transmitting at a higher level than previously observed.   Thanks for the work on this!

jgromes:
@lewisxhe should be fixed in master now, so I will close this PR. Again, thanks for letting me know!

@K4KDR 

> My LR1110 (LR1110MB1LCKS from SemTech) does have 2.4 GHz

Where is that information coming from? According to the documentation, only the LR1120/21 should have the ability to transmit and receive in S-band. Also, the functionality to select a high-power PA on the LR1110 is already implemented, see the LR1110 docs.

K4KDR:
Good day!

**According to the documentation, only the LR1120/21 should have the ability to transmit and receive in S-band**

... since I was not using 2.4 GHz for my LF-FHSS testing, all I knew was that the unit came both w/ an antenna port for 2.4 GHz and even a little antenna for it.   

Screenshot from 2024-06-21 16-19-12


But from:

`https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/RQ000001Df4q/Mz_qfFd7oSbsmETxZ4BgwuQ5vIKrHp2vdfF23DzaC58`

... I now see that this is apparently RX-only for the **wifi scanning** capability of the board.   Thanks so much for calling that to my attention!!

**the functionality to select a high-power PA on the LR1110 is already implemented, see the LR1110 docs**

... thank you SO much for clarifying that.   I had seen some references to the high-power PA in the documentation,  but I'm afraid that I was not able to determine what syntax to use to bring that online.   It 'appeared' to my very amateur eyes that there were multiple settings required to do that - in addition to simply 'turning it on', I got the impression that there were related voltage/current settings or something similar.   But again, I was not understanding all the syntax & terms being used.

 By chance could you please take a moment to point me in the direction of the appropriate #DEFINE or other method to utilize the high-power capability of the LR1110?  So sorry that I was not able to figure it out from exhaustively scanning the source files & other documentation.

K4KDR:
Follow-up:

As I continue to try to learn the proper way to enable the high-power PA, I find that the following does NOT produce any errors  (so, it gives me the impression that it is working):

`state = radio.setOutputPower(22.0, true);     // LR1110 force enable high-power PA`

... however, I've not found a command to VIEW the currently configured output power setting nor whether the high-power PA is enabled.    By chance are either of those available?   (viewing over-the-air transmits w/ various power settings is very subjective... a little difficult to tell if I am REALLY changing the output power or not - but I'll keep experimenting!)

jgromes:
@K4KDR the high-power PA will get selected autoamtically if the low-power PA is insufficient to provide the requested output level, so in that specific example, you don't have to force the high-power option.

Unfortunately there is no way to retrieve the currently configured PA from the chip, at least none officially described. Most likely it is somewhere in the SPI registers, but Semtech does not provide the full map for LR11xx (and other modules like SX126x).

K4KDR:
For reference, here are two over-the-air TX tests w/ nothing changed other than the LR1110 TX power  (definitely a difference, but it would be nice to KNOW what my TX power really is if the module makes that info available)

-15_vs_22dbm

lewisxhe:
@jgromes  OK! I tested this modification and it works fine. Thank you

2.4G
image

868M
image

Issue #1143, opened 2024-06-30
[LR11x0] IRQ is always applied to DIO9. DIO11 is not supported as an event signal yet.  


lyusupov:
**Describe the bug**

Some new hardware design _(due to h/w compatibility reasons)_ uses DIO11 signal to deliver the LR11x0 radio interrupt to the MCU. This is allowed by LR1121 'User Manual'.

<img width="808" alt="image" src="https://github.com/jgromes/RadioLib/assets/5849637/c4af1a21-69aa-43be-9fc4-765d887e8b9a">

<br>
<br>

However, `irq2` argument for the RadioLib `setDioIrqParams()` function is always zero.

<br>
<br>

<img width="898" alt="image" src="https://github.com/jgromes/RadioLib/assets/5849637/4c3ef090-5ea4-4690-8229-12dfb5d0c44f">

<br>
<br>

**Expected behavior**

When a hardware design uses DIO11 signal to deliver the LR11x0 radio interrupt to the MCU - we can either:
1) apply a copy of `irq1` in every setDioIrqParams(irq1, irq2) call to `irq2` argument, or
2) use an additional `IrqSwap` argument for LR11x0 **begin()** method. When `IrqSwap` is `true` - apply the `irq mask` to the second argument of setDioIrqParams() rather than the first one.

I will be happy to use your own method of the issue resolution other than two my ones listed above.

**Additional info :**

 - MCU: ESP32
 - Library version 6.6.0

HeadBoffin:
> Some new hardware design _(due to h/w compatibility reasons)_

Can you clarify if this is your hardware design or some commodity item generally for sale? If the latter, link please.

jgromes:
@lyusupov I agree with @HeadBoffin that it would be good to know if we are aiming for a specific hardware (e.g. one that allows to use both DIO9 and DIO11, that would require some thinking). To add some context, I have used SeeedStudi WM1110 as the development platform, hence no DIO11 usage. 

However, at the moment, I don't really see a good use-case for separate functions on DIO pins, so the first proposed solution (to apply the same IRQ configuration to both DIO pins) is OK from my point of view, at least for now. So if you could send that as a PR, that would be great!

lyusupov:
The hardware is an RF module under development.
It is pin compatible with RFM95.
Both DIO9 and DIO11 signals  are exposed on it's pads.

**New PCB designs**  with this module on-board can use either DIO9 or DIO11 for the purpose of IRQ signal.
[...]

PR #1144, opened 2024-07-01
[LR11x0] use the same irq mask for both DIO9 and DIO11

lyusupov:
This is a proposed fix for #1143


jgromes:
Looks good - merged, thank you for the contribution!

PR #1147, opened 2024-07-04
NonArduino: Tock: Improve the reliability of delays

alistair23:
Using the Tock libtocksync_alarm_delay_ms() syscall has too much overhead, both in terms of extra time but also jitter. This means it's not a reliable source for short but accurate delays, such as the kind used in LoRaWAN.

This patch instead uses a busy loop for short (less then 5 second) delays. This might have some impact on performance and power as we are busy running in a loop, but overall we end up with a much more accurate time and working LoRaWAN.

### DIFF
--- a/examples/NonArduino/Tock/libtockHal.h
+++ b/examples/NonArduino/Tock/libtockHal.h
@@ -59,6 +59,11 @@ gpioIrqFn gpio_funcs[4] = { NULL, NULL, NULL, NULL};
 uint32_t frequency = 0;
 
 /*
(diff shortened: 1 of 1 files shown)


jgromes:
It seems like this is something to fix in Tock's syscall - I imagine a lot of use cases will require short delays. On a side note, "less than 5 seconds" is not really a short delay when talking about ms resolution.

alistair23:
Yeah, good point about the less then 5 seconds. I think it's actually a clock bug in the chip that I managed to work around in the Tock kernel :)

PR #1149, opened 2024-07-05
NonArduino: Tock: Handle zero delays

alistair23:
Calling `libtocksync_alarm_delay_ms()` on a zero ms/us delay is a large overhead, so let's just return if the `RADIOLIB_CLOCK_DRIFT_MS` scaling results in a 0ms delay

### DIFF
--- a/examples/NonArduino/Tock/libtockHal.h
+++ b/examples/NonArduino/Tock/libtockHal.h
@@ -157,19 +157,23 @@ class TockHal : public RadioLibHal {
     }
 
     void delay(unsigned long ms) override {
+      uint32_t delay;
+
 #if !defined(RADIOLIB_CLOCK_DRIFT_MS)
-      libtocksync_alarm_delay_ms(ms);
+      delay = ms;
 #else
-      libtocksync_alarm_delay_ms(ms * 1000 / (1000 + RADIOLIB_CLOCK_DRIFT_MS));
+      delay = ms * 1000 / (1000 + RADIOLIB_CLOCK_DRIFT_MS);
 #endif
+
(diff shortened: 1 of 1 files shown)


jgromes:
As far as I can tell, there should be no case in which RadioLib calls `delay(0)`. If it does, it is a mistake and likely `yield` should be called instead (to allow the possibility of task control switch in multi-threaded environments).

It could be argued that if the delay argument is low enough, it could get truncated to 0 during integer division with sufficiently large `RADIOLIB_CLOCK_DRIFT_MS`. However, in all cases where timing of intervals shorter than about 100ms matters, RadioLib calls `millis()` or even `micros()` instead of the blocking delay. In fact, `delay` is usually only used in cases where precise timing does not matter, exactly because of this uncertainty that is also present on other platforms.

Hence, I would propose to close this.

alistair23:
I see at least a `delay(1)` when running the LoRaWAN code. That gets truncated to 0 with a negative `RADIOLIB_CLOCK_DRIFT_MS`.

The delay of `0` causes some issues with the libtock-c calculation and it's a bit strange so I think it makes sense to just return if someone asks for a delay that is so small

jgromes:
> The delay of 0 causes some issues with the libtock-c calculation and it's a bit strange

I don't think it's that strange - for example, in FreeRTOS kernel, `vTaskDelay(0)` is a perfectly valid call and is actually handled, see here: https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/29b202a41e0e2810272b5765805d9fa4af247736/tasks.c#L2441

So again, I would argue this should be handled in libtock-c.

PR #1151, opened 2024-07-07
Added class member cc1101::setBitRateTolerance with new error code -109 in typeDef.h

OlliLausS:
Changes in CC1101.cpp, CC1101.h and TypeDef.h

Function added to set Bitrate Tolerance in BSCFG register of CC1101
Added class member CC1101::setBitRateTolerance as proposed from jgromes after discussion.

Added new Error code "RADIOLIB_ERR_INVALID_BIT_RATE_TOLERANCE_VALUE" with value -109 to TypeDef.h 

Tested with ESP32 and PlatformIO Environment. 
Could not test on other platforms.

### DIFF
--- a/src/TypeDef.h
+++ b/src/TypeDef.h
@@ -292,6 +292,11 @@
 */
 #define RADIOLIB_ERR_INVALID_OOK_RSSI_PEAK_TYPE                (-108)
 
+/*!
+  \brief Supplied Bitrate tolerance value is out of Range.
+*/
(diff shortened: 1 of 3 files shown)


jgromes:
Thank you for the contribution! It looks like cppcheck is reporting one minor issue, which is with the `RADIOLIB_CHECK_RANGE` macro being used. Since the value of `brt` can never be negative, you can replace the macro with just a simple `if(brt >= 0x03)`, that should get rid of that problem.

PR #1158, opened 2024-07-10
protocol: LoRaWAN: Allow configuring scanGuard

alistair23:
The current code uses scanGuard to increase the Rx window. Sometimes the default 10ms scanGuard isn't enough. So allow uses of LoRaWAN to set a larger scanGuard.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -1155,14 +1155,10 @@ int16_t LoRaWANNode::uplink(uint8_t* data, size_t len, uint8_t fPort, bool isCon
 int16_t LoRaWANNode::downlinkCommon() {
   Module* mod = this->phyLayer->getMod();
 
-  // according to the spec, the Rx window must be at least enough time to effectively detect a preamble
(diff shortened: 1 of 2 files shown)


HeadBoffin:
Does it come with a honking great warning about the consequential impact on the battery life of the device when someone over enthusiastically sets it to something silly like 1000?

jgromes:
@alistair23 Thank you for the contribution, I think it's a valid idea. Though as @HeadBoffin suggests, it would be good to have a warning in the Doxygen description, to make it clear that this is for advanced users only.

It is important to realize that by exposing this, we essentially allow the users to break the LoRaWAN Rx delay if they set some unexpectedly large value - for example, larger than the Rx delay itself, I don't know how will it behave at that point.

StevenCellist:
The value must be restricted to the interval [0,500). 500 means that the Rx2 window would open at the very moment the Rx1 window closes, higher than that means negative delay (which is handled but means not complying to LW). So either a very very big warning or a setter that returns ERR_UNKNOWN if out of range to urge users to visit an exorcist... ;)
(Although I'm not really for a setter since it means more functions.)

alistair23:
Fixed!

jgromes:
@alistair23 thank you for the contribution, merged!

PR #1167, opened 2024-07-18
Update SX1278.cpp Turn on LnaBoost even in AutoAGC mode

kb9nate:
LnaBoost can still be enabled when the gain is not manually set. The Receiver Sensitivity ratings in the SX1276/7/8 datasheet are specified with the LnaBoost enabled.

I tested this on a few boards here and on average get a 3dB increase in signal strength. This is comparable to the SX1262 "Boosted Gain" setting


jgromes:
3dB is a great improvement! Looking good, merged - thank you for the contribution!

Issue #1183, opened 2024-08-09
SX127x in continous mode has packet<>length mismatch


mycosd:
Hi,

I'm using the SX1278 (but the whole series has this issue) in LoRa mode. When using the continuous operation mode and an interrupt to then pull the received data, there are cases where you can receive multiple packets before actually reading the last one. 

The `getPacketLength` returns the length of the last packet, but the read pointer in the FIFO is still at the previous packet. So when you read the packet either using `getPacketLength` or 0 as length for `readData`, you are actually reading the previous packet with the wrong length and all packets after that get scrambled as well.

Here is a very exaggerated example

TX:
```
#include <Arduino.h>
#include <RadioLib.h>

#define LORA_PIN_NSS    5
#define LORA_PIN_DIO0   4
#define LORA_PIN_RESET  17
#define LORA_PIN_DIO1   16

SX1278 radio = new Module(LORA_PIN_NSS, LORA_PIN_DIO0, LORA_PIN_RESET, LORA_PIN_DIO1);
uint8_t count = 0;

void setup() {
    Serial.begin(115200);
    radio.begin();
}
void loop() {
    String s(++count);
    if (count & 1)s += "x";
    radio.transmit(s);
    Serial.println(s);
    delay(1000);
}
```
RX:
```
#include <Arduino.h>
#include <RadioLib.h>

#define LORA_PIN_NSS    5
#define LORA_PIN_DIO0   4
#define LORA_PIN_RESET  17
#define LORA_PIN_DIO1   16

SX1278 radio = new Module(LORA_PIN_NSS, LORA_PIN_DIO0, LORA_PIN_RESET, LORA_PIN_DIO1);
volatile bool receivedFlag = false;

#if defined(ESP8266) || defined(ESP32)
ICACHE_RAM_ATTR
#endif
void setFlag(void) {
    receivedFlag = true;
}
void setup() {
    Serial.begin(115200);
    radio.begin();
    radio.setPacketReceivedAction(setFlag);
    radio.startReceive();
}
void loop()
{
    if (receivedFlag)
    {
        receivedFlag = false;
        String s;
        if (radio.readData(s) == RADIOLIB_ERR_NONE)
            Serial.println(s);
    }
    delay(2000);
}
```

I think in continous mode, the library should follow the datasheet and advance the FIFO read pointer to the current packet. That way the packet length actually matches the data:
[...]

PR #1184, opened 2024-08-09
SX127x: Set FIFO read pointer to the start of the current packet

mycosd:
This addresses #1183. It moves the FIFO read pointer to the current packet so it matches the packet length returned from `getPacketLength` in case there were multiple packets of different lengths received between reads.


jgromes:
Thank you for the contribution! All looks good, so I merged this. 

I also wanted to point out that this only works in LoRa mode, since the registers that are being manipulated are only available in LoRa.

PR #1185, opened 2024-08-10
[SX126x/SX128x] Start reading from Rx buffer offset

GUVWAF:
As discussed in #1159 and similar to #1184, this will let SX126x and SX128x radios read the data in its buffer from the offset as returned by _GetRxBufferStatus_ in order to match the data with the returned packet length of the last packet in case you skipped reading a packet. The behaviour is the same for both radios and this is the relevant part of the datasheet:

> Two possibilities exist to obtain the offset value:
> •First is to use the RxBaseAddr value since the user defines it before receiving a payload.
> •Second, offset can be initialized with the value of RxStartBufferPointer returned by GetRxbufferStatus(...) command.
> Note:
> (...) When receiving, if the packet size exceeds the buffer memory allocated for the Rx, it will overwrite the
> transmit portion of the data buffer.

So we're now changing to the second method, with the downside that there's a higher chance you will overwrite Rx data when transmitting. However, I would argue that it's easier to make sure you don't transmit before reading available data than to make sure you always read the data in time.

For the SX126x, I had to remove the call to `setBufferBaseAddress()`, since it will force the data to start at 0, while the returned pointer is still from the previous packet (if you don't go to standby in between). I tested that this works with and without going to standby in between.

For the SX128x, there was no call to `setBufferBaseAddress()`, but it calls `standBy()` in `readData()` and therefore the returned pointer will be correct. I'm not sure if it's really necessary anymore to call `standBy()` here (and thus the interrupt example wouldn't need to call `startReceive()` again), but since I don't have the hardware, I haven't tested this.

### DIFF
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -1391,7 +1391,6 @@ float SX126x::getFrequencyError() {
 }
 
 size_t SX126x::getPacketLength(bool update) {
-  (void)update;
   return(this->getPacketLength(update, NULL));
 }
 
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -709,20 +709,17 @@ int16_t SX126x::readData(uint8_t* data, size_t len) {
   if((irq & RADIOLIB_SX126X_IRQ_CRC_ERR) || (irq & RADIOLIB_SX126X_IRQ_HEADER_ERR)) {
     crcState = RADIOLIB_ERR_CRC_MISMATCH;
   }
-
-  // get packet length
-  size_t length = getPacketLength();
+  
+  // get packet length and Rx buffer offset
+  uint8_t offset = 0;
+  size_t length = getPacketLength(true, &offset);
   if((len != 0) && (len < length)) {
     // user requested less data than we got, only return what was requested
     length = len;
   }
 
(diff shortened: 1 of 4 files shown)


src/modules/SX126x/SX126x.cpp
```diff
@@ -1395,6 +1392,11 @@ float SX126x::getFrequencyError() {
 
 size_t SX126x::getPacketLength(bool update) {
   (void)update;
```
jgromes, line 1394: This is no longer necessarry, the only reason this line was here originally was to prevent an unused argumetn warning.

src/modules/SX128x/SX128x.cpp
```diff
@@ -1254,6 +1255,11 @@ float SX128x::getFrequencyError() {
 
 size_t SX128x::getPacketLength(bool update) {
   (void)update;
```
jgromes, line 1257: SAme comment as for `SX126x::getPacketLength`

jgromes (changes requested):
Thank you for this fix! I left two minor comments, after those are addressed this can be merged.

src/modules/SX126x/SX126x.cpp
GUVWAF: Removed in last commit.

jgromes (approved):

PR #1188, opened 2024-08-15
[PHY, modules, LoRaWAN] Generalize IRQ checks

StevenCellist:
Instead of using a specific `isRxTimeout()` function, this PR adds a general function `isIrqSet()` for most common IRQ values on a range of modules (SX126x, SX127x, SX128x and LR11x0). This makes it easy to retrieve low-level IRQ status flags while retaining the radio abstraction. I am not sure if / how to implement this for other classes; they seem to work in a different way.

As the LoRaWAN stack used the function `isRxTimeout()`, it has been modified accordingly and is checked to run as usual.

### DIFF
--- a/src/protocols/PhysicalLayer/PhysicalLayer.h
+++ b/src/protocols/PhysicalLayer/PhysicalLayer.h
@@ -4,6 +4,20 @@
 #include "../../TypeDef.h"
 #include "../../Module.h"
 
+// common IRQ flags
+#define RADIOLIB_IRQ_TX_DONE                                    0x00
+#define RADIOLIB_IRQ_RX_DONE                                    0x01
+#define RADIOLIB_IRQ_PREAMBLE_DETECTED                          0x02
+#define RADIOLIB_IRQ_SYNC_WORD_VALID                            0x03
+#define RADIOLIB_IRQ_SYNC_WORD_ERROR                            0x04
+#define RADIOLIB_IRQ_HEADER_VALID                               0x05
+#define RADIOLIB_IRQ_HEADER_ERR                                 0x06
+#define RADIOLIB_IRQ_CRC_ERR                                    0x07
+#define RADIOLIB_IRQ_CAD_DONE                                   0x08
+#define RADIOLIB_IRQ_CAD_DETECTED                               0x09
+#define RADIOLIB_IRQ_TIMEOUT                                    0x0A
+#define RADIOLIB_IRQ_LR_FHSS_HOP                                0x0B
+
 /*!
   \struct LoRaRate_t
   \brief Data rate structure interpretation in case LoRa is used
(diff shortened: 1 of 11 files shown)


src/protocols/PhysicalLayer/PhysicalLayer.h
```diff
@@ -4,6 +4,20 @@
...
+#define RADIOLIB_IRQ_SYNC_WORD_VALID                            0x03
+#define RADIOLIB_IRQ_SYNC_WORD_ERROR                            0x04
+#define RADIOLIB_IRQ_HEADER_VALID                               0x05
+#define RADIOLIB_IRQ_HEADER_ERR                                 0x06
+#define RADIOLIB_IRQ_CRC_ERR                                    0x07
+#define RADIOLIB_IRQ_CAD_DONE                                   0x08
+#define RADIOLIB_IRQ_CAD_DETECTED                               0x09
+#define RADIOLIB_IRQ_TIMEOUT                                    0x0A
+#define RADIOLIB_IRQ_LR_FHSS_HOP                                0x0B
+
```
jgromes, line 20: I would propose to remove LR_FHSS_HOP IRQ since it is very specific to LoRa modules. The other ones can be applied in some way even to FSK modules.

src/protocols/PhysicalLayer/PhysicalLayer.h
```diff
@@ -346,28 +360,29 @@ class PhysicalLayer {
...
     virtual int16_t irqRxDoneRxTimeout(uint32_t &irqFlags, uint32_t &irqMask);
 
     /*!
-      \brief Check whether the IRQ bit for RxTimeout is set
-      \returns Whether RxTimeout IRQ is set
+      \brief Check whether a specific IRQ bit is set (e.g. RxTimeout, CadDone).
+      \returns Whether requested IRQ is set.
     */
-    virtual bool isRxTimeout();
+    virtual bool isIrqSet(uint8_t irq);
```
jgromes, line 382: I'm not sure it is sufficient to return boolean. There are three different outcomes for any specific IRQ check:

1. The IRQ is set - returns true
2. The IRQ is not set - returns false
3. The IRQ is not supported by the radio driver - returns false

Right now, it is not possible to distinguish whether the radio IRQ is not set, or unsupported.

The simplest solution I can think of is to return `int16_t` value of 0 if IRQ is not set, 1 if it is set and `RADIOLIB_ERR_UNSUPPORTED` if it does not exist.

Alternatively we would need to document which IRQs are supported for each of the radio drivers, which seems a bit tedious.

jgromes (changes requested):
Thank you - I left some comments, but nothing major.

PS: I also pushed an update to master to fix ESP-IDF CI and merged it here.

src/protocols/PhysicalLayer/PhysicalLayer.h
StevenCellist: In that spirit, what about SyncWordError? That one is only implemented by SX128x but given that it is in the same spirit as the other IRQs I thought I'd add it.

src/protocols/PhysicalLayer/PhysicalLayer.h
StevenCellist: Slightly inconvenient that negative values also turn into `true` when turned into a `bool`, but I understand and am OK with the solution.

jgromes: Then it might be a good idea to change the name of the method, e.g. `irqCheck`, and add macros for `RADIOLIB_IRQ_SET (1)` and `RADIOLIB_IRQ_UNSET (0)`, just to slightly discourage something like `if(isIrqSet(irq))`?

StevenCellist: Sorry, I don't really get what you mean?

jgromes: I meant that the name of the method `PhysicalLayer::isIrqSet()` could lead to some confusion - if you expect it to return boolean based on that name, you could do something like `if(radio.isIrqSet(RADIOLIB_IRQ_LR_FHSS_HOP))` and always have it evaluate as `true` if `radio` does not implement LR-FHSS. Personally, when I see a function named `ìsSomethingHappenning()` I do expect it to only return true or false, so that is a good point.

The name of the method could be changed to `PhysicalLayer::irqCheck()`, to not include that implicit "true/false" in the name itself.

src/protocols/PhysicalLayer/PhysicalLayer.h
jgromes: Agreed, let's remove it - it's not really that useful outside very specific cases (where you most likely wouldn't be using PHY anyway).

jgromes (approved):

jgromes:
All looks good now - merged, thanks!

PR #1190, opened 2024-08-18
Channel scan configuration

jgromes:
Added channel scan configuration via PHY. Currently implemented for SX126x, SX128x and LR11x0.

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -295,7 +295,8 @@ int16_t LR11x0::scanChannel() {
       .detMin = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
       .exitMode = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
       .timeout = 0,
-      .irqFlags = RADIOLIB_IRQ_NOT_SUPPORTED,
+      .irqFlags = RADIOLIB_IRQ_CAD_DEFAULT_FLAGS,
+      .irqMask = RADIOLIB_IRQ_CAD_DEFAULT_MASK,
     },
   };
   return(this->scanChannel(config));
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -480,10 +480,9 @@ int16_t LR11x0::startReceive(uint32_t timeout, uint32_t irqFlags, uint32_t irqMa
   // set DIO mapping
   uint32_t irq = irqFlags;
   if(timeout != RADIOLIB_LR11X0_RX_TIMEOUT_INF) {
-    irq |= RADIOLIB_LR11X0_IRQ_TIMEOUT;
+    irq |= (1UL << RADIOLIB_IRQ_TIMEOUT);
   }
-
-  state = setDioIrqParams(irq);
+  state = setDioIrqParams(getIrqMapped(irq));
   RADIOLIB_ASSERT(state);
 
   // clear interrupt flags
@@ -1351,12 +1350,6 @@ RadioLibTime_t LR11x0::calculateRxTimeout(RadioLibTime_t timeoutUs) {
   // the calling function should provide some extra width, as this number of units is truncated to integer
   RadioLibTime_t timeout = timeoutUs / 30.52;
   return(timeout);
-}
-
-int16_t LR11x0::irqRxDoneRxTimeout(uint32_t &irqFlags, uint32_t &irqMask) {
-  irqFlags = RADIOLIB_LR11X0_IRQ_RX_DONE | RADIOLIB_LR11X0_IRQ_TIMEOUT;  // flags that can appear in the IRQ register
-  irqMask  = irqFlags; // on LR11x0, these are the same
-  return(RADIOLIB_ERR_NONE);
 }
 
 uint32_t LR11x0::getIrqFlags() {
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -574,7 +574,7 @@ int16_t LR11x0::startChannelScan() {
   return(this->startChannelScan(config));
 }
 
-int16_t LR11x0::startChannelScan(ChannelScanConfig_t config) {
+int16_t LR11x0::startChannelScan(const ChannelScanConfig_t &config) {
   // check active modem
   int16_t state = RADIOLIB_ERR_NONE;
   uint8_t modem = RADIOLIB_LR11X0_PACKET_TYPE_NONE;
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -568,7 +568,8 @@ int16_t LR11x0::startChannelScan() {
       .detMin = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
       .exitMode = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
       .timeout = 0,
-      .irqFlags = RADIOLIB_IRQ_NOT_SUPPORTED,
+      .irqFlags = RADIOLIB_IRQ_CAD_DEFAULT_FLAGS,
+      .irqMask = RADIOLIB_IRQ_CAD_DEFAULT_MASK,
     },
   };
   return(this->startChannelScan(config));
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -278,12 +278,21 @@ int16_t LR11x0::receiveDirect() {
 }
 
 int16_t LR11x0::scanChannel() {
-  return(this->scanChannel(RADIOLIB_LR11X0_CAD_PARAM_DEFAULT, RADIOLIB_LR11X0_CAD_PARAM_DEFAULT, RADIOLIB_LR11X0_CAD_PARAM_DEFAULT));
+  ChannelScanConfig_t config = {
+    .cad = {
+      .symNum = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
+      .detPeak = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
+      .detMin = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
+      .exitMode = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
+      .timeout = 0,
+    },
+  };
+  return(this->scanChannel(config));
 }
 
-int16_t LR11x0::scanChannel(uint8_t symbolNum, uint8_t detPeak, uint8_t detMin) {
+int16_t LR11x0::scanChannel(ChannelScanConfig_t config) {
   // set mode to CAD
-  int state = startChannelScan(symbolNum, detPeak, detMin);
+  int state = startChannelScan(config);
   RADIOLIB_ASSERT(state);
 
   // wait for channel activity detected or timeout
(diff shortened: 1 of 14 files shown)


src/protocols/PhysicalLayer/PhysicalLayer.h
```diff
@@ -55,6 +55,48 @@ union DataRate_t {
...
+  /*! \brief Number of peak detection symbols */
+  uint8_t detPeak;
+  
+  /*! \brief Number of minimum detection symbols */
+  uint8_t detMin;
+  
+  /*! \brief Exit mode after signal detection is complete - module-specific value */
+  uint8_t exitMode;
+  
+  /*! \brief Timeout in milliseconds */
```
StevenCellist, line 75: I would suggest to put this value in microseconds to match the `getTimeOnAir()` function. That would make it easy to plug the expected timeout straight into the Rx timeout value :)

src/modules/SX126x/SX126x.cpp
```diff
@@ -757,7 +775,7 @@ int16_t SX126x::startChannelScan(uint8_t symbolNum, uint8_t detPeak, uint8_t det
   RADIOLIB_ASSERT(state);
 
   // set mode to CAD
-  state = setCad(symbolNum, detPeak, detMin);
+  state = setCad(config.cad.symNum, config.cad.detPeak, config.cad.detMin, config.cad.exitMode, config.cad.timeout);
```
StevenCellist, line 778: Not sure if this comment ends up on the right line, but if the exit mode is CadToRx, then we would also need to configure the IRQs for Rx. For my usecase, I use PreambleDetected, may end up using HeaderOk, and of course Timeout and RxDone (not sure if all of these are covered by the combined Rx flags you have as defaults on some modules).

StevenCellist (changes requested):
@jgromes thank you very much!! I think two points could use a slight change, but overall much appreciated and very useful :)

src/protocols/PhysicalLayer/PhysicalLayer.h
jgromes: Good point, chagned that.

src/modules/SX126x/SX126x.cpp
jgromes: Hmm, I'm thinking that the most clean way to do that would be to add an `irqFlags` field to `CADScanConfig_t`, and set the IRQ falgs based on that. What do you think?

StevenCellist: That works iff there is a translation from generalized flags to module-specific flags, just as there now is in the opposite direction for `checkIrq()`, otherwise it would not be portable to another module without changes.

StevenCellist: Side note that such a translation function would be an equal generalization of `irqRxDoneRxTimeout()` currently in use by the LW stack.

jgromes: The only other solution I can think of is to set the flags in the radio-specific implementation method, e.g. in `SX126x::startChannelScan` or `LR11x0::startChannelScan`, but that feels rather hacky.

> That works iff there is a translation from generalized flags to module-specific flags, just as there now is in the opposite direction for checkIrq()

Actually, that makes sense even as a standalone change, so I will do it like that.

StevenCellist:
@jgromes thanks for the changes. I will test the CAD specifics later, and I am currently looking at `PhysicalLayer::startReceive(uint32_t timeout, uint32_t irqFlags, uint32_t irqMask, size_t len)`. The `irqFlags` and `irqMask` methods now seem redundant, as these can now be set through `setIrq()` in a module-independent way. Do you agree on this?

StevenCellist (changes requested):
Spent the past hour investigating the new things in this PR. The problem I encountered, is that it is not possible to configure multiple flags through the `RadioIrqFlags_t` type as it is now, because the IRQ flags are not unique bits, but rather indices. This will require some rethinking, as using bits then again doesn't work as a straight index into `irqMap`... hmm.

StevenCellist:
Part of the problem is that most methods (re)configure the IRQs themselves. The new functions to set IRQs are all good, but overruled as soon as a method is called.
Sorry, just a bunch of thoughts here, not much structure.

jgromes:
>  it is not possible to configure multiple flags through the RadioIrqFlags_t type as it is now

That should be fairly straight-forward, we can have `PhysicalLayer::setIrq(uint32_t)` which takes multiple bits, then iterates over them and calls the lower level for every enabled bit, passing its index.

> most methods (re)configure the IRQs themselves

That is indeed a bit more tricky. I think we can leave it as it is now and revisit later, since CAD configures the flags on the module level, not on the PHY level. But at least these new abstractions allow us to get rid of the `PhysicalLayer::irqRxDoneRxTimeout` method.

StevenCellist:
New IRQs passed the test through LoRaWAN's `downlinkCommon()` after one compilation error:
```
lib/RadioLib/src/modules/SX127x/SX127x.cpp: In member function 'virtual int16_t SX127x::startReceive()':
lib/RadioLib/src/modules/SX127x/SX127x.cpp:389:121: error: call of overloaded 'startReceive(int, int, long unsigned int, long unsigned int)' is ambiguous
   return(this->startReceive(0, RADIOLIB_SX127X_RXCONTINUOUS, RADIOLIB_IRQ_RX_DEFAULT_FLAGS, RADIOLIB_IRQ_RX_DEFAULT_MASK));
                                                                                                                         ^
In file included from lib/RadioLib/src/modules/SX127x/SX127x.cpp:1:
lib/RadioLib/src/modules/SX127x/SX127x.h:822:13: note: candidate: 'int16_t SX127x::startReceive(uint8_t, uint8_t, RadioLibIrqFlags_t, RadioLibIrqFlags_t)'
     int16_t startReceive(uint8_t len, uint8_t mode = RADIOLIB_SX127X_RXCONTINUOUS, RadioLibIrqFlags_t irqFlags = RADIOLIB_IRQ_RX_DEFAULT_FLAGS, RadioLibIrqFlags_t irqMask = RADIOLIB_IRQ_RX_DEFAULT_MASK);
             ^~~~~~~~~~~~
lib/RadioLib/src/modules/SX127x/SX127x.h:835:13: note: candidate: 'virtual int16_t SX127x::startReceive(uint32_t, RadioLibIrqFlags_t, RadioLibIrqFlags_t, size_t)'
     int16_t startReceive(uint32_t timeout, RadioLibIrqFlags_t irqFlags, RadioLibIrqFlags_t irqMask, size_t len) override;
             ^~~~~~~~~~~~
*** [.pio\build\esp32-heltec-v3\lib5ef\RadioLib\modules\SX127x\SX127x.cpp.o] Error 1
```

Specifically casting the `mode` argument to `uint8_t` resolved the problem in this case.

StevenCellist:
And the actual CAD is working all good now, too :)
Here's a snippet I use right now which is still partially module-dependent (the `0xFF` values translate to `CAD_PARAM_DEFAULT`, `0x01` is for `CadToRx` for SX126x specifically).

```cpp
  ChannelScanConfig_t cadCfg;
  cadCfg.cad.symNum = 0xFF;
  cadCfg.cad.detPeak = 0xFF;
  cadCfg.cad.detMin = 0xFF;
  cadCfg.cad.exitMode = 0x01;
  cadCfg.cad.irqFlags  = 0;
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_CAD_DONE);
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_CAD_DETECTED);
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_PREAMBLE_DETECTED);
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_HEADER_VALID);
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_RX_DONE);
  cadCfg.cad.irqFlags |= (1UL << RADIOLIB_IRQ_TIMEOUT);
  cadCfg.cad.irqMask = cadCfg.cad.irqMask;
  state = this->phyLayer->startChannelScan(cadCfg);
```

Depending on how we would want to resolve this, all looks very good to me - thank you very much!!

StevenCellist (approved):

jgromes:
@StevenCellist 

> New IRQs passed the test through LoRaWAN's downlinkCommon() after one compilation error:

Ah, the error was caused by the old SX127x start receive method, which exposed the Rx mode directly to user. That no longer makes sense, so the Rx mode is now deduced based on timeout - fixed now.

FYI because if the changes to the channel scanning and receive methods, I will keep this open until I can run some basic checks on a few devices, hopefully in the next few days.

src/modules/SX126x/SX126x.cpp
```diff
@@ -425,12 +435,23 @@ int16_t SX126x::packetMode() {
...
-
-int16_t SX126x::scanChannel(uint8_t symbolNum, uint8_t detPeak, uint8_t detMin) {
+  ChannelScanConfig_t config = {
+    .cad = {
+      .symNum = RADIOLIB_SX126X_CAD_PARAM_DEFAULT,
+      .detPeak = RADIOLIB_SX126X_CAD_PARAM_DEFAULT,
+      .detMin = RADIOLIB_SX126X_CAD_PARAM_DEFAULT,
+      .exitMode = RADIOLIB_SX126X_CAD_PARAM_DEFAULT,
+      .timeout = 0,
+      .irqFlags = RADIOLIB_IRQ_NOT_SUPPORTED,
```
GUVWAF, line 445: This should be `RADIOLIB_IRQ_CAD_DEFAULT_FLAGS`. When calling `scanChannel()` without arguments, it was waiting forever in my testing. The config in `startChannelScan()` has it correct.

jgromes: Good catch - thank you! I probably forgot to change this at some point. Tehre was also a similar issue in the LR11x0 driver.

jgromes:
@StevenCellist @GUVWAF I was able to test this on SX1278, SX1262 and LR1110, all seems to be working (as in: transmission, reception and CAD). Hopefully I didn't miss something :)

StevenCellist:
Thank you very much!!

src/modules/LR11x0/LR11x0.cpp
```diff
@@ -460,10 +481,9 @@ int16_t LR11x0::startReceive(uint32_t timeout, uint32_t irqFlags, uint32_t irqMa
   // set DIO mapping
   uint32_t irq = irqFlags;
   if(timeout != RADIOLIB_LR11X0_RX_TIMEOUT_INF) {
-    irq |= RADIOLIB_LR11X0_IRQ_TIMEOUT;
+    irq |= (1UL << RADIOLIB_IRQ_TIMEOUT);
   }
-
-  state = setDioIrqParams(irq);
+  state = setDioIrqParams(getIrqMapped(irq));
```
GUVWAF, line 486: Just stumbled upon this: I believe due to `getIrqMapped()` the flags in `startReceive()` on line 442 should be `RADIOLIB_IRQ_RX_DEFAULT_MASK`.

jgromes: Ah, you're right - not sure why I didn't catch this in the testing, but it's fixed in e0af297c76ef09df9fef6afcafe67196ab58957f, thanks!

GUVWAF: @jgromes I'm sorry but I believe he flags shouldn't be set to `RADIOLIB_IRQ_RX_DEFAULT_FLAGS`, because for the LR11x0 it will then also set the mask to this, so it would generate interrupts on all the default flags. It should set the flags to the default masks (yes, confusing).

jgromes: Good point - since LR11x0 doesn't support the IRQ mask ,we need to the masking ourselves, much like in SX127x: `getIrqMapped(irqFlags & irqMask`

src/modules/SX127x/SX127x.cpp
```diff
@@ -1295,36 +1309,66 @@ RadioLibTime_t SX127x::calculateRxTimeout(RadioLibTime_t timeoutUs) {
...
-      return(flags & RADIOLIB_SX127X_CLEAR_IRQ_FLAG_CAD_DETECTED);
-    case RADIOLIB_IRQ_TIMEOUT:
-      return(flags & RADIOLIB_SX127X_CLEAR_IRQ_FLAG_RX_TIMEOUT);
-    default:
-      return(RADIOLIB_ERR_UNSUPPORTED);
+int16_t SX127x::setIrqFlags(uint32_t irq) {
+  // this is a bit convoluted, but unfortunately SX127x IRQ flags are not used to enable/disable that IRQ ...
+  int16_t modem = getActiveModem();
+  if(modem == RADIOLIB_SX127X_LORA) {
+    switch(irq) {
```
GUVWAF, line 1320: @jgromes It seems that with this `switch` statement, you cannot set more than one IRQ in the mask, because it checks whether `irq` is exactly equal to one specific flag.

jgromes: That is a good point - unfortunately it also hits a limitation of the SX127x, in which some combinations of IRQ flags cannot be set simultaneously. For example, Tx done and Rx done are both signalled on DIO1 and their configuration is mutually exclusive.

The only (partial) solution I can think of is to do this in a "best effort" approach, and notify the user via a return code (+mabye a debug print) if the combination of IRQ cannot be configured. For example, CAD done and CAD detected 
- which can be a common use case - may be set at the same time because they use a different DIO pin. What do you think?

GUVWAF: Yes, that sounds reasonable to me.

jgromes: OK, in that case I'll update it accordingly so that we can include it in the 7.0.1 patch ;)

jgromes: @GUVWAF fixed in 38fc7a97a4c195b7c10aa94215a1c53ec18a56ef

PR #1199, opened 2024-08-29
Reconcile implicit double / float promotion in math functions.

FrauBluher:
PR drops implicit conversion of floats and double literal values via the proposed changes. This resolves a bunch of warnings during compilation with all warnings enabled and prevents needless conversion of floats to doubles and back.

Lib continues to work as expected, though I'm utilizing this through a bespoke NonArduinoHAL in a Zephyr-based project.

PR also suppresses unused variable warning in situations where a var is used in print macros.


StevenCellist:
@FrauBluher somewhat sorry for bumping this old thread, but I'd be curious to see your Zephyr HAL! I have plans to try RadioLib on Zephyr, but as I don't have much experience yet, I'd be happy to dip into your experience :) any chance you can share it?

PR #1203, opened 2024-09-01
[SX126x][SX128x][LR11x0] Don't report CRC mismatch if valid header received

### LINKED ISSUE #1202: SX126x reports CRC mismatch upon previously received corrupted header
For SX126x (and SX128x and LR11X0) radios, a CRC mismatch is returned when either the header CRC or the payload CRC is incorrect:
https://github.com/jgromes/RadioLib/blob/d3f9eaf30b438faa2406982a75f54be64f681909/src/modules/SX126x/SX126x.cpp#L709-L710
Although it is not clear from the data sheet, in explicit header mode I would assume the radio would not continue decoding the payload when receiving a corrupt header, because it doesn’t know whether the coding rate and packet length is correct. It will then set the `HEADER_ERR` IRQ, but it will not generate the `RX_DONE` IRQ. Then, when receiving a subsequent packet, the `HEADER_ERR` is still set, while it might not be corresponding to the current packet. 

I confirmed this by enabling the `HEADER_VALID` IRQ flag and then caught a case where the `HEADER_ERR` IRQ was set, but **also** the `HEADER_VALID` IRQ, and also the payload CRC was okay. 

I think there are three options for fixing this: 
1. Simply remove the check for `HEADER_ERR`, as it’s the payload we care about anyway;
2. Enable the `HEADER_VALID` flag by default and ignore the `HEADER_ERR` when also a valid header is received;
3. Return `RADIOLIB_ERR_LORA_HEADER_DAMAGED` when only the `HEADER_ERR` IRQ is set, such that the user can decide what to do in this case.

I would like to hear what you think such that I continue with a PR.

*Edit*: After reading the SX12**7**x data sheet, I now understand why this was probably done. They mention there that if the header is corrupt, it might not have the "CRC on payload" bit set, so it would not generate the CRC error IRQ even if it was incorrect. They do not mention this in the SX12**6**x datasheet, but to be safe maybe disregard option 1.

GUVWAF:
Fixes #1202. 

Tested this based on #1190 already using an SX1262 radio. I was able to catch a packet where indeed the `HEADER_ERR` IRQ was set, but also `HEADER_VALID` and the payload CRC was OK, and the payload could be decrypted and parsed successfully.


jgromes (approved):
Oh wow, you're fast :D as discussed in the linked issue, this solution works for me - @StevenCellist I don't think this should have any negative effects on the work you're doing at the LoRaWAN stack, but please confirm.

StevenCellist:
Nope, the LoRaWAN stack doesn't touch any Header IRQ, so good to go.

jgromes:
Merged - @GUVWAF thank you for another contribution, it is greatly appreciated!

GUVWAF:
Also thanks to you both for the quick review!

PR #1204, opened 2024-09-02
[LoRaWAN] Major rework

StevenCellist:
@jgromes best of luck... :)
Culmination of a whole summer of development.

Totally non-exhaustive list of changes in somewhat random order of importance:

* Implement LoRaWAN v1.0.4 (for real), alongside v1.1
* Implement Regional Parameters v1.0.4 (v1.1 now only available as fallback)
* Rework MAC command processing ...
  * ... repeated MAC commands are now implemented (until downlink is received)
  * ... not tied anymore to arbitrary payload length and queue length (sacrificing some ease of use, but some optional MAC commands completely invalidated the previous system with a length of up to 26 bytes)
  * ... ADR is now pre- and post-processed into a complete channel mask and processes & responds as required for both v1.0.4 and v1.1 on dynamic and fixed bands (this was honestly a hell)
  * ... configurations are now only applied if the whole MAC command succeeded
  * ... and some misc bugs
* Equally use all available channels by keeping track of used channels
* Support frame retransmission as part of ADR
* Implement TR013 = CSMA = LBT (for real)
* Remove `uplink()` and `downlink()` as this separation is not possible with ADR and LBT
* Don't receive downlinks that are (much) longer than allowed
* Prevent stack crashes in case of slight user error (resolves #1175)
* Implement actual Rx1-datarate lookup table as RP 1.0.4 doesn't allow being smart
* Fix maximum payload calculation
* Add option to schedule Join/Uplink at specific time
* Add virtual tag to some functions to allow a derived class
* Slights hints of FSK and LR-FHSS modem, requires `PhysicalLayer::setModem()` to be implemented before completely working

Additional changes:
* Return Rx window number (0 for no downlink, > 0 for downlink) instead of `NO_DOWNLINK` or `ERR_NONE` - this is a 'breaking' change.
* Also use `getMaxPayloadLen()` for non-dwell-time bands for ease of use under ADR.
* Update some return codes to reflect error state.


HeadBoffin:
> Add option to schedule Join/Uplink at specific time

Is this required by some corner case on the spec - because it's the simplest way to DDOS a gateway - have all your nodes set to uplink at the top of the hour ...

It also contradicts TR007 with jitter - randomisation of uplink time - I usually use ±60 seconds for deployment.

Making the function inconspicuous is fine.

StevenCellist:
It's not in TS001, it is in another TS.. but I can surely make it private.

src/protocols/LoRaWAN/LoRaWANBands.cpp
```diff
@@ -20,16 +20,18 @@ const LoRaWANBand_t* LoRaWANBands[RADIOLIB_LORAWAN_NUM_SUPPORTED_BANDS] = {
 const LoRaWANBand_t EU868 = {
   .bandNum = BandEU868,
   .bandType = RADIOLIB_LORAWAN_BAND_DYNAMIC,
-  .payloadLenMax = {  59,  59,  59, 123, 230, 230, 230, 230,   0,   0,   0,   0,   0,   0,   0 },
+  .payloadLenMax = {  59,  59,  59, 123, 250, 250, 250, 250,  58, 123,  58, 123,   0,   0,   0 },
+  .payloadLenRel = {  59,  59,  59, 123, 230, 230, 230, 230,  58, 123,  58, 123,   0,   0,   0 },
```
jgromes, line 24: Perhaps I'm missing some context (haven't gone through the bulk of changes yet, but it looks to me that `payloadLenRel` could be replaced by something like `max(payloadLenMax[i], 230)`.

src/protocols/LoRaWAN/LoRaWANBands.cpp
```diff
@@ -20,16 +20,18 @@ const LoRaWANBand_t* LoRaWANBands[RADIOLIB_LORAWAN_NUM_SUPPORTED_BANDS] = {
...
   .dwellTimeUp = 0,
   .dwellTimeDn = 0,
+  .txParamSupported = false,
   .txFreqs = {
-    { .enabled = true, .idx = 0, .freq = 868.100, .drMin = 0, .drMax = 5},
-    { .enabled = true, .idx = 1, .freq = 868.300, .drMin = 0, .drMax = 5},
-    { .enabled = true, .idx = 2, .freq = 868.500, .drMin = 0, .drMax = 5},
+    { .enabled = true, .idx = 0, .freq = 8681000, .drMin = 0, .drMax = 5, .dr = 5 },
+    { .enabled = true, .idx = 1, .freq = 8683000, .drMin = 0, .drMax = 5, .dr = 5 },
+    { .enabled = true, .idx = 2, .freq = 8685000, .drMin = 0, .drMax = 5, .dr = 5 },
```
jgromes, line 34: I know this has been here for a while, but I only noticed it now - why is the member `idx` needed? It seems to be just incrementaly increasing for each entry, so it can be inferred if needed.

StevenCellist: Welcome to the Dark Side, where channel index is not equal to channel slot. For example the US915 band: the default subband 2 uses channels 8-15 and 65. We're not allocating 72 (80) channel slots (16 is the mandatory / recommended minimum), so we have to keep track of index too.

src/protocols/LoRaWAN/LoRaWANBands.cpp
StevenCellist: That was something I found out only after adding all of those values... Honestly quite puzzled me as the faster data rates are those that do _not_ hit dwell time limits or very long ToA in general, but they are allowed additional metadata.. oh well. Will do.

StevenCellist:
Edited original post to reflect additional change.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -6,54 +6,208 @@
...
+
+  return(state);
+}
+
+int16_t LoRaWANNode::sendReceive(const char* strUp, uint8_t fPort, uint8_t* dataDown, size_t* lenDown, bool isConfirmed, LoRaWANEvent_t* eventUp, LoRaWANEvent_t* eventDown) {
+  int16_t state = RADIOLIB_ERR_UNKNOWN;
+  
+  state = this->sendReceive((uint8_t*)strUp, strlen(strUp), fPort, dataDown, lenDown, isConfirmed, eventUp, eventDown);
+
+  return(state);
```
jgromes, line 63: `state` is unused here, so this can just `return(this->sendReceive(...));`

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -6,54 +6,208 @@
...
+}
+
+int16_t LoRaWANNode::sendReceive(uint8_t* dataUp, size_t lenUp, uint8_t fPort, uint8_t* dataDown, size_t* lenDown, bool isConfirmed, LoRaWANEvent_t* eventUp, LoRaWANEvent_t* eventDown) {
+  int16_t state = RADIOLIB_ERR_UNKNOWN;
```
jgromes, line 67: This is a public method, we need to assume the user may pass null pointers for `dataUp`, `dataDown` and/or `lenDown`. Are those a valid use case? If not, we need to catch that. If yes, we need to catch cases when the pointer is null, but length is non-zero.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -6,54 +6,208 @@
...
+    
+    // send it (without the MIC calculation blocks)
+    state = this->transmitUplink(&this->channels[RADIOLIB_LORAWAN_UPLINK],
+                                &uplinkMsg[RADIOLIB_LORAWAN_FHDR_LEN_START_OFFS], 
+                                (uint8_t)(uplinkMsgLen - RADIOLIB_LORAWAN_FHDR_LEN_START_OFFS));
+    if(state != RADIOLIB_ERR_NONE) {
+      #if !RADIOLIB_STATIC_ONLY
+      delete[] uplinkMsg;
+      #endif
+      RADIOLIB_ASSERT(state);
```
jgromes, line 142: This macro here effectively adds another if condition, which is not necessary - we already know it will fail, so just `return(state)` is better here.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -118,6 +271,166 @@ int16_t LoRaWANNode::setBufferNonces(uint8_t* persistentBuffer) {
...
+  cOcts[0] |= 0;                          // default to max Tx Power
+  (void)execMacCommand(cid, cOcts, cLen);
+
+  cid = RADIOLIB_LORAWAN_MAC_DUTY_CYCLE;
+  this->getMacLen(cid, &cLen, RADIOLIB_LORAWAN_DOWNLINK);
+  uint8_t maxDCyclePower;
+  switch(this->band->dutyCycle) {
+    case(0):
+      maxDCyclePower = 0;
+      break;
```
jgromes, line 367: If we do not want to error on an "invalid" value of `this->band->dutyCycle`, then this can be simplified by initializing `maxDCyclePower = 0` and then removing both `case(0)` as well as `default`.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -183,402 +500,167 @@ int16_t LoRaWANNode::setBufferSession(uint8_t* persistentBuffer) {
...
-    cmd.cid = RADIOLIB_LORAWAN_MAC_NEW_CHANNEL;
+    cid = RADIOLIB_LORAWAN_MAC_NEW_CHANNEL;
+    (void)this->getMacLen(cid, &cLen, RADIOLIB_LORAWAN_DOWNLINK);
     for(int i = 0; i < RADIOLIB_LORAWAN_NUM_AVAILABLE_CHANNELS; i++) {
-      cmd.len = MacTable[RADIOLIB_LORAWAN_MAC_NEW_CHANNEL].lenDn;
-      memcpy(cmd.payload, startChannelsUp + (i * cmd.len), cmd.len);
-      if(memcmp(cmd.payload, bufferZeroes, cmd.len) != 0) { // only execute if it is not all zeroes
-        cmd.repeat = 1;
-        (void)execMacCommand(&cmd);
+      memcpy(cOcts, startChannelsUp + (i * cLen), cLen);
```
jgromes, line 525: Since `cLen` can be modified by derived classes, I would strongly suggest to always check in this (and similar) `memcpy` calls that `cLen <= 14`;

jgromes: Also, a check to make sure `startChannelsUp + (i * cLen)` cannot overflow `startChannelsUp` should be added.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -183,402 +500,167 @@ int16_t LoRaWANNode::setBufferSession(uint8_t* persistentBuffer) {
...
+  (void)execMacCommand(cid, cOcts, cLen);
 
-  cmd.cid = RADIOLIB_LORAWAN_MAC_ADR_PARAM_SETUP;
-  cmd.len = MacTable[RADIOLIB_LORAWAN_MAC_ADR_PARAM_SETUP].lenDn;
-  memcpy(cmd.payload, &this->bufferSession[RADIOLIB_LORAWAN_SESSION_ADR_PARAM_SETUP], cmd.len);
-  (void)execMacCommand(&cmd);
+  cid = RADIOLIB_LORAWAN_MAC_REJOIN_PARAM_SETUP;
+  (void)this->getMacLen(cid, &cLen, RADIOLIB_LORAWAN_DOWNLINK);
+  memcpy(cOcts, &this->bufferSession[RADIOLIB_LORAWAN_SESSION_REJOIN_PARAM_SETUP], cLen);
+  (void)execMacCommand(cid, cOcts, cLen);
```
jgromes, line 576: My copy-paste sense is tingling here. A loop that iterates over an array of CID values would be more elegant.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: What would be the appropriate error for this? `RADIOLIB_ERR_MEMORY_ALLOCATION_FAILED`?

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: I am a bit split on what to do here. While you do make a somewhat valid point, there is absolutely no way in the world that this is going to cause a problem. Derived classes can only extend the MAC world (see e.g. `getMacCommand`, it first cycles through the default table before looking for additional MAC commands), and it would be bonkers for someone to modify the MAC table or session buffer indices. So to me it feels like wasted effort and flash to completely safe-guard this. Please let me know your opinion given this additional information!

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -883,333 +986,378 @@ int16_t LoRaWANNode::activateABP(uint8_t initialDr) {
...
+      break;
+    case RADIOLIB_LORAWAN_FPORT_TS009:
+      encKey = this->appSKey;
+      break;
+    case RADIOLIB_LORAWAN_FPORT_TS011:
+      encKey = this->nwkSEncKey;
+      break;
+    default:
+      encKey = this->appSKey;
+      break;
```
jgromes, line 1213: This can be done with a simpler condition (or a ternary assignment) ;)

```c++
uint8_t* encKey = this->appSKey;
if((fPort == RADIOLIB_LORAWAN_FPORT_MAC_COMMAND) | (fPort == RADIOLIB_LORAWAN_FPORT_TS011)) {
  encKey = this->nwkSEncKey;
}
```

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -883,333 +986,378 @@ int16_t LoRaWANNode::activateABP(uint8_t initialDr) {
...
+// interrupt service routine to handle downlinks automatically
+#if defined(ESP8266) || defined(ESP32)
+  IRAM_ATTR
+#endif
+static void LoRaWANNodeOnDownlinkAction(void) {
+  downlinkAction = true;
 }
 
-int16_t LoRaWANNode::downlinkCommon() {
+int16_t LoRaWANNode::receiveCommon(uint8_t dir, const LoRaWANChannel_t* dlChannels, const RadioLibTime_t* dlDelays, uint8_t numWindows, RadioLibTime_t tReference) {
```
jgromes, line 1317: Is there a specific reason why `receiveCommon` has arguments? Both calls to this method use the same values (which are themselves members of `LoRaWANNode`, or a macro. Or is this needed for the relay, seeing as it is `protected`?

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -1475,1524 +1574,1682 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len, LoRaWANEvent_t* event)
...
+    case RADIOLIB_LORAWAN_FPORT_TS009:
+      encKey = this->appSKey;
+      break;
+    case RADIOLIB_LORAWAN_FPORT_TS011:
+      encKey = this->nwkSEncKey;
+      break;
+    default:
+      encKey = this->appSKey;
+      break;
+  }
```
jgromes, line 1668: Also can be done by a simpler if condition.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -1475,1524 +1574,1682 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len, LoRaWANEvent_t* event)
...
-    #if !RADIOLIB_STATIC_ONLY
-      uint8_t* fOpts = new uint8_t[RADIOLIB_MAX(RADIOLIB_LORAWAN_FHDR_FOPTS_LEN_MASK, (int)fOptsLen)];
-    #else
-      uint8_t fOpts[RADIOLIB_STATIC_ARRAY_SIZE];
-    #endif
+  #if !RADIOLIB_STATIC_ONLY
+    uint8_t* fOpts = new uint8_t[fOptsLen];
+  #else
+    uint8_t fOpts[RADIOLIB_STATIC_ARRAY_SIZE];
+  #endif
```
jgromes, line 1640: I seem to recall there can be no more than 16 fOpts bytes in LW 1.1. If that is also the case for 1.0.x, it would make more sense to just have 16B buffer here instead of dynamic memory.

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -1475,1524 +1574,1682 @@ int16_t LoRaWANNode::downlink(uint8_t* data, size_t* len, LoRaWANEvent_t* event)
...
 
   return(true);
 }
 
-uint8_t LoRaWANNode::getMacPayloadLength(uint8_t cid) {
-  for (LoRaWANMacSpec_t entry : MacTable) {
-    // cppcheck warns here we should use std::find_if, but some platforms may not have that
-    if (entry.cid == cid) { // cppcheck-suppress useStlAlgorithm
-      return entry.lenDn;
+void LoRaWANNode::printChannels() {
```
jgromes, line 3115: Because this only actually prints in debug mode, I think it would be better to guard this with `#if RADIOLIB_DEBUG_PROTOCOL`.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: That is true for piggybacked MAC commands (in the FOpts field of the LoRaWAN header), but if FPort = 0, the whole payload consists of FOpts with theoretically up to 230-ish bytes.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Yup, re-used logic for derived classes. With different arguments there :)

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: I can imagine the following scenario:

1. I have a class that extends `LoRaWANNode`, and I have some custom MAC commands.
2. My class implements `derivedMacFinder`, overriding the default implementation from `LoRaWANNode`.
3. I have a MAC command that is longer than 14 bytes. In my implementation of `derivedMacFinder`, that command will be returned.
4. Now, when `LoRaWANNode` tries to handle that command, it will lead to buffer overflows when copying, for example in the place I commented on (though it is not the only one).

I think the fix can be easy here -  MAC command returned by `derivedMacFinder` can be validated, at least by checking the length. My approach here would be to handle `virtual` classes as a form of user input. It takes a more advanced user to take advantage of it, sure - but user input it still is.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: Ah, I see - thanks!

src/protocols/LoRaWAN/LoRaWAN.h
```diff
@@ -1016,94 +979,128 @@ class LoRaWANNode {
...
-    void setCSMA(uint8_t backoffMax, uint8_t difsSlots, bool enableCSMA = false);
+    // push MAC command to queue, done by copy
+    int16_t pushMacCommand(uint8_t cid, uint8_t* cOcts, uint8_t* out, uint8_t* lenOut, uint8_t dir);
+
+    int16_t getMacPayload(uint8_t cid, uint8_t* in, uint8_t lenIn, uint8_t* out, uint8_t dir);
+
+    // delete a specific MAC command from queue, indicated by the command ID
+    int16_t deleteMacCommand(uint8_t cid, uint8_t* inOut, uint8_t* lenInOut, uint8_t dir);
+
+    void clearMacCommands(uint8_t* inOut, uint8_t* lenInOut, uint8_t dir);
```
jgromes, line 1049: While these are protected members, it would still be good to have at least a short comment explaining the methods here, like the others have.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: That's fair, but for all of these calls, the CID is fixed right before. E.g.:
```cpp
    cid = RADIOLIB_LORAWAN_MAC_NEW_CHANNEL;
    (void)this->getMacLen(cid, &cLen, RADIOLIB_LORAWAN_DOWNLINK);
    for(int i = 0; i < RADIOLIB_LORAWAN_NUM_AVAILABLE_CHANNELS; i++) {
      memcpy(cOcts, startChannelsUp + (i * cLen), cLen);
```
For all of these calls, the CID is present in the `MacTable` in `LoRaWAN.h`, which means that `getMacCommand` always finds this MAC command and will never even call `derivedMacFinder`.

src/TypeDef.h
```diff
@@ -564,9 +564,9 @@
 #define RADIOLIB_ERR_CHECKSUM_MISMATCH                          (-1115)
 
 /*!
-  \brief No downlink was received - most likely none was sent from the server.
+  \brief No JoinAccept was received - check your keys, or otherwise likely a range issue!
 */
-#define RADIOLIB_LORAWAN_NO_DOWNLINK                            (-1116)
+#define RADIOLIB_LORAWAN_NO_JOIN_ACCEPT                         (-1116)
```
jgromes, line 569: MAybe this as well as `RADIOLIB_LORAWAN_INVALID_MODE`, `RADIOLIB_LORAWAN_INVALID_MODE` and others should use the prefix `RADIOLIB_ERR_LORAWAN_`? These are errors after all.

src/protocols/LoRaWAN/LoRaWAN.cpp
jgromes: We can keep it as it is provided `derivedMacFinder` never gets called - I was not able to follow the logic of `getMacLen` everywhere, hence my paranoia.

StevenCellist:
@jgromes all done (plus minor edit of original post)

jgromes (approved):

jgromes:
Merged - great job @StevenCellist! I ran the full CI manually and it still fails in couple of places, but that's unrelated to this PR, so I will fix it later.

StevenCellist:
Woohoo! 🎉

PR #1206, opened 2024-09-07
fix getRangingResult to properly account for signed numbers

Jorropo:
I got numbers like this as I approached the slave (raw values before converting to meters):
```
Ranged: 8
Ranged: 6
Ranged: 3
Ranged: 3
Ranged: 5
Ranged: 5
Ranged: 5
Ranged: 6
Ranged: 1
Ranged: 1
Ranged: 800000
Ranged: 800003
Ranged: 800003
```

This is because the ToF becomes smaller than the correction factor resulting in a negative number.

This patch performs Sign Extension from 24bits to 32bits. This result in returning a negative meter value which makes more sense than an impossibly big one.

### DIFF
--- a/src/modules/SX128x/SX1280.cpp
+++ b/src/modules/SX128x/SX1280.cpp
@@ -176,7 +176,8 @@ float SX1280::getRangingResult() {
   RADIOLIB_ASSERT(state);
 
   // calculate the real result
-  uint32_t raw = ((uint32_t)data[0] << 16) | ((uint32_t)data[1] << 8) | data[2];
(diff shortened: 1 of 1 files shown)


Jorropo:
This could also happen if your ToF beacon time travel from the future. Altho I lack the hardware to test this properly.

jgromes:
Looks good - merged, thank you for the contribution!

> This could also happen if your ToF beacon time travel from the future. Altho I lack the hardware to test this properly.

True, but I do not think we should we should be too worried about handling the presumably rare cases when the SX128x device doing the ranging is moving at superluminal speeds :)

PR #1216, opened 2024-09-12
LR11x0 GNSS Almanac and more

tve:
- Make LR11x0 GNSS Almanac update from satellites work
- Improve other GNSS functionality, such as power measurements and getting a position
- Add example sketch
- Move much of the debug printing into the example sketch


tve:
This is not yet ready for prime-time but I believe it's a step forward and cleans some things up. Lots more work to do, some marked with FIXME comments.

lyusupov:
@tve Thanks for the impressive work!

Is there any chance to add an equivalent of PPS signal functionality that would synchronize local MCU time with precise atomic satellites clock ?

tve:
> Is there any chance to add an equivalent of PPS signal functionality that would synchronize local MCU time with precise atomic satellites clock ?

The lr1110 does not have a PPS output signal. It does sync and slew its clock and it's possible to query the time over SPI so depending on the accuracy you need you can keep the MCU's RTC in-sync.

lyusupov:
Thank you for the explanation!

> It does sync and slew its clock and it's possible to query the time over SPI 

It would be nice to have an example in the RadioLib on how to do this...

PR #1222, opened 2024-09-19
getRssi(false) without packet  

Genaker:
Add getRssi() without packet like SX1262 works. Tested works with SX1280


jgromes:
CI runs have passed and all looks good - merged, thank you for the contribution!

PR #1228, opened 2024-09-22
[SX126x] LR-FHSS support

jgromes:
This pull request adds support for LR-FHSS transmission from the SX126x module.

It is currently only verified by comparison of the transmitted signal to one sent from LR1110. The frequency hopping and packet length seems to match, however, it is not possible to test the encoding this way. A proper test would involve a LR-FHSS-capable LoRaWAN gateway.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, ARM64]
     permissions:
       security-events: write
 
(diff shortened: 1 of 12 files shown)


c3podaniel:
https://github.com/jumanamirza/LR-FHSS-receiver

Have you already seen this?

jgromes:
@c3podaniel no, I haven't seen that project (thanks for pointing me to it!), however, I'm not sure how I would use it. From what I can tell it works with recorded data, but it does not specify how they are recorded.

c3podaniel:
> @c3podaniel no, I haven't seen that project (thanks for pointing me to it!), however, I'm not sure how I would use it. From what I can tell it works with recorded data, but it does not specify how they are recorded.

This is the paper behind the project. https://dl.acm.org/doi/10.1145/3694971
They used a SDR to record the data.

jgromes:
@c3podaniel that is actually a pretty good paper, thanks - shame the repository itself doesn't link to it. I'll take a look, maybe this could be a way to validate this PR.

StevenCellist:
The used SDR is a casual €2200, wow. Thought I'd maybe get an SDR after all, but definitely _not_ that one! 😆

jgromes:
After some modifications, I was able to run the [LR-FHSS-Receiver](https://github.com/jumanamirza/LR-FHSS-receiver) project in Octave. I also figured out how to record the data in a format the project expects, with the following simple GNUradio flowgraph:

Screenshot_109

With that, I was able to decode the headers, so at least those are being constructed, encoded and sent correctly:

```
lrfh sim detecting header ... done
Elapsed time is 3.61704 seconds.
decoding candi header 1, found header 1: idx 2, CR 3, start 54421, freq 16260.98, power 4.16, cost 27.91
decoding candi header 2, FAIL, coarse time 166003, coarse freq -57540.00
decoding candi header 3, found header 2: idx 1, CR 3, start 171211, freq -57725.85, power 5.57, cost 25.56
decoding candi header 4, found header 3: idx 0, CR 3, start 287723, freq -26739.02, power 4.67, cost 33.75
```

However, when decoding the payload, the receiver unfortuantely uses `vitdec` function from Matlab's communication toolbox, which is not available in Octave, so I will have to figure out a way around that. But since the header is already encoded with 1/2 convolutional code and obviously decodes correctly, I'm about 90% sure the payload is encoded correctly as well.

@StevenCellist 

> The used SDR is a casual €2200

The B210 is a total overkill for this, you should be able to do with a cheap RTL-SDR dongle ;) They used just 500 kSps sampling rate for the EU868 data rates - B210 can go up to 56 MSps. My HackRF One actually can't even go down to 500 kSps, hence the resampler in my flowgraph above.

PR #1236, opened 2024-09-26
[SX127x] Fix missing CRC mismatch error

GUVWAF:
After upgrading to version 7.0 we saw that SX127x radios sometimes got garbled data through. Seems due to #1184 the variable `state` gets overwritten (normally to 0) and thus it would loose a potential `RADIOLIB_ERR_CRC_MISMATCH` set before this.
I therefore introduced a new variable for it, and I also moved the SPI calls outside the `else` clause, because the data might still be valuable even if the header is damaged.


jp-bennett:
Just tested this with Meshtastic master and a T-Beam with the older RF95 chipset. With this patch, we are once again catching CRC errors. Thanks!

jgromes:
Good catch - merged, thank you! This looks like a rather serious issue (price we pay for not releasing more often I suppose ...), I will see if we can get 7.0.1 out in the next few days.

PR #1237, opened 2024-09-26
[LR11x0] Implement automatic and forced LDRO

GUVWAF:
Fully based on the implementation of the SX126x. 

`ldroAuto` was already declared and set to true but `ldrOptimize` was never set.
Previously I couldn’t communicate with an SX126x when using a bandwidth of 125kHz and SF11 with automatic LDRO. With this on the LR11x0, it is fixed.

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -1414,6 +1414,33 @@ void LR11x0::setRfSwitchTable(const uint32_t (&pins)[Module::RFSWITCH_MAX_PINS],
   this->setDioAsRfSwitch(enable, modes[0], modes[1], modes[2], modes[3], modes[4], modes[5], modes[6]);
 }
 
+int16_t LR11x0::forceLDRO(bool enable) {
+  // check packet type
+  uint8_t type = RADIOLIB_LR11X0_PACKET_TYPE_NONE;
+  int16_t state = getPacketType(&type);
+  RADIOLIB_ASSERT(state);
+  if(type != RADIOLIB_LR11X0_PACKET_TYPE_LORA) {
+    return(RADIOLIB_ERR_WRONG_MODEM);
+  }
+
+  // update modulation parameters
+  this->ldroAuto = false;
+  this->ldrOptimize = (uint8_t)enable;
+  return(setModulationParamsLoRa(this->spreadingFactor, this->bandwidth, this->codingRate, this->ldrOptimize));
+}
+
+int16_t LR11x0::autoLDRO() {
+  uint8_t type = RADIOLIB_LR11X0_PACKET_TYPE_NONE;
+  int16_t state = getPacketType(&type);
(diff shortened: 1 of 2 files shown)


jgromes:
Merged, thank you! To be honest I forgot about the LDRO in the original LR11x0 implementation ...

Any more PRs we should expect from Meshtastic?

caveman99:
The 2.4G stuff for the TLora T3S3 LR1121 would make our needs complete i think... :-)

GUVWAF:
Thanks for the quick reviews @jgromes!

That's it for now I think, although we still have an issue with our non-standard sync word (0x2b); we can't receive packets from SX127x radios on LR11x0 radios. When changing to the default private network sync word (0x12), this seems resolved. @caveman99 has contacted Semtech already, but we're still waiting for a response.

jgromes:
> @caveman99 has contacted Semtech already, but we're still waiting for a response.

Let me know if there's some results from that (perhaps some more magic nibbles to be enabled in some register ... ?) ;) There have been other questions about the LR1110, mainly about the GNSS scanning (see #1216 for a list), don't know if that's something you guys are also looking into.

One a bit off-topic question: It seems like every time we release a larger RadioLib update we cause some issues in Meshtastic. Try as I might, it's not really in my power to test every combination of possible radio/platform interfaction. Would it be worth it for you to have access to some RadioLib pre-release/release candidate before we ship major updates? It feels like we just dump them into the world now, and in larger projects (like Meshtastic) that can cause probles when applying the update.

caveman99:
Sometimes we're tracking commits of radiolib in our own development process for new platforms, notably the LR11xx, so in principle we are aware of any ABI breakage before you release a version. The stuff that came up today was discovered by chance during the hunt for another bug. I don't think we would have caught that prematurely, we usually parse the release notes of radiolib to see if there's interesting new stuff or bugfixes inside :-)

However there's an effort to establish a platform for end2end tests with hardware nodes, and once we get that off the ground, a 'bleeding' radiolib version could be included in the test rig. I'll keep you posted on the progress.

caveman99:
The GNSS scanning is on our list for sure, but not on high priority. The first end user device with LR1110 has a dedicated GPS chip in the design.

jgromes:
@caveman99

>  I don't think we would have caught that prematurely

OK, thanks for letting me know. I definitely don't want to be doing pre-releases only for nothing to happen with them, so I think we can keep this as is.

> However there's an effort to establish a platform for end2end tests with hardware nodes

That would be very cool! I did some preparatory work for hardware-in-the-loop testing [here](https://github.com/jgromes/RadioLib/blob/master/extras/test/SX126x/main.cpp), however, I realized very quickly that I would need quite a few devices to make this work in any meaningful way (testing communication between different radios for example) or even including an SDR somewhere along the line. So now the test only tests whether it can talk to the SX126x in my experimental LoRaWAN gateway.

PR #1240, opened 2024-09-27
[LR11x0] Fix CAD IRQ setting

GUVWAF:
Just found this one. 
It seems without this the IRQs for **`CRC_ERR`** and `CAD_DONE` were set (that's why it did work in testing), but it would thus never fire `CAD_DETECTED`.


jgromes:
That's odd, I could swear I did this in #1190 ... I'm probably confusing it with SX126x/8x.

Merged, thank you for the fix!

Issue #1244, opened 2024-09-30
[SX127x] IRQ mask handling uses the wrong DIO register


caveman99:
https://github.com/jgromes/RadioLib/commit/38fc7a97a4c195b7c10aa94215a1c53ec18a56ef checks the DIO mapping for collisions. Compared to the code before, the assignment in line 1403 swaps `RADIOLIB_SX127X_REG_DIO_MAPPING_1` for `RADIOLIB_SX127X_REG_DIO_MAPPING_2` and vice versa. Mapping_1 is used for DIO4 now and Mapping_2 for DIO0-3

PR #1246, opened 2024-09-30
[SX127x] fix DIO register mapping

caveman99:
this fixes https://github.com/jgromes/RadioLib/issues/1244

### DIFF
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -1400,7 +1400,7 @@ int16_t SX127x::setIrqFlags(uint32_t irq) {
 
     // DIO pin is unused, set the flag and configure it
     usedPinFlags |= (1UL << dioNum);
-    uint8_t addr = (dioNum > 3) ? RADIOLIB_SX127X_REG_DIO_MAPPING_1 : RADIOLIB_SX127X_REG_DIO_MAPPING_2;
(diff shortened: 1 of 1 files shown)


jgromes:
Well that's a rather silly mistake isn't it ... fortunately these new IRQ methods are not being used internally by SX127x or LoRaWAN (yet), so I don't think we need to push this as another hotfix. Regardless, thank you for the fix!

caveman99:
unfortunately the setting of the registers is used in startReceive() https://github.com/jgromes/RadioLib/blob/5a9ff5a4912f87918390348c8caa54ea8a6afada/src/modules/SX127x/SX127x.cpp#L392C1-L401C26 

After upgrading to 7.0.1 we were seeing RX not triggering the IRQ any more. It does not warrant a new release, we can just track the commit point for our builds, and i still have to check if this is really the fix for those problems. Test node is building as i write :-)

jgromes:
I see - that is a significant issue then. I will push this as 7.0.2 then, otherwise it will be broken for everyone else using 7.0.1 and not tracking the latest commit.

caveman99:
I can confirm, operating two TLora 2.1-1.6 with SX1276 alongside each other, with the fix RX is fine, without it does not trigger the IRQ.

jgromes:
FYI @caveman99 released as 7.0.2, hopefully that will be the last hotfix for a while ...

caveman99:
Thanks a lot for the quick response! :-) I hope so too.

Issue #1249, opened 2024-10-02
[SX128x] Cannot transmit after sleep()


jacobeva:
**Describe the bug**
After putting the module into sleep mode, even if the module is woken up after by pulling SS LOW, all attempts to transmit will result in `RADIOLIB_ERR_TX_TIMEOUT`. In my example sketch below, the "transmitting..." line is never sent over serial if the call to sleep is not commented out.

**To Reproduce**
```c++
#include <Arduino.h>
#include "Adafruit_TinyUSB.h"
#include <SPI.h>
#include <variant.h>
#include <RadioLib.h>

    SX1280 radio = new Module(24, 15, 16, 25);

volatile bool packet_rx = false;
int packets = 0;

void setflag() {
    packet_rx = true;
}

void setup() {
    Serial.begin(115200);
    while(!Serial);
    uint8_t sf = 10;
      int state = radio.begin(2408, 1625, sf, 5);// 0x14 1, 30);
      radio.setRfSwitchPins(19, 20);
      radio.explicitHeader();
radio.setCRC(2);
      //radio.setSyncWord(0x14, 0x24);
  if (state == RADIOLIB_ERR_NONE) {
    Serial.println(F("success!"));
  } else {
    Serial.print(F("failed, code "));
    Serial.println(state);
    while (true) { delay(10); }
  }
  radio.sleep();
}

void loop() {
    digitalWrite(24, LOW);
    delay(1);
    digitalWrite(24, HIGH);
    delay(1);
    radio.standby();
    uint8_t data[255] = {0};
    for (int i = 0; i < 255; i++) {
        data[i] = i;
    }
    int16_t status;
    while (true) {
    status = radio.transmit(data, 255);
    //radio.transmit(data, 7);
    if (status == RADIOLIB_ERR_NONE) Serial.println("transmitting...");
    }
}
```

</p>
</details> 

**Expected behavior**
Transmitting after sleep should work as expected.

**Additional info (please complete):**
 - Library version 7.0.0

jgromes:
Resolved by merging #1250

PR #1250, opened 2024-10-02
[SX128x] Save context before sleep to allow TX to work on wake (#1249)

jacobeva:


jacobeva:
Fixes #1249

jgromes:
Thank you for the contribution! It looks like I missed this in the SX1280 driver implementation. Though in my defense on the SX126x (which has nearly identical SPI interface), this configuration is not needed.

All loooks good - merged, thanks again!

jacobeva:
It's a weird one to be honest. I only figured out what the issue was by
taking a look at the source of Stuart's Lib "SX12XX Arduino".

PR #1251, opened 2024-10-02
[SX128x] Add setDataRate method for LoRa modem

jacobeva:
This PR is tested (on an SX1280 only). It adds the setDataRate function for the LoRa modem of the SX128x series chips, as I was using this modem in a situation where it was preferable to address it through the use of a `PhysicalLayer` pointer instead of the base class. I noticed setDataRate wasn't implemented yet, so I implemented it myself.

One thing I should mention, would it be a good idea to add support for the ranging modem also? This could always be done at a later date of course, but currently I don't have time to look into it.

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -902,6 +902,22 @@ int16_t SX128x::setPreambleLength(uint32_t preambleLength) {
   return(RADIOLIB_ERR_WRONG_MODEM);
 }
 
+int16_t SX128x::setDataRate(DataRate_t dr) {
+  // check active modem
+  uint8_t modem = getPacketType();
+  int16_t state = RADIOLIB_ERR_NONE;
+  if (modem == RADIOLIB_SX128X_PACKET_TYPE_LORA) {
+      state = this->setBandwidth(dr.lora.bandwidth);
+      RADIOLIB_ASSERT(state);
+      state = this->setSpreadingFactor(dr.lora.spreadingFactor);
+      RADIOLIB_ASSERT(state);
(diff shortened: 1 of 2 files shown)


jgromes:
Thank you for the contribution! The reason this was not implemented in SX128x is that there wasn't a need for it yet.

Regarding the ranging modem, I don't think it's needed and/or very useful. The primary use case for abstract interface methods (such as `setDataRate`) is to provide cross-module support, mainly for LoRaWAN. Ranging is quite a unique feature of the SX128x series. It is also supported by LR11x0, but not implemented in RadioLib yet.

Do you have a specific use-case for ranging modem configuration via `setDataRate` in mind? If not, then I think this is good to be merged as-is.

jacobeva:
To be honest I haven't even looked into how I would use the ranging
modem yet. Maybe let's just merge this for now, and I can always open
another PR later.

In my current project, RNode Firmware CE, I don't like keeping the
pointer to the base class (I have an array of pointers to each modem
configured for the device), so I simply have one array of PhysicalLayer*
for example. That is why I'd like to use the ranging modem with
`setDataRate`, but thinking about it, this may not be possible anyway
because there's likely other methods I'd need to use which won't be
supported by PhysicalLayer anyway!

jgromes:
> Maybe let's just merge this for now, and I can always open another PR later.

Agreed - merged, than you again!

> this may not be possible anyway because there's likely other methods I'd need to use which won't be supported by PhysicalLayer anyway!

Depends on what you're trying to do. Ranging is pretty well out-of-scope of the PhysicalLayer class, it is intended to abstract basic stuff, like transmission and reception.

PR #1275, opened 2024-10-16
[LR11x0] GNSS support

jgromes:
Based on the work that has been done in #1216 - full credit to @tve for that!

I've reworked the GNSS API to make it more easily usable. I've also split the single example sketch into three different examples (almanac update from satellites, autonomous position and satellite information dump).

Couple more small cleanup steps are still missing (keywords, some fixes etc.), but this is mostly ready.


src/modules/LR11x0/LR11x0.h
```diff
@@ -564,6 +565,65 @@
...
+
+// RADIOLIB_LR11X0_CMD_GNSS_READ_ALMANAC_STATUS
+#define RADIOLIB_LR11X0_GNSS_ALMANAC_STATUS_UP_TO_DATE          (0)             //  7     0     GPS/BeiDou almanac status: all satellites up-to-date
+#define RADIOLIB_LR11X0_GNSS_ALMANAC_STATUS_OUTDATED            (1)             //  7     0                                at least one satellite needs update
+
+// RADIOLIB_LR11X0_CMD_GNSS_READ_DOPPLER_SOLVER_RES
+#define RADIOLIB_LR11X0_GNSS_SOLVER_ERR_NONE                    (0)             //  7     0     internal 2D solver error: no error
+#define RADIOLIB_LR11X0_GNSS_SOLVER_ERR_RES_HIGH                (1)             //  7     0                               residue too high
+#define RADIOLIB_LR11X0_GNSS_SOLVER_ERR_NOT_CONVERGED           (2)             //  7     0                               not converged on solution
+#define RADIOLIB_LR11X0_GNSS_SOLVER_ERR_NOT_ENOUG_SV            (3)             //  7     0                               not enough satellites
```
tve, line 620: typo: NOT_ENOUGH_SV

jgromes: Fixed in [5597899](https://github.com/jgromes/RadioLib/pull/1275/commits/5597899dad125aefb7de771f67abe5c2423f26d9)

jgromes:
With the last few fixes, I think this concludes the GNSS suport for LR1110/20. To be honest I'm mildly disappointed by the performance, as the error I was seeing was always in the range of 10-15 km, even with 7+ satellites. It remains to be seen whether the performance with LoRaCloud will be better.

tve:
> To be honest I'm mildly disappointed by the performance, as the error I was seeing was always in the range of 10-15 km, even with 7+ satellites. It remains to be seen whether the performance with LoRaCloud will be better.

This is exactly what I wrote in [the discussion](https://github.com/jgromes/RadioLib/discussions/1211#discussioncomment-10653023). Using the cloud service you get the usual ~10 meters of accuracy.

Why this is so: the location produced by the chip itself is based on almanac data, which is coarse satellite trajectory info. The chip does not download ephemeris data, which is fine-grained satellite trajectory info and is necessary to produce a more accurate position.

jgromes:
@lyusupov good question - probably just an oversight on my part. Looks like that change happened in https://github.com/jgromes/RadioLib/pull/1275/commits/74a8d9abf226bc4579c7979aaa926c801e9775b5, however, there's many other changes in that commit as well, so I'm not 100% certain it was intentional. I'll test whether this is really necessary and if not, I will return it to the state that was merged in #1275.

tve:
Because, as stated in AN1200-74:
image

jgromes:
@lyusupov looks like @tve is correct, DIO11 is required for GNSS scanning. Setting any sort of configuration for DIO11 in this case breaks advanced GNSS scanning. The clean way to resolve this is what you sugest, checking whether GNSS scanning is enabled and if so, reserve DIO11 for that purpose.

jgromes:
@lyusupov FYI, changed in [ef16e43](https://github.com/jgromes/RadioLib/commit/ef16e431af979538327967944df56af7e727675b)

lyusupov:
> @lyusupov FYI, changed in [ef16e43](https://github.com/jgromes/RadioLib/commit/ef16e431af979538327967944df56af7e727675b)

Great job, thanks!

PR #1286, opened 2024-10-23
[LR11X0] pay more attention to selection of FSK preamble detector length

lyusupov:
## Brief description

Most of FSK radio protocols supported by SoftRF project are using 8-bit FSK preamble:
1) https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/protocol/radio/OGNTP.h#L24-L25
2) https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/protocol/radio/ADSL.h#L24-L25
3) https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/protocol/radio/Legacy.h#L25-L26

Current RadioLib code uses explicit 16 bit FSK preamble detector length

image

This causes **almost 95% loss of packets** while trying to receive radio packets sent by other (SX1262 and SX1276) devices  ( these packets use 8-bit FSK preamble ).

This PR is filed to fix the issue of not receiving 8-bit FSK preamble packets.

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -1117,7 +1117,11 @@ int16_t LR11x0::setPreambleLength(size_t preambleLength) {
     return(setPacketParamsLoRa(this->preambleLengthLoRa, this->headerType,  this->implicitLen, this->crcTypeLoRa, (uint8_t)this->invertIQEnabled));
   } else if(type == RADIOLIB_LR11X0_PACKET_TYPE_GFSK) {
     this->preambleLengthGFSK = preambleLength;
-    this->preambleDetLength = RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_16_BITS;
+    this->preambleDetLength = preambleLength >= 32 ? RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_32_BITS :
+                              preambleLength >= 24 ? RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_24_BITS :
+                              preambleLength >= 16 ? RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_16_BITS :
+                              preambleLength >   0 ? RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_8_BITS :
+                              RADIOLIB_LR11X0_GFSK_PREAMBLE_DETECT_DISABLED;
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you for the contribution! Everything seems OK so this can be merged.

I think it would also make sense to add a similar configuration to SX126x, which also uses 16 bits as the default, so I would expect it to behave the same as LR11x0 in terms of the packet loss. SX127x does not change the default setting which should be 8 bits, but the configuration could also be added. Would you be willing to add those changes in another PR? If not I will add it myself.

Thanks again!

lyusupov:
> I think it would also make sense to add a similar configuration to SX126x, which also uses 16 bits as the default, so I would expect it to behave the same as LR11x0 in terms of the packet loss.

I am not sure that it is necessary....

The SoftRF uses other libraries to drive SX1276 and SX162.radios on a regular basis

However, before I got access to the actual LR1110 device - I've emulated the logic of the LR11XX SoftRF driver by building and running on SX1262: https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/driver/radio/radiolib.cpp#L60-L61

I can say that I **have not noticed any issues** while running FSK modulated radio protocols with 8-bit preamble together with **SX126X RadioLib driver**.

It looks like the LR11X0 uses different values:
1) preamble length for Transmitter
2) preamble Detect length for Receiver

jgromes:
> It looks like the LR11X0 uses different values:
> 1. preamble length for Transmitter
> 2. preamble Detect length for Receiver

That is also the case for SX126x, the last argument sets detector length to `RADIOLIB_SX126X_GFSK_PREAMBLE_DETECT_16`:
https://github.com/jgromes/RadioLib/blob/709fbdb13a6a5a2010d0f2887c8cde6f6cadc1a2/src/modules/SX126x/SX126x.h#L1236

So I would expect it to behave in a simlar to way tthe LR11x0 did. I will go ahead and add this, even if it seems like this is not causing a problem, there might be some other reason for that.

lyusupov:
Yes, I agree that SX126x needs a proper setting of `preambleDetectorLen` as well:

image

Hmm, I am curious, why I have not noticed high Rx packets loss ratio with 8-bit preamble detector on **SX1262 and RadioLib.**..
I should probably test it again, when able.

SoftRF project uses a customized BASICMAC library to drive SX1262 on a regular basis.
We select the preamble detector length as follows:

https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/libraries/arduino-basicmac/src/lmic/radio-sx126x.c#L503-L529

PR #1294, opened 2024-10-26
[PHY] Get/Set modem

jgromes:
This PR adds an interface to get and set modem type (LoRa, FSK or LR-FHSS) through PhysicalLayer. Needed for LoRaWAN to support FSK and LR-FHSS uplinks.

EDIT: Also drupped Arduino Uno from the CI build list, since the library is too large to build on a large portion of the examples.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
         # platform-dependent settings - extra board options, board index URLs, skip patterns etc.
         include:
           - id: arduino:avr:uno
-            run: echo "skip-pattern=(STM32WL|SSTV|LoRaWAN|LR11x0_Firmware_Update|Pager|APRS|Morse|SX126x)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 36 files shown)


lyusupov:
Jan,
use of too much generic symbols here:
https://github.com/jgromes/RadioLib/blob/master/src/protocols/PhysicalLayer/PhysicalLayer.h#L138-L140

```
/*!
  \enum ModemType_t
  \brief Type of modem, used by setModem.
*/
enum ModemType_t {
  FSK = 0,
  LoRa,
  LRFHSS,
};

```
\- **causes namespace conflict with other software** at build time

Could you, please, consider to replace it with more RadioLib specific names such as:

```
/*!
  \enum ModemType_t
  \brief Type of modem, used by setModem.
*/
enum ModemType_t {
  RADIOLIB_FSK = 0,
  RADIOLIB_LoRa,
  RADIOLIB_LRFHSS,
};
```

Thank you!

PR #1308, opened 2024-11-02
hal: Tock: Re-enable pin after detaching interrupt

alistair23:
The Tock libtock_lora_phy_gpio_disable_interrupt() syscall will disable interrupts for the pin, but also put the pin into a disabled low power state. This isn't what RadioLib expects and casues subsequent LoRaWAN transfers to fail [1].

So after we disable interrupts and send the pin to low power let's re-enable inputs as RadioLib expects.

1: https://github.com/jgromes/RadioLib/discussions/1303


StevenCellist:
Good find, happy you were able to resolve it with that hint :) 
I guess it'll work, so merged.

PR #1309, opened 2024-11-03
[SX127x/RF69] Added setFifoThreshold

SzczepanLeon:
This PR add possibility to set custom FIFO threshold level.

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -321,6 +321,10 @@ void RF69::clearFifoEmptyAction() {
   clearDio1Action();
 }
 
+void RF69::setFifoThreshold(uint8_t threshold) {
+  this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_FIFO_THRESH, threshold, 6, 0);
+}
+
 void RF69::setFifoFullAction(void (*func)(void)) {
   // set the interrupt
   this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_FIFO_THRESH, RADIOLIB_RF69_FIFO_THRESH, 6, 0);
--- a/src/modules/RF69/RF69.h
+++ b/src/modules/RF69/RF69.h
@@ -650,6 +650,8 @@ class RF69: public PhysicalLayer {
 
     /*!
       \brief Set FIFO threshold level.
+      Be aware that threshold is also set in setFifoFullAction method.
+      setFifoThreshold method must be called AFTER calling setFifoFullAction!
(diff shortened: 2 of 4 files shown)


jgromes:
I'm not sure I understand the use-case here. Sicne the FIFO threshold register is also being set from `setFifoFullAction`, which uses a constant, so it will overwrite whatever was set by `setFifoThreshold`.

SzczepanLeon:
Lets assume that in FSK mode I don't know packet len. It can vary from 12 to 200 (for example), but it is defined in one of first received bytes (lets say it is 3rd received byte).

So my flow is:
  - set infinite mode
  - set Rx interrupt function
  - set FIFO threshold to 10 (less than minimum packet len - in my example 12)
when first interrupt is triggered (first part of data is received) I can read packet len and:
  - set FIFO threshold to MAX
  - on interrupt get more data 
  - ...

jgromes:
That makes sense, however, the problem I have is that this will only work if the user calls `setFifoThreshold` *after* `setFifoFullAction`. Ideally those should be independent of each other.

Additionally, because Stream mode is not specific to SX127x, this should also be added to the RF69 class.

SzczepanLeon:
I see one option, remove _setFifoFullAction_, add _setFifoAction_ with threshold as a parameter with default option Full/MAX

jgromes:
> I see one option, remove setFifoFullAction, add setFifoAction with threshold as a parameter

I was thinking about that too, and backwards compatibility can be addressed by using a default value. The thing I'm more worried about is repeatedly calling `setDio1Action`, some HALs do not like that.

Probably the easiest option is to just make this limitation explicit in the description of the `setFifoThreshold` method. That way if users see some unexpected behavior, the docs can point them in the rigth direction (as opposed to digging through the source code).

SzczepanLeon:
What you mean by _Stream mode_?
CC1101 also support interrupt on GDO when FIFO is filed above X.
obraz
From CC1101 point it is not stream - it is ordinary SPI read triggered by event.
Seems that only `attachInterrupt` is missing n CC1101 class.

jgromes:
I meant [this](https://github.com/jgromes/RadioLib/blob/master/examples/Stream/Stream_Receive/Stream_Receive.ino) - all the FIFO methods are only necessary in cases where the packet length exceeds the size of the radio FIFO. For CC1101, this is not implemented yet - see https://github.com/jgromes/RadioLib/discussions/1138 for proof it is not as simple as attaching an interrupt ;)

jgromes:
Looks good now - merged, thank you for the contribution!

PR #1313, opened 2024-11-07
hal: Tock: Namespace HAL, update timer implementation and update libtock-c

alistair23:
This is based on the discussion here: https://github.com/tock/libtock-c/pull/456#issuecomment-2460931559

This namespaces the Tock HAL public macro and structs and uses the native timer implementation.

This also updates to the latest libtock-c commit. libtock-c now includes a libtockHal.h, so we can use that instead of the version in RadioLib.


alistair23:
The Tock CI failure is because there is a chicken and egg problem with the libtock-c submodule. We need this merged and then update libtock-c based on these changes.

jgromes:
Looks good to me! @alistair23 since the PR in libtock was merged, I guess that now this is ready to be merged as well, correct?

alistair23:
> Looks good to me! @alistair23 since the PR in libtock was merged, I guess that now this is ready to be merged as well, correct?

Yep! This is ready to go

jgromes:
Merged, thank you!

Issue #1368, opened 2025-01-03
SX1278 setFrequency() does not work anymore in 7.1.1


BBluemer:
Hello,
with my setup and minimal working example code (see below) the frequency setting of the SX1278 does not work anymore with library version 7.1.1. Using library version 7.1.0 it works as expected. 
The Debug output below is identical with both Library Versions.
Thanks for any help.

<details><summary>Debug mode output</summary>
<p>

```
Hello World!
SX1278: INIT returns 0
SX1278: TRANSMIT_DIRECT returns 0
```

</p>
</details> 

<details><summary>Sketch that is causing the module fail</summary>
<p>

```c++
#include <RadioLib.h>

SX1278 radio = new Module(10, 4, 5, 3);
int sx1278_ret = 0;


/*------------------------------------------------------------------------------------------*/
/* setup()                                                                                  */
/*------------------------------------------------------------------------------------------*/
void setup() {

  // System
  Serial.begin(115200);
  delay(500);
  printf("Hello World!\n\r");

  // SX1278 Init
  radio.reset();
  sx1278_ret = radio.beginFSK(450, 19.2, 7.7, 25, 17, 16, false);
  printf("SX1278: INIT returns %d\n\r", sx1278_ret);

  sx1278_ret = radio.transmitDirect();
  printf("SX1278: TRANSMIT_DIRECT returns %d\n\r", sx1278_ret);

}

/*------------------------------------------------------------------------------------------*/
/* loop()                                                                                   */
/*------------------------------------------------------------------------------------------*/
void loop() {
}
```

</p>
</details> 

**Expected behavior**

The example code sets the Module to 450.0 MHz and switches the transmitter on. With library version 7.1.0 the spectrum analyzer shows the carrier at 450.0 MHz. Compiling with library version 7.1.1 the carrier is at 434.0 Mhz. Additional setFrequency(450) after initialisation do not help.

**Hardware / Software Setup**
 - MCU: Teensy 4.1
 - Wireless module type SX1278 (RA-02)
 - Arduino IDE version 2.3.4
 - Library version 7.1.1 or 7.1.0

CrispyAlice2:
I just encountered this issue as well, it seems like the issue was the module not actually changing the frequency until the LSB frequency register was written. I wrote a quick PR, I'll submit that in a second.

jgromes:
[...]

PR #1369, opened 2025-01-03
Fix frequency issue on SX127x (#1368)

CrispyAlice2:
The SX127x requires that the LSB of the frequency register (Frf) be written in order to change the frequency, but the library was using a SPI writing function that checked for overwrites. If the LSB did not change, the new frequency would be written in successfully, but the frequency would stay at the module's default.

I changed it to use the lower level SPI writing function with no overwrite protection if there was no difference between the old and new contents for the LSB.

I have this commit working on my RFM96W board connected to an Arduino Nano Every, but I have not tested on any other platforms.

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -56,7 +56,7 @@ int16_t Module::SPIgetRegValue(uint32_t reg, uint8_t msb, uint8_t lsb) {
   return(maskedValue);
 }
 
-int16_t Module::SPIsetRegValue(uint32_t reg, uint8_t value, uint8_t msb, uint8_t lsb, uint8_t checkInterval, uint8_t checkMask) {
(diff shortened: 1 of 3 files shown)


jgromes:
Thank you for the contribution! Although, I'm not sure this is the right way to fix this because we may encounter more bugs like this, where the radio is sensitive to register access. I think the better way would be to add a flag to `SPIsetRegValue`, e.g. `bool force = false` that will always write the register if set. Then from `SX127x::setFrequencyRaw` we can set `force=true` to always write it (and perhaps in other places if those come up).

CrispyAlice2:
Do you think it is worth optimizing out the initial register read when force is enabled?

jgromes:
That needs to be kept, since we might only want to change some of the bits. The only point at which we could drop it is if force is enabled and all bits will be modified, but that seems like too much of an edge case now. Let's keep this PR to just fix this one thing ;)

jgromes (approved):

jgromes:
Thank you so much - all looks good, the failed CI run is unrelated and can be ignored.

PR #1387, opened 2025-01-13
fix implicit conversion from float to double warnings

MatejFranceskin:
Converted constants from double to float to avoid compiler "warning: implicit conversion from 'float' to 'double'".

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -58,7 +58,7 @@ int16_t CC1101::transmit(const uint8_t* data, size_t len, uint8_t addr) {
 
 int16_t CC1101::receive(uint8_t* data, size_t len) {
   // calculate timeout (500 ms + 400 full max-length packets at current bit rate)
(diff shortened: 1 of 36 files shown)


jgromes:
Thank you for the contribution - it will take me some time to go through it to double-check all of those are really supposed to be floats. I will also have to fix the conflict that just got introduced by merging #1388 ...

BTW, which warning did you enable to get this? I thought we should have nearly all enabled in the CI, but this was not in there ... thanks!

MatejFranceskin:
The option seems to be -Wdouble-promotion
warning: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Wdouble-promotion]
The way I see it is that you never use doubles so it is just better to have float constants.

jgromes (approved):

jgromes:
Thank you so much - I went through the changes and it all looks good, so merged. The CI fails are unrelated and can be ignored. I will also add the appropriate flag to the CMake.

PR #1388, opened 2025-01-13
[LR11x0] Skip frequency range verification

Genaker:
LR1120 can set the frequency of the documented range. So, allowing another frequency was tested with the LR1121 Lyligo board, which works. It doesn't impact any other code, just an additional feature. Thanks!

## Pull request template
Thank you for taking the time to contribute to RadioLib development!
To keep this library organized, please follow these rules.

1. Make sure the the code in your PR is tested and that you understand all its impacts.
2. Ensure that all CI actions pass - PRs with failed CI will not be merged. CI actions run automatically for every commit pushed to the PR and test the following:
  a. Compilation for Arduino, ESP-IDF and on Raspberry Pi
  b. Runtime test on Raspberry Pi
  c. GitHub CodeQL check
  d. Cppcheck static code scan
3. Follow code style guidelines in [CONTRIBUTING.md](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md)
4. Heads up - all PRs undergo review, during which you may be asked to correct or change some things. The purpose of this review is to keep regressions and bugs at the minimum, and to keep consistent coding style. Please take them as constructive criticism from people who may have a different point-of-view than you do.

After addressing/accepting the points above, delete the contents of this template and replace it with text explaining what is the goal of your PR, why you want to add it to the upstream and what are the foreseen impacts. Once again, thank you for taking the time to contribute!

### DIFF
--- a/src/modules/LR11x0/LR1120.cpp
+++ b/src/modules/LR11x0/LR1120.cpp
@@ -51,13 +51,13 @@ int16_t LR1120::setFrequency(float freq) {
 }
 
 int16_t LR1120::setFrequency(float freq, bool skipCalibration, float band) {
-#if RADIOLIB_CHECK_PARAMS
+  #if RADIOLIB_CHECK_PARAMS
   if(!(((freq >= 150.0) && (freq <= 960.0)) ||
     ((freq >= 1900.0) && (freq <= 2200.0)) ||
     ((freq >= 2400.0) && (freq <= 2500.0)))) {
       return(RADIOLIB_ERR_INVALID_FREQUENCY);
   }
-#endif
+  #endif
 
   // check if we need to recalibrate image
   int16_t state;
--- a/src/modules/LR11x0/LR1120.cpp
+++ b/src/modules/LR11x0/LR1120.cpp
@@ -50,8 +50,8 @@ int16_t LR1120::setFrequency(float freq) {
   return(this->setFrequency(freq, false));
 }
 
-int16_t LR1120::setFrequency(float freq, bool skipCalibration, float band) {
-  if(!(((freq >= 150.0) && (freq <= 960.0)) ||
+int16_t LR1120::setFrequency(float freq, bool skipCalibration, float band, bool skipRfValidation) {
+  if(!skipRfValidation && !(((freq >= 150.0) && (freq <= 960.0)) ||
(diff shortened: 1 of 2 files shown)


Genaker:
Hi, Could you please merge this PR faster? We need this to use this library? Thanks

jgromes:
The allowed frequency range is based on information from the datasheet/user manual, and I think the manufacturer has pretty good reasons why the ranges and defined in this way.

There is already a macro that can be used to disable parameter checking - the proper way to implement skipping frequency range verification is to use that macro in `setFrequency`, not adding new parameters to the  method, so this can't be merged as-is.

https://github.com/jgromes/RadioLib/blob/9a8affde8e0b3d66ff91016dad0da9ef03d41eb3/src/BuildOpt.h#L49-L58

> Could you please merge this PR faster? We need this to use this library?

Is there anything preventing you from using your `patch-3` branch in the meantime?

Also, please don't open PRs with generic names, it will make it hard to track this change in the future.

Genaker:
@jgromes fixed! Please validate. 

As a reference, this implementation: 

https://github.com/jgromes/RadioLib/blob/04b96873a35bbaeb7f41b0dba634df4dbe390449/src/modules/CC1101/CC1101.cpp#L362-L366

jgromes:
Looks good now - the CI fail is unrelated and can be ignored. Merged, thank you for the contribution!

lyusupov:
@jgromes 

> The allowed frequency range is based on information from the datasheet/user manual, and I think the manufacturer has pretty good reasons why the ranges and defined in this way.

Yes, a datasheet defines the best carrier frequency range a radio device can operate with.
However, due to the nature of internal PLL circuits, it is typically allowed to set the PLL beyond the declared range.

**Transmit** feature is typically possible in "out of band" scenario, however **Receiver** does typically have very low sensitivity out of declared bands which makes it almost useless.

One known good example is SX1268 (4XX MHz only) IC.
It actually transmits quite nicely on 868-915 MHz band with 20-22 dBm on RF_OUT pin.
However, it's reported RSSI level is 20 dB less than it has on 433 MHz band. Thus. **receiver** of SX1268 can work on 868-915 MHz in a laboratory but it is useless for real world use case due to  **low sensitivity**.

PR #1404, opened 2025-01-28
[CC1101] FIFO Refills to transmit packets up to 255 bytes

Crsarmv7l:
Hi @jgromes,

Been a while, but finally submitting the CC1101 pull for FIFO refills per our long discussion here:
https://github.com/jgromes/RadioLib/discussions/1138 

You may or may not want to incorporate it as I introduced some blocks in the startTransmit function.

I still get some corrupt packets when using this method (about 1 in 20 randomly) with both an ESP32-S2 and ESP32-S3. A 4.7uf ceramic capacitor certainly seems to help things as the issue is more frequent without the capacitor.

Based on all of the research I have done over the months I have also found open issues on both those chips regarding SPI on the espressif/arduino-esp32 github.

I have an SAMD21 coming to try testing with that


Crsarmv7l:
Ha, some checks fail if I use min some fail if I use std::min. I am using min with arduino IDE

jgromes:
@Crsarmv7l so great to see you are back, and so glad this is finally a PR! I'll do a proper review in the next few days.

> Ha, some checks fail if I use min some fail if I use std::min

The internal `RADIOLIB_MIN` macro should work for everything. Arduino gets confused by `std::min` and non-Arduino stuff doesn't know what `min` is unless we ad `using namespace std;` everywhere.

Crsarmv7l:
> @Crsarmv7l so great to see you are back, and so glad this is finally a PR! I'll do a proper review in the next few days.
> 
> > Ha, some checks fail if I use min some fail if I use std::min
> 
> The internal `RADIOLIB_MIN` macro should work for everything. Arduino gets confused by `std::min` and non-Arduino stuff doesn't know what `min` is unless we ad `using namespace std;` everywhere.

For sure. I'll do the fix for RADIOLIB_MIN along with anything else you find after review.

I know some of the blocks are probably going to be contentious. I could probably also define the MARC states I am checking instead of using magic numbers.

But I'll wait for your feedback.

src/modules/CC1101/CC1101.cpp
```diff
@@ -236,30 +236,66 @@ int16_t CC1101::startTransmit(const uint8_t* data, size_t len, uint8_t addr) {
...
+
+    //If there is room add more data to the FIFO
+    if (fifoBytes < RADIOLIB_CC1101_FIFO_SIZE) {
+        uint8_t bytesToWrite = RADIOLIB_MIN((uint8_t)(RADIOLIB_CC1101_FIFO_SIZE - fifoBytes), (uint8_t)(len - dataSent));
+        SPIwriteRegisterBurst(RADIOLIB_CC1101_REG_FIFO, const_cast<uint8_t*>(&data[dataSent]), bytesToWrite);
+        dataSent += bytesToWrite;
+    }
+  }
+  // Check MARCSTATE for Idle
+  while(SPIgetRegValue(RADIOLIB_CC1101_REG_MARCSTATE, 4, 0) != 0x01) {};
```
jgromes, line 298: Both this as well as the previous do/while loop should have some limit so that they don't become infinite if the radio fails. Usually I do this with a time limit by calling `this->mod->hal->millis()` at the start and checking for a timeout. If the timeout is reached, we can return `RADIOLIB_ERR_TX_TIMEOUT`.

jgromes:
@Crsarmv7l I left some feedback. More generally, I think there are two additional points that should also be addressed:

1. All references to the 64-byte limit on CC1101 throughout the library should be udpated. It's at least in the examples, though maybe some of the doxygen comments in CC1101.h mention it as well.
2. A note to `startTransmit` Doxygen and the interrupt transmit example should be added, explaining that if the packet length exceeds 64 bytes, `startTransmit` will block until only 64 bytes remain in the outgoing packet. I think it is a valid solution given the limitations of CC1101 (since if this is a problem for the user, they can just use short packets only), but it needs to be documented since `CC1101::startTransmit` will now behave slightly differently than the other modules.

src/modules/CC1101/CC1101.cpp
Crsarmv7l: I can certainly put a limit on this one. Going to be busy with work but I will try and get to it on Sun/Mon. I need to see what a reasonable timeout is.

The Do/while loop will fallout if the TXBYTES register is zero because you will at some point get two equal zero reads so we don't have to worry about forever blocking on that one. TBH the double read is accurate per the errata, but we can get by with just a single read of the fifo register in most cases. If you are ok with trading a bit more potential error with one less SPI transaction per loop we can go that direction. In favor of one less read is the argument that you will have less of an underflow risk.

Second thing I have been thinking about. Since I am using SPI to check the MARC state for tx completion, (SPI proved more reliable for refills, I did a bunch of interrupt testing as well) using an interrupt:

```
int16_t state = SPIsetRegValue(RADIOLIB_CC1101_REG_IOCFG2, RADIOLIB_CC1101_GDOX_SYNC_WORD_SENT_OR_PKT_RECEIVED, 5, 0);
  RADIOLIB_ASSERT(state);
```
Doesn't make much sense.   I left it for usage with packets 64 bytes or less to maintain compatibility. I should probably also do something like:

```
if (len <= RADIOLIB_CC1101_FIFO_SIZE) {
    int16_t state = SPIsetRegValue(RADIOLIB_CC1101_REG_IOCFG2, RADIOLIB_CC1101_GDOX_SYNC_WORD_SENT_OR_PKT_RECEIVED, 5, 0);
    RADIOLIB_ASSERT(state);
}
```

Crsarmv7l:
> @Crsarmv7l I left some feedback. More generally, I think there are two additional points that should also be addressed:
> 
> 1. All references to the 64-byte limit on CC1101 throughout the library should be udpated. It's at least in the examples, though maybe some of the doxygen comments in CC1101.h mention it as well.
> 2. A note to `startTransmit` Doxygen and the interrupt transmit example should be added, explaining that if the packet length exceeds 64 bytes, `startTransmit` will block until only 64 bytes remain in the outgoing packet. I think it is a valid solution given the limitations of CC1101 (since if this is a problem for the user, they can just use short packets only), but it needs to be documented since `CC1101::startTransmit` will now behave slightly differently than the other modules.

Aww Documentation is no fun at all 🥲 . I think I can handle those.

src/modules/CC1101/CC1101.cpp
Crsarmv7l: It currently blocks until the transmission is done. My suggested edit for the MARC state read would keep it from blocking on transmissions 64 bytes or less.

That would keep the old functionality while still offering longer (blocking) tx

jgromes: >  I could move checking for Marc state idle to finishTransmit

I think that's the correct approach since that would no longer block once there is nothing more to add to the FIFO and we are just waiting for it to empty out. This is pretty much how other radios behave (wit hthe caveat that their FIFO is 265 bytes, so we can just write the whole packet, exit and wait for some Tx done interrupt to fire).

Crsarmv7l: Sounds like a plan!

Crsarmv7l:
Going to mark as a draft for now.

Had another idea that I want to investigate that would be worth including if beneficial.

Crsarmv7l:
Been looking at a few different things:
1. Checking for PLL lock before tx (probably a good idea for rx too) and recalibrate if not locked with the appropriate delay but like everything it adds more time and another SPI transaction.
2. And the major one I have been looking at is standby(). I believe this is a lost opportunity for faster state changes as idle stops the freq synthesizer, and when started again recalibrates (See Commands table from datasheet). Going to SFSTXON state keeps the crystal running and can go to tx or rx state (See datasheet flow chart)

Thoughts on if those would be desired? I'm thinking rather than do a massive pull I should probably just finish this up then work on those if you want them.
IMG_20250211_151912
IMG_20250211_151708

Crsarmv7l:
Sorry this was the diagram I meant to reference showing paths from SFTXON to both RX and TX

My read is we can simply jump into SFTXON instead of idle, and jump from that to either TX or RX
Screenshot 2025-02-11 8 19 54 PM

Crsarmv7l:
> > I'm thinking rather than do a massive pull I should probably just finish this up then work on those if you want them
> 
> Definitely, it's always easier to move in incremental steps.
> 
> Regarding the standby, I think the only reason it goes to idle mode is for simplicity/compatibility with other radios. But we can certainly go to other modes if there's a specific need to do so. SX126x for example has a `SX126x::standby(uint8_t mode, bool wakeup)` where the users are free to choose either standby with crystal oscillator enabled (which I guess is similar to FSTXON here), or one with only internal RC oscillator enabled, which saves power.

Sure. My reasoning is turning on/off the crystal results in more settle time and calibration. 

This pull is ready to go unless you see something. 

I am playing with avoiding the In/Out buffer creation/copying and SPI paranoid checks for fifo refills, a new function similar to SPIWriteRegBurst, no point filling the in buffer and checking because we can't do anything about it once the bytes fly. But thus far it seems to have minimal impact.

jgromes:
All looks good - merged, thank you so much for this great contribution! And congratulations on getting this completed, job well done!

Crsarmv7l:
> All looks good - merged, thank you so much for this great contribution! And congratulations on getting this completed, job well done!

Thanks. I am not very happy reading TXBYTES twice so you may see one in the coming weeks that uses ISR instead.

PR #1408, opened 2025-01-31
feat: Add liveSpectrumScan

JahazielLem:
Hi!

First of all, thank you for this great library—it's been incredibly useful. While working with the SX126, I needed a more real-time approach for spectrum scanning, so I developed this script and modified it to align with the format of the original example.

This script introduces a software-defined frequency range. If the firmware is configured to scan, for example, from 860 MHz to 920 MHz, you can use this script to focus on a narrower range, such as 900 MHz to 920 MHz. However, note that since the full range is still being captured, you may need to wait for the desired frequencies to be scanned.

I also have a modified version of the firmware that supports real-time frequency changes via commands, making the scanning process much more efficient and works very well with this script.

I hope this script proves useful to others as well. Looking forward to your feedback!

Best regards.

### DIFF
--- a/extras/SX126x_Spectrum_Scan/liveSpectrumScan.py
+++ b/extras/SX126x_Spectrum_Scan/liveSpectrumScan.py
@@ -11,7 +11,6 @@ import sys
 import serial
 import threading
 import argparse
-import platform
 import numpy as np
 import matplotlib.pyplot as plt
 import matplotlib.animation as animation
@@ -22,19 +21,11 @@ FREQ_FRAME_MARK = "FREQ"
 DEFAULT_COLOR_MAP = "BuGn"
 DEFAULT_RSSI_OFFSET = -11
 SCAN_WIDTH = 33
-DEFAULT_START_FREQ = 860
+DEFAULT_START_FREQ = 150
 DEFAULT_STEP_PER_FREQ = 0.2
-DEFAULT_END_FREQ = 928
+DEFAULT_END_FREQ = 960
 DEFAULT_BAUDRATE = 115200
 LIMIT_COUNT = 2
-
-if platform.system() == "Windows":
-    DEFAULT_COMPORT = "COM1"
-elif platform.system() == "Darwin":
-    DEFAULT_COMPORT = "/dev/tty.usbmodem0001"
-else:
-    DEFAULT_COMPORT = "/dev/ttyACM0"
-
 
 def LOG_INFO(message):
     """Function to log information."""
@@ -88,7 +79,6 @@ class SpectrumScan:
             "port",
             type=str,
             help="COM port to connect to the device",
-            default=DEFAULT_COMPORT,
         )
         self.parser.add_argument(
             "-b",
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you for the contribution! This looks very cool, however, when I was trying to run the script with an ESP32 + SX1262, I ran into a problem. When opening the COM port, the ESP32 dumps a bunch of stuff into it, either at a baudrate other than 115200 or just something altogether garbled. However, the script tries to decode that as UTF-8 and fails:

```
bytestream = com.readline().decode("utf-8").strip()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd1 in position 1: invalid continuation byte
```

I'm guessing this is not a problem with the hardware you are using, however, it would be nice for this to be a bit more resilient ;)

JahazielLem:
Hi @jgromes 
Thanks for your comment! I hadn't considered a different baud rate configuration. Your input actually helped me with a serial tool I'm developing, as I hadn't taken that issue into account.

I’ve added warnings and an error message—if you have any other suggestions for improvement, I’d be happy to implement them. For now, I believe this will help in cases where it's necessary to wait for the correct data, such as with the ESP32, which sends its bootloader configuration at startup.

Best regards, and I’ll be on the lookout for any further feedback!

jgromes:
> in cases where it's necessary to wait for the correct data, such as with the ESP32, which sends its bootloader configuration at startup

That was not the issue, the baudrate was consistently set to 115200. The problem is that at the start, the ESP32 just dumps a bunch of data at a different baudrate, or it just toggles some pins that cause bunch opf noise to appear in the terminal. Either way, it cannot be decoded.

I still cannot get this to work as-is, I only get UTF decode errors. The baudrate is the same on both sides and my original scanning script works. So I started looking around the script a bit more and found some more issues:

* Why is there a default COM port value? The argument is positional and cannot be ommitted so the default doesn't make sense
* The default start and and end frequencies defined in the script actually prevent user from setting a different frequency range by commandline arguments. The default is set around 868/915 MHz ISM bands which will prevent scanning of the 434 MHz band. The range check seems altogether redundant.
* The baudrate argument is not used anywhere, so it is actually not possible to change it by argument right now
* Setting end frequency to be exactly equal to start + step size (e.g. 434 and 434.2) will crash the program with:

```
File "C:\Users\user\Documents\Arduino\libraries\RadioLib\extras\SX126x_Spectrum_Scan\liveSpectrumScan.py", line 139, in __data_dissector
    self.data_matrix[:, index] = data
IndexError: index 0 is out of bounds for axis 1 with size 0
```

The only way I was able to kind of get this to work is by opening the Serial port directly in the receive task by calling `with serial.Serial('COM3', 115200, timeout=2) as com:`, which is also what my original script does. Anything else and there was just a bunch of of UTF decode errors. 

Also, should this be run together with SX126x_Spectrum_Scan or SX126x_Spectrum_Scan_Frequency? The helptext says "SX126x_Spectrum_Scan", but doing that, I only get output as shown below, which doesn't seem correct.

out

jgromes:
@JahazielLem is there any update on this? I don't think the latest change addresses my main problem, which is that the new scan version doesn't really seem to be working, at least when testing on Windows. Thanks!

JahazielLem:
Hi @jgromes I'm working on this I found a ESP based board with radio for testing, so I'm working in a dev branch, meanwhile we can put it in draft.

Regards/

jgromes:
Seeing as this has been in draft for 2+ months, I will go ahead and close the PR as this is just a nice-to-have. @JahazielLem feel free to reopen if you start working on this again.

PR #1410, opened 2025-02-01
[LoRaWAN] Add methods to allow user-provided sleep function

jgromes:
Discussed in #1401

cc @StevenCellist - so far I only used this function in the Rx windows, are there others where this would make sense? And should we add this to the reference example?

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -377,6 +377,7 @@ getLastToA	KEYWORD2
 dutyCycleInterval	KEYWORD2
 timeUntilUplink	KEYWORD2
 getMaxPayloadLen	KEYWORD2
+setSleepFunction	KEYWORD2
 
 #######################################
 # Constants (LITERAL1)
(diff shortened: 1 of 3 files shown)


StevenCellist:
We could a-synchronize the transmission calls - at SF9+ they could easily rise above the threshold too. Requires a pre-calculation of the timeout value but this is done anyway on Fixed bands (US915-style) for the Dwell time limits so I guess we can do that for Dynamic (EU868-style) as well.

StevenCellist:
@jgromes I added a commit in the new branch. I know that that is not ideal, but given that I was working on that anyway (and needed asynchronous calls to work around an erroneous cast), I thought I'd add that there. Feel free to port that over if you want to close this first, otherwise we must merge this manually after the mode staging is done.

michapr:
@jgromes Would be nice to add this in reference example - of course with the "warning note" ... ;)

StevenCellist:
@jgromes will you add it in the transmission as well, or shall I do that?

jgromes:
@StevenCellist yes please! And anywhere else you find it appropriate.

StevenCellist:
Turns out every one of the delays could potentially be longer than a few (tens of) milliseconds, so they're now all `sleepDelay()` :)
Time and tests will tell how useful / effective certain thresholds are.

jgromes:
@StevenCellist thank you so much - indeed, we'll see how well this works ;)

PR #1414, opened 2025-02-03
[PHY] Staged modes

jgromes:
Background: When switching between modes (e.g. from standby to receive) there is a significant overhead of commands that must be sent to the device to configure it. Such overhead adds timing uncertainty, because there is a not-insignificant delay between calling the appropriate method (e.g. `startReceive`) and the device actually starting being able to receive packets. The issue becomes more significant at low clock/SPI speeds typically used in battery-powered devices. This causes problems especially for LoRaWAN.

This PR adds support for "staged modes". This is an API of the `PhysicalLayer` class that allow to "stage" a mode to be set by calling `stageMode` method, and then "launching" this mode with `launchMode` that only performs a couple of SPI transactions.

Marking as draft for now since there is still some cleanup left to be done.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -343,6 +343,8 @@ setDataRate	KEYWORD2
 checkDataRate	KEYWORD2
 setModem	KEYWORD2
 getModem	KEYWORD2
+stageMode	KEYWORD2
+launchMode	KEYWORD2
 
 # LoRaWAN
 getBufferNonces	KEYWORD2
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -414,73 +414,6 @@ void LR11x0::clearPacketSentAction() {
   this->clearIrqAction();
 }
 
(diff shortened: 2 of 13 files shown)


HeadBoffin:
Nice - I call this a digital-twin - saves so much time with some of the more complex ADC chips - the drivers just love to read settings, alter a few bits, put it back, do that for 8+ settings and ... yawn ...

tve:
Q: did you consider implementing separate functions for TX and RX? It seems to me that that makes for a nicer and cleaner API. E.g. `stageTransmit`/`stageReceive` and `launchTransmit`/`launchReceive` instead of the current 'stageMode`/`launchMode` with a struct to describe what to actually do? From the code there really doesn't seem to be much that is shared so  IMHO separate functions would make for a cleaner API and a cleaner implementation.

jgromes:
@tve the difference is that the current configuration via a struct allows to launch modes programatically. Also, at the moment only Tx and Rx are implemented, however, it would now be easy to also add standby, LoRa CAD etc. To me the current interface is more generic.

PR #1429, opened 2025-02-19
Increase timeout while waiting for MARC state to be ready to transmit

Crsarmv7l:
Hi Jan,

Apologies, but I noticed today in further testing that while waiting for Marc state to be ready to tx the datasheet timeout can sometimes be triggered (it can take longer than the datasheet says). 

I increased the timeout to 800us and haven't gotten another timeout.


jgromes:
Looks good - merged, thank you for the continued support!

PR #1440, opened 2025-03-02
[MOD] Add debug timestamps

jgromes:
This PR adds a timestamp to the debug output in the format `[seconds.mciroseconds]` since program start. I added this because the default Arduino IDE serial monitor only has milliseconds, needs to be enabled by the user and a lot of people don't use Arduino IDE anyway.

Sample output from LoRaWAN OTAA is shown below.

```
[3466.210338] RLB_PRO: JoinRequest sent (DevNonce = 0) <-- Rx Delay start
[3466.210457] RLB_PRO: 00000000: 00 ee ff c2 b0 11 10 ad 12 26 e1 05 d0 7e d5 b3  .........&...~..
[3466.210511] RLB_PRO: 00000010: 70 00 00 50 d7 57 18                             p..P.W.           
[3466.211804] RLB_PRO: 
[3466.211839] RLB_PRO: PHY:  Frequency = 868.100 MHz, TX = 16 dBm
[3466.213533] RLB_PRO: LoRa: SF = 9, BW = 125.0 kHz, CR = 4/5, IQ: D
[3471.205778] RLB_PRO: Opening Rx1 window (110 ms timeout)... <-- Rx Delay end 
[3471.315959] RLB_PRO: Closing Rx1 window
```

@StevenCellist I tried to catch all the corner cases, but it's possible I missed something, so there might be a timestamp appearing in the middle of the debug output - so just an FYI.


StevenCellist:
Interesting, is this somewhat accurate given that Serial usually isn't the definitive source of truth? It'll definitely open a new universe of bug / optimization hunts... ;)

jgromes:
It's accurate up to the accuracy of the HAL `micros()` timer. So not very accurate on low-end microcontrollers, but should be reasonably accurate on a Raspberry Pi. But more importantly, unlike the Arduino IDE timestamp, it's the timestamp of the debug message being generated (as opposed to it being received at the computer), which will be very different over UART.

StevenCellist:
Ahh, right, that's indeed a much better timestamp. Nice addition, and we will see if there are useful places to improve accuracy.

PR #1444, opened 2025-03-03
[SX128x] fix for invalid GSFK sync words order

### LINKED ISSUE #1443: [SX1280] reverse order of sync words in GFSK frame
## Describe the bug

The issue was mentioned in this thread https://github.com/jgromes/RadioLib/pull/1441#issuecomment-2692874790

It was noticed that RadioLib SX1280 driver transmits GFSK packets with sync words in reverse order which makes it incompatible with the RadioLib driver for LR1121 IC

## To Reproduce

Use this demo sketch for the library
https://github.com/jgromes/RadioLib/blob/master/examples/SX128x/SX128x_GFSK_Modem/SX128x_GFSK_Modem.ino

Take an SDR receiver and capture packets with URH tool

Image

## Expected behavior

The sync words order is expected to follow example sketch

Image

## Additional info

 - MCU: ESP32
 - Wireless module type Ebyte E28 (SX1280)
 - Library version 7.1.2

lyusupov:
Closes #1443

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -996,10 +996,9 @@ int16_t SX128x::setSyncWord(const uint8_t* syncWord, uint8_t len) {
     this->syncWordLen = len;
   }
 
-  // reverse sync word byte order
   uint8_t syncWordBuff[] = { 0x00, 0x00, 0x00, 0x00, 0x00 };
   for(uint8_t i = 0; i < len; i++) {
-    syncWordBuff[4 - i] = syncWord[i];
+    syncWordBuff[i] = syncWord[i];
   }
 
   // update sync word


jgromes:
This is indeed incorrect - I don't know why I did it this way, probably some confusion in the datasheet. Anyway, since the sync word order sent over the SPI matches the order of the `syncWord` array, there is no need for the whole for loop and we can pass the `syncWord` directly to the following `SX128x::writeRegister`. @lyusupov if you could do that, that would be great!

lyusupov:
>  there is no need for the whole for loop and we can pass the `syncWord` directly to the following `SX128x::writeRegister`. @lyusupov if you could do that, that would be great!

I am not sure that it is safe - direct use of`SX128x::writeRegister` may pick garbage from memory if syncword length is less than 5.
I like your 5 bytes buffer method more because it looks like 100% safe.

jgromes:
But we know the sync word length from the user, `len` - we don't have to write all 5 bytes to the register, just the first `len` bytes.

lyusupov:
My opinion that to clear unused sync word registers of SX1280 with 0 value from the buffer is a good safety measure.
You are free to adjust the PR upon your discretion.

jgromes:
OK, let's merge this as is and I will edit it afterwards. I still think that there is no added benefit, as the length of the sync word is controlled separately, so it does not matter what we write (or don't write) to the unused registers.

Merged, thanks!

lyusupov:
Be aware that the build of https://github.com/jgromes/RadioLib/commit/08e0c0e61372e7ccbed430884e92167e37a59635 may fail if no cast:

```
/home/user/Arduino/libraries/RadioLib/src/modules/SX128x/SX128x.cpp:1116:81: error: invalid conversion from 'const uint8_t*' {aka 'const unsigned char*'} to 'uint8_t*' {aka 'unsigned char*'} [-fpermissive]
   int16_t state = SX128x::writeRegister(RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_4, syncWord, len);
                                                                                 ^~~~~~~~
In file included from /home/user/Arduino/libraries/RadioLib/src/modules/SX128x/SX128x.cpp:1:
/home/user/Arduino/libraries/RadioLib/src/modules/SX128x/SX128x.h:907:51: note:   initializing argument 2 of 'int16_t SX128x::writeRegister(uint16_t, uint8_t*, uint8_t)'
     int16_t writeRegister(uint16_t addr, uint8_t* data, uint8_t numBytes);
                                          ~~~~~~~~~^~~~
```

jgromes:
@lyusupov thank you - interesting this slipped by our CI, but should be fixed now.

PR #1446, opened 2025-03-04
[SX128x] fix for improper GFSK syncword setting with length other than 5 bytes

lyusupov:
This is a follow up for https://github.com/jgromes/RadioLib/pull/1444

The PR has been verified with sync word lengths of 2 and 4 bytes.
Expected to work with lengths 1 and 3 bytes as well.

`SX128x::setSyncWord()` method should behave so that LSB of the sync word has to be written into LSB of the `SyncWord1` register.

image

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -996,14 +996,8 @@ int16_t SX128x::setSyncWord(const uint8_t* syncWord, uint8_t len) {
     this->syncWordLen = len;
   }
 
-  uint16_t syncAddr = len == 5 ? RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_4 :
-                      len == 4 ? RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_3 :
-                      len == 3 ? RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_2 :
-                      len == 2 ? RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_1 :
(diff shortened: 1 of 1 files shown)


src/modules/SX128x/SX128x.cpp
```diff
@@ -996,8 +996,14 @@ int16_t SX128x::setSyncWord(const uint8_t* syncWord, uint8_t len) {
     this->syncWordLen = len;
   }
 
+  uint16_t syncAddr = len == 5 ? RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_4 :
```
jgromes, line 999: This is bit too complex (and will make possible future extension into more sync words harder). What this really does is `RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_4 + (5 - len)`.

jgromes (commented):
Thank you! I left one (minor) comment.

src/modules/SX128x/SX128x.cpp
lyusupov: > possible future extension into more sync

friendly reminder that 5 bytes is the maximum limit that SX1280  can do

idea was to make the code independent of actual RADIOLIB_SX128X_REG_SYNC_WORD_1_BYTE_x addresses

I prefer to make the source code readable by humans and to let the modern compilers do the binary code optimization job.

lyusupov: updated with one more commit

jgromes: Great, thank you! Usually I try not to leave too much up to the magic black optimization box - and I'm really not sure those ternary operators would optimize to addition.

jgromes:
All is OK now - merged, thank you!

Issue #1400, opened 2025-01-27
[SX1231 / RF69] invalid order of sync word length setting


lyusupov:
**Describe the bug**

The SX1231 / RF69 module applies sync a word length taken from `this->syncWordLength` variable which has not been updated yet prior to this moment. This results in a wrong (default = 2 byes) setting of the sync word length.

**To Reproduce**

Take a 'settings' demo sketch that comes with the library. 
Alter the code to use a sync word other than 2 bytes in length.

**Expected behavior**

Apply right value of sync word length.

**Screenshots**

![Image](https://github.com/user-attachments/assets/7d5b6223-8b38-44f1-a577-11387e45d8af)

**Additional info:**

 - Library version 7.1.2 (most recent)

jgromes:
Well spotted, thank you! Mentioend above is the commit addressing this, it should be fixed now.

Thanks for reporting!

PR #1448, opened 2025-03-05
[RF69] a fix for invalid sync word length setting

lyusupov:
A range of `len` argument for the `RF69::setSyncWord()` method is &nbsp;  1 ... 8
while the range of `SyncSize` bitmap &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;  is &nbsp;  0 ... 7

image

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -696,7 +696,7 @@ int16_t RF69::setOutputPower(int8_t pwr, bool highPower) {
 
 int16_t RF69::setSyncWord(const uint8_t* syncWord, size_t len, uint8_t maxErrBits) {
   // check constraints
-  if((maxErrBits > 7) || (len > 8)) {
+  if((maxErrBits > 7) || (len == 0) || (len > 8)) {
     return(RADIOLIB_ERR_INVALID_SYNC_WORD);
   }
 
@@ -712,12 +712,12 @@ int16_t RF69::setSyncWord(const uint8_t* syncWord, size_t len, uint8_t maxErrBit
   RADIOLIB_ASSERT(state);
 
   // set the length
-  state = this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_SYNC_CONFIG, len, 5, 3);
+  state = this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_SYNC_CONFIG, len-1, 5, 3);
 
   // set sync word register
   this->mod->SPIwriteRegisterBurst(RADIOLIB_RF69_REG_SYNC_VALUE_1, syncWord, len);
   if(state == RADIOLIB_ERR_NONE) {
-    this->syncWordLength = len;
+    this->syncWordLength = len-1;
   }
 
   return(state);


jgromes:
All looks good - merged, thank you!

jgromes:
@lyusupov I'm a bit confused here - what exactly is needed to fix the RF69 sync word? fc6ff698b9d593d31ee1580f847a15d9ab59bed2 is from #1400, which also implements your suggested fix. So it would be great if we could get this cleaned up. I also don't see much in that commit that could be causing issues.

lyusupov:
1) I realised yesterday that your earlier commit https://github.com/jgromes/RadioLib/commit/fc6ff698b9d593d31ee1580f847a15d9ab59bed2 broke the operation of the Si4432 transmit function in my use case. I suspect that you did not verify your commit on real Si4432 hardware, right? ;

2) this PR makes your commit compliant with the datasheet but not complete enough to fix regression of Si4432 operation for my use case

jgromes:
Could you specify "your use case"? Is it different for SX1231 and RF69? I admit I did not test that specific commit as it seemed a minor change. I will be able to run this with RF69 but not with SX1231 as I don't have the hardware, so if there are some differences they I won't be able to test those.

lyusupov:
> Could you specify "your use case"?

The source code for RF69HCW (a.k.a. SX1231H) is highlighted by this link
https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/driver/radio/radiolib.cpp#L2075-L2815

There are 2 FSK radio protocols supported by the code
- 1 protocol (P3I) uses 2 byte sync word
- 3 others  (FLARM, OGNT, ADS-L) use 8 byte sync word (different values for each protocol)
- some of protocols use 0x55 preamble, some use 0xAA, length varies depending on the protocol
- payload size and format is different for every protocol
- P3I protocol uses 38400 bps bit rate and 9600 kHz deviation
- other protocols use 100 kbps rate, 50 kHz deviation, payload is whitened by Manchester method
 
> Is it different for SX1231 and RF69?

For sure it is not, we all know that RF69 is re-marked SX1231 and RF69H is re-marked SX1231H
HopeRF company is not smart enough to make hardware design their own RF ICs

> I will be able to run this with RF69 but not with SX1231 

I use RF69H @ 868 MHz band. The module is pictured here:

lyusupov:
## Example

[RF69_Transmit_Blocking.ino](https://github.com/jgromes/RadioLib/blob/master/examples/RF69/RF69_Transmit_Blocking/RF69_Transmit_Blocking.ino) demo sketch with default preamble and sync word settings

## before commit https://github.com/jgromes/RadioLib/commit/fc6ff698b9d593d31ee1580f847a15d9ab59bed2

image

## after commit https://github.com/jgromes/RadioLib/commit/fc6ff698b9d593d31ee1580f847a15d9ab59bed2

image


## after commit https://github.com/jgromes/RadioLib/commit/6bf0c3f7357518e3c00c9b88b158872b8ef91469

image

Velocet:
Why not have a look at the app for the SX123x? It is written in C# and [dotPeek](https://www.jetbrains.com/decompiler/) does a great job at decompilation.
This did in fact help me find the changes made by one of the modes. Maybe it can also help you.

PR #1450, opened 2025-03-06
Fix for CN470 (formerly CN500) (#1447)

StevenCellist:
As per #1447, the naming scheme of CN500 was a pretty odd choice and is now renamed to CN470 which is much more commonly recognized.
Also, the JoinRequest datarate is modified from DR0 to DR3. There is no standard for the JoinRequest datarate specified, but the way RadioLib handles fixed bands (such as CN470) requires one to be pre-set. I've elected to use the DR3 as it's in the middle of the allowed range (DR0-DR5), but as DR2.5 does not exist, the more neighbourhood-friendly DR3 is the choice I've settled on.


jgromes:
All looks good, merged - honestly not sure why I named it CN500 in the first place ...

jgromes:
Probably, somewhere in the early days of the LoRaWAN implementation here - feel free to poke around git blame ;)

StevenCellist:
That's where I took that link from indeed ;)

Issue #1430, opened 2025-02-19
[Si4432] failure to receive a fixed size FSK packet


lyusupov:
## **Describe the bug**

I transmit a fixed size FSK modulated packet from SX1276 radio by the demo sketch 
https://github.com/jgromes/RadioLib/blob/master/examples/SX127x/SX127x_FSK_Modem/SX127x_FSK_Modem.ino

The packet size is fixed to 8 bytes **on both sides** (Tx and Rx) by doing this in both Tx and Rx sketches:

```
  state = radio.fixedPacketLengthMode(8);
```

However, the data received from Si4432 FIFO is 0x01 0x01 0x01  0x01  0x01  0x01  0x01 0x01 garbage.

See the SPI log below

```
RLB_SPI: R      7       1
RLB_SPI: W      7       1
RLB_SPI: R      7       1
RLB_SPI: R      8       0
RLB_SPI: W      8       2
RLB_SPI: R      8       2
RLB_SPI: R      8       2
RLB_SPI: W      8       0
RLB_SPI: R      8       0
RLB_SPI: R      3       20      12
RLB_SPI: W      5       3
RLB_SPI: W      6       0
RLB_SPI: W      7       5
RLB_SPI: R      3E      8
RLB_SPI: R      3       21      32
RLB_SPI: R      7F      1       1       1       1       1       1       1      1
RLB_SPI: R      7       5
RLB_SPI: W      7       1
RLB_SPI: R      7       1
RLB_SPI: R      3       A0      12
[Si4432] Received packet!
[Si4432] Data: 11111111
```

## **To Reproduce**

Use https://github.com/jgromes/RadioLib/blob/master/examples/SX127x/SX127x_FSK_Modem/SX127x_FSK_Modem.ino sketch to transmit and https://github.com/jgromes/RadioLib/tree/master/examples/Si443x/Si443x_Receive_Interrupt to receive.

## **Expected behavior**

Displayed payload (data received) is expected to be 0x01 0x23  0x45 0x67 0x89 0xab 0xcd 0xef

## **Screenshots**

URH tool can confirm that Tx signal is fine. 
It has
- 16 bits of 0x5555 preamble
- 0x12AD sync word
- 0x0123456789abcdef of payload data

![Image](https://github.com/user-attachments/assets/2a291850-973b-4ff9-9b8d-4deb7d685d74)

## **Additional info:**
 - MCU: ESP32
 - Wireless module type SX1276 (TX) and Si4432 (RX)
[...]

PR #1452, opened 2025-03-08
[RF69] further fix for sync word length

lyusupov:
Follow up for PR https://github.com/jgromes/RadioLib/pull/1448

It fixes regression happened after commit https://github.com/jgromes/RadioLib/commit/fc6ff698b9d593d31ee1580f847a15d9ab59bed2

SPI log before the fix
```

RLB_SPI: R      2E      98
RLB_SPI: W      2E      80
RLB_SPI: R      2E      80
RLB_SPI: W      2F      2D      1
```

SPI log after the fix

```
RLB_SPI: R      2E      98
RLB_SPI: W      2E      88
RLB_SPI: R      2E      88
RLB_SPI: W      2F      2D      1
```

### DIFF
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -712,7 +712,7 @@ int16_t RF69::setSyncWord(const uint8_t* syncWord, size_t len, uint8_t maxErrBit
   RADIOLIB_ASSERT(state);
 
   // set the length
-  state = this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_SYNC_CONFIG, len-1, 5, 3);
+  state = this->mod->SPIsetRegValue(RADIOLIB_RF69_REG_SYNC_CONFIG, (len-1)<<3, 5, 3);
 
   // set sync word register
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you, I was able to verify with an SDR that this indeed seems to fix the issue. I can see the sync word 0x12ad with the fix applied, but only 0x12 without it.

I think that should resolve these RF69 sync word issues, right?

lyusupov:
> I think that should resolve these RF69 sync word issues, right?

yes, this fix is the final in the series of 3 actions
- len -1
- maxErrBits
- (len-1) << 3

that fixes regression after https://github.com/jgromes/RadioLib/commit/fc6ff698b9d593d31ee1580f847a15d9ab59bed2

[My use case](https://github.com/jgromes/RadioLib/pull/1448#issuecomment-2704880643) works fine after applying this PR.

jgromes:
OK, glad to have this sorted - merged, thank you!

lyusupov:
> I was able to verify with an SDR

if you have some instruments readily available - could you, please, take a look at **FIFO** https://github.com/jgromes/RadioLib/issues/1430 issue with Si443x ?

jgromes:
@lyusupov it's in my todo list, but unfortunately many things are these days.

Issue #1482, opened 2025-04-15
CC1101 Module: field length in variable length mode with address filtering enabled should be +1 byte length


jcrespoc:
When using variable length packet mode and address filtering, the Length byte sent must be increased by one.

Failure to do this results in the last payload byte not being transmitted, and the library, upon reception, delivers a corrupted last byte, erroneous RSSI and LQI values, and arbitrary CRC verification.

<details><summary>Corrupted payload in variable length mode with address filtering enabled</summary>
<p>

```
[CC1101] Setting BitRate ... success!
[CC1101] Setting node address (0x02)... success!
[CC1101] Setting variable packet length mode ... success!
=========== After Setup Start ============
INTERNAL Memory Info:
------------------------------------------
  Total Size        :   389068 B ( 379.9 KB)
  Free Bytes        :   350676 B ( 342.5 KB)
  Allocated Bytes   :    33152 B (  32.4 KB)
  Minimum Free Bytes:   347208 B ( 339.1 KB)
  Largest Free Block:   278516 B ( 272.0 KB)
------------------------------------------
GPIO Info:
------------------------------------------
  GPIO : BUS_TYPE[bus/unit][chan]
  --------------------------------------  
     9 : GPIO
    10 : GPIO
    11 : SPI_MASTER_MOSI[0]
    12 : SPI_MASTER_SCK[0]
    13 : SPI_MASTER_MISO[0]
    19 : USB_DM
    20 : USB_DP
    43 : UART_TX[0]
    44 : UART_RX[0]
    46 : GPIO
============ After Setup End =============
[CC1101] Waiting for incoming transmission ... failed, code -6
[CC1101] Waiting for incoming transmission ... success!
received data:
0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0x3d, // .#Eg...=
expected data:
0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef, // .#Eg....
[CC1101] RSSI:          -117.00 dBm
[CC1101] LQI:           42
```

</p>
</details> 

**To Reproduce**
This is a slightly modified version of the CC1101_Receive_Address.ino demo.

```cpp
/*
  RadioLib CC1101 Receive with Address Example

  This example receives packets using CC1101 FSK radio
  module. Packets can have 1-byte address of the
  destination node. After setting node address, this node
  will automatically filter out any packets that do not
  contain either node address or broadcast addresses.

[...]

PR #1483, opened 2025-04-15
fix for #1482

jcrespoc:
## Pull request template
Thank you for taking the time to contribute to RadioLib development!
To keep this library organized, please follow these rules.

1. Make sure the the code in your PR is tested and that you understand all its impacts.
2. Ensure that all CI actions pass - PRs with failed CI will not be merged. CI actions run automatically for every commit pushed to the PR and test the following:
  a. Compilation for Arduino, ESP-IDF and on Raspberry Pi
  b. Runtime test on Raspberry Pi
  c. GitHub CodeQL check
  d. Cppcheck static code scan
3. Follow code style guidelines in [CONTRIBUTING.md](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md)
4. Heads up - all PRs undergo review, during which you may be asked to correct or change some things. The purpose of this review is to keep regressions and bugs at the minimum, and to keep consistent coding style. Please take them as constructive criticism from people who may have a different point-of-view than you do.

After addressing/accepting the points above, delete the contents of this template and replace it with text explaining what is the goal of your PR, why you want to add it to the upstream and what are the foreseen impacts. Once again, thank you for taking the time to contribute!

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -262,17 +262,19 @@ int16_t CC1101::startTransmit(const uint8_t* data, size_t len, uint8_t addr) {
   // data put on FIFO
   uint8_t dataSent = 0;
 
+  uint8_t filter = SPIgetRegValue(RADIOLIB_CC1101_REG_PKTCTRL1, 1, 0);
+
   // optionally write packet length
   if(this->packetLengthConfig == RADIOLIB_CC1101_LENGTH_CONFIG_VARIABLE) {
     if (len > RADIOLIB_CC1101_MAX_PACKET_LENGTH - 1) {
         return(RADIOLIB_ERR_PACKET_TOO_LONG);
     }
-    SPIwriteRegister(RADIOLIB_CC1101_REG_FIFO, len);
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you for the contribution! I tried to find in the CC1101 data sheet where exactly is it specified that the length value written into the FIFO should be incremented for the address field. But I guess it does make sense since there is no other way that would tell the receiving way the address field is present.

Merged, thank you!

jcrespoc:
> Thank you for the contribution! I tried to find in the CC1101 data sheet where exactly is it specified that the length value written into the FIFO should be incremented for the address field. But I guess it does make sense since there is no other way that would tell the receiving way the address field is present.
> 
> Merged, thank you!

I couldn't find references to this in the datasheet. 
I reviewed examples provided by TI (Chipcon) but none of them use address filtering.
In reality, the chip expects an address in the second byte whenever address filtering is activated. It shouldn't count for the length, but after a week of going crazy with my code (interrupt driven support for packets 65 to 255 bytes long) I decided to do a simpler test and that was the result. 

Thanks to you.

PR #1504, opened 2025-05-14
[LoRaWAN] Implement Class C (unicast)

StevenCellist:
As per title. 

Tested on LoRaWAN v1.0.4 (OTAA / EU868) and LoRaWAN v1.1 (OTAA / EU868).  
Implementation as per spec except for one item in the 1.1 specification: 
> The device SHALL switch mode as soon as the first `DeviceModeInd` command is transmitted.

Instead, it switches once `DeviceModeConf` is received. This tackles the note that is included that suggests using some timeout in case there is no response from the network.

No example included yet, but it's pretty simple:
```cpp
void setup() {
  ...
  // join
  ...
  node.setClass(RADIOLIB_LORAWAN_CLASS_C);
  // send a confirmed uplink (assuming it receives a downlink at once) as per TS001 par. 6.2.7
  // this way, the LNS knows that the device has an active session (received the JoinAccept)
  node.sendReceive(payload, payLen, 1, payload, &payLen, true, &eventUp, &eventDown);
  ...
}

void loop() {
  state = node.getDownlinkClassC(payload, &payLen, &eventDown);
  if(state > 0) {
    Serial.println(F("Received a Class C downlink!"));
    // Did we get a downlink with data for us
    if(payLen > 0) {
      Serial.println(F("Downlink data: "));
      arrayDump(payload, payLen);
    }

    // print extra information about the event
    Serial.println(F("[LoRaWAN] Event information:"));
    Serial.print(F("[LoRaWAN] Confirmed:\t"));
    Serial.println(eventDown.confirmed);
    Serial.print(F("[LoRaWAN] Confirming:\t"));
    Serial.println(eventDown.confirming);
    Serial.print(F("[LoRaWAN] Datarate:\t"));
    Serial.println(eventDown.datarate);
    Serial.print(F("[LoRaWAN] Frequency:\t"));
    Serial.print(eventDown.freq, 3);
    Serial.println(F(" MHz"));
    Serial.print(F("[LoRaWAN] Frame count:\t"));
    Serial.println(eventDown.fCnt);
    Serial.print(F("[LoRaWAN] Port:\t\t"));
    Serial.println(eventDown.fPort);
    Serial.print(F("[LoRaWAN] Time-on-air: \t"));
    Serial.print(node.getLastToA());
    Serial.println(F(" ms"));
    Serial.print(F("[LoRaWAN] Rx window: \t"));
    Serial.println(state);
  }
}
```

This 'should' be called continuously in the loop - it doesn't really need to, you can surely delay for seconds, but then a subsequent downlink may already have arrived and overwritten the previous one...

Feedback on higher and lower levels of implementation welcome. Marked as draft to discourage actual use, testing only.

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -1977,6 +1977,11 @@ int16_t LR11x0::stageMode(RadioModeType_t mode, RadioModeConfig_t* cfg) {
         state = setPacketParamsLoRa(this->preambleLengthLoRa, this->headerType, this->implicitLen, this->crcTypeLoRa, this->invertIQEnabled);
         RADIOLIB_ASSERT(state);
       }
+
+      // if max(uint32_t) is used, revert to RxContinuous
+      if(cfg->receive.timeout == 0xFFFFFFFF) {
+        cfg->receive.timeout = 0xFFFFFF;
+      }
       this->rxTimeout = cfg->receive.timeout;
     } break;
   
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -1554,6 +1554,10 @@ int16_t SX126x::stageMode(RadioModeType_t mode, RadioModeConfig_t* cfg) {
       state = startReceiveCommon(cfg->receive.timeout, cfg->receive.irqFlags, cfg->receive.irqMask);
       RADIOLIB_ASSERT(state);
 
+      // if max(uint32_t) is used, revert to RxContinuous
+      if(cfg->receive.timeout == 0xFFFFFFFF) {
+        cfg->receive.timeout = 0xFFFFFF;
+      }
       this->rxTimeout = cfg->receive.timeout;
     } break;
   
--- a/src/modules/SX127x/SX127x.cpp
+++ b/src/modules/SX127x/SX127x.cpp
@@ -1662,6 +1662,10 @@ int16_t SX127x::stageMode(RadioModeType_t mode, RadioModeConfig_t* cfg) {
 
       int16_t modem = getActiveModem();
       if(modem == RADIOLIB_SX127X_LORA) {
+        // if max(uint32_t) is used, revert to RxContinuous
+        if(cfg->receive.timeout == 0xFFFFFFFF) {
+          cfg->receive.timeout = 0;
+        }
         if(cfg->receive.timeout != 0) {
           // for non-zero timeout value, change mode to Rx single and set the timeout
           this->rxMode = RADIOLIB_SX127X_RXSINGLE;
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -1426,6 +1426,10 @@ int16_t SX128x::stageMode(RadioModeType_t mode, RadioModeConfig_t* cfg) {
         state = setPacketParamsLoRa(this->preambleLengthLoRa, this->headerType, this->payloadLen, this->crcLoRa, this->invertIQEnabled);
         RADIOLIB_ASSERT(state);
       }
+      // if max(uint32_t) is used, revert to RxContinuous
+      if(cfg->receive.timeout == 0xFFFFFFFF) {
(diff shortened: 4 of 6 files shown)


StevenCellist:
One additional comment: this PR invalidates existing Nonces (and Session) buffers, as LoRaWAN Class does not make a device unique. As such, the Class has been moved from the Nonces to the Session buffer.

StevenCellist:
Multicast now also available using Class C. Multicast does not necessarily require the node to already be in Class C - it could be on top of normal Class A. But it can also be both at the same time.

To register a Multicast group on TTN, you can register a new (virtual) device on the console and set the Activation method to `ABP & Multicast`. Or use the CLI:
```ps
$APP_ID="your-application-id" 
$DEVICE_ID="new-device-id"
$FREQUENCY_PLAN="EU_863_870_TTN"
$LORAWAN_VERSION="MAC_V1_0_4"
$LORAWAN_PHY_VERSION="RP002_V1_0_4"
$MULTICAST_GROUP_ADDRESS="260B0000"
$MULTICAST_SESSION_DATARATE="DATA_RATE_3"
$MULTICAST_SESSION_FREQUENCY="869525000"
$MULTICAST_APP_SESSION_KEY="CE771570000000000000000000000003"
$MULTICAST_NWK_SESSION_KEY="CE771570000000000000000000000004"

ttn-lw-cli end-devices create $APP_ID $DEVICE_ID \
           --frequency-plan-id $FREQUENCY_PLAN \
           --lorawan-version $LORAWAN_VERSION \
           --lorawan-phy-version $LORAWAN_PHY_VERSION \
           --session.dev-addr $MULTICAST_GROUP_ADDRESS \
           --session.keys.app-s-key.key $MULTICAST_APP_SESSION_KEY \
           --session.keys.nwk-s-key.key $MULTICAST_NWK_SESSION_KEY \
           --multicast \
           --mac-settings.rx2-data-rate-index $MULTICAST_SESSION_DATARATE \
           --mac-settings.rx2-frequency $MULTICAST_SESSION_FREQUENCY \
           --supports-class-c
```

In this example, the default Rx2 frequency and datarate are used to test simultaneous decryption of Multicast-C and Unicast-C downlinks. But you can configure these differently and subsequently pass this information to the `startMulticastSession()` function using additional arguments.

```cpp
void setup() {
  ...
  // join
  // possibly already start Class C + send a confirmed uplink as before
  ...
  node.startMulticastSession(RADIOLIB_LORAWAN_CLASS_C, 0x260B0000, mcAppKey, mcNwkKey);
  ...
}

void loop() {
  state = node.getDownlinkClassC(payload, &payLen, &eventDown);
  if(state > 0) {
    Serial.println(F("Received a Class C downlink!"));
    // Did we get a downlink with data for us
    if(payLen > 0) {
      Serial.println(F("Downlink data: "));
      arrayDump(payload, payLen);
    }

    // print extra information about the event
    Serial.println(F("[LoRaWAN] Event information:"));
    Serial.print(F("[LoRaWAN] Rx window: \t"));
    Serial.println(state);
    Serial.print(F("[LoRaWAN] Cast:\t\t"));
    Serial.println(eventDown.multicast ? "Multi" : "Uni");
  }
}
```

To schedule a Multicast downlink on TTS, the Console cannot be used; you must do this through MQTT (or webhook). Configure MQTT through the Application's "Other integrations", and get the username and password. Set these as MQTT Basic Auth, set MQTT v3.1.1, host `mqtt://eu1.cloud.thethings.network:1883`. Then publish to `v3/{mqtt-username}/devices/{device-id}/down/replace`:
```json
{
   "downlinks" : [
      {
         "class_b_c" : {
            "gateways" : [
               {
                  "gateway_ids" : {
                     "gateway_id" : "your-ttn-gateway-id"
                  }
               }
            ]
         },
         "f_port" : 42,
         "frm_payload" : "zncVcA==",
         "priority" : "NORMAL"
      }
   ]
}
```
You should see a Multicast downlink with the payload content `CE771570`.

jgromes (commented):
@StevenCellist I left some comments, should be fairly easy to implement. Thank you!

StevenCellist:
Bloody dynamic allocation... 😅

PR #1521, opened 2025-06-10
Fix ESP32 heap corruption in POCSAG transmission

brooksbUWO:
# Fix ESP32 heap corruption in POCSAG transmission

## Problem
ESP32 crashes with heap corruption when transmitting specific POCSAG ASCII message lengths. The issue occurs with 42-byte escaped messages while 52-byte messages work fine.

**Error**: `assert failed: block_trim_free heap_tlsf.c:371 (block_is_free(block) && "block must be free")`

## Root Cause
Dynamic allocation in `Pager.cpp` line 119 creates different heap allocation patterns for different message lengths, triggering corruption on smaller allocations:
```cpp
uint32_t* msg = new uint32_t[msgLen];
```

## Solution
Replace dynamic allocation with static buffer to eliminate heap corruption entirely.

## Changes
- Replace `new uint32_t[msgLen]` with static buffer in Pager.cpp
- Remove corresponding `delete[] msg`
- Add bounds checking with `RADIOLIB_ERR_PACKET_TOO_LONG`
- 512-word buffer supports messages up to ~16 batches (extremely large)

## Testing
Reproduction code that crashes before fix and works after:

```cpp
#include <RadioLib.h>

SX1276 radio = new Module(18, 26, 23, 33);
PagerClient pager(&radio);

void setup() {
  Serial.begin(115200);
  radio.beginFSK(462.45, 0.512, 4.5, 12.5, 17, 0, false);
  pager.begin(462.45, 512);
}

void loop() {
  uint32_t address = 800828;
  
  // Working: 52-byte escaped message
  uint8_t longEscaped[] = "~00~00~00~00~00~01FF00~02WA202205051827524~03039a~04";
  
  // Crashing: 42-byte escaped message (before fix)
  uint8_t shortEscaped[] = "~00~00~00~00~00~01FF00~02WS81001~030187~04";
  
  Serial.println("52-byte message:");
  int result = pager.transmit(longEscaped, strlen((char*)longEscaped), address, RADIOLIB_PAGER_ASCII);
  Serial.println(result);
  delay(3000);
  
  Serial.println("42-byte message:");
  result = pager.transmit(shortEscaped, strlen((char*)shortEscaped), address, RADIOLIB_PAGER_ASCII);
  Serial.println(result);
  delay(3000);
}
```

**Platform**: ESP32 with any RadioLib-compatible radio module  
**Impact**: No functional changes, only improved memory safety

### DIFF
--- a/src/protocols/Pager/Pager.cpp
+++ b/src/protocols/Pager/Pager.cpp
@@ -136,7 +136,12 @@ int16_t PagerClient::transmit(const uint8_t* data, size_t len, uint32_t addr, ui
   #if RADIOLIB_STATIC_ONLY
     uint32_t msg[RADIOLIB_STATIC_ARRAY_SIZE];
   #else
-    uint32_t* msg = new uint32_t[msgLen];
+    #define RADIOLIB_PAGER_MAX_MSG_LEN 512
+    static uint32_t staticMsgBuffer[RADIOLIB_PAGER_MAX_MSG_LEN];
+    if(msgLen > RADIOLIB_PAGER_MAX_MSG_LEN) {
+      return(RADIOLIB_ERR_PACKET_TOO_LONG);
+    }
+    uint32_t* msg = staticMsgBuffer;
   #endif
 
   // build the message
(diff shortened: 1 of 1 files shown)


jgromes:
> Dynamic allocation in Pager.cpp line 119 creates different heap allocation patterns for different message lengths, triggering corruption on smaller allocations:

Dynamic allocation in itself should not cause issues, what can be problematic is when the calculated length of that allocation is incorrect and later operations take place after the end of the buffer. So this should probably address the actual root cause - is there such a mismatch? What could also be added is that the dynamic allocation actually suceeded (and did not return a null pointer). 

> Impact: No functional changes

I disagree - the proposed fix has a `static` buffer, which will be kept even after `PagerClient::transmit` exits.

Issue #1526, opened 2025-06-18
Channel Activity Detection


christianh104:
1. Channel Activity Detection (CAD) is not supported for spreading factors 5 and 6 despite being supported by the SX1261/SX1262. There's no warning in the documentation about this which is confusing for the user when CAD simply doesn't work. In addition, SX126x::setCad() writes over the spreadingFactor property without actually changing the spreading factor on the device, which screws up timeout and airtime calculations.
2. A comment in SX126x::setCad() states "default CAD parameters are shown in Semtech AN1200.48, page 41" which confuses the meaning and context of the table referenced. That table is titled "Best Settings Using 2 Symbols", is in a section only applicable for 125kHz bandwidth, and is meant for minimal energy consumption and minimal false positives. Additionally, 2 symbols for detection is suboptimal for most spreading factors according to the table on AN1200.48, page 15. The document also states for spreading factors 9-12, "4 symbols are necessary to have a good performance without false detection".

I would be happy to write a PR to take care of these issues, though in AN1200.48 there are only "best settings" published for 125kHz and 500kHz. DS.SX1261-2.W.APP, page 92 specifies recommended settings for all spreading factors with a consistent 4 symbols, and no variation by bandwidth. Also, the calculation for cadDetPeak is a simple spreadingFactor + 13 which avoids hard-coded parameters. These settings appear to bring slightly more false positives but also detect more real packets at lower RSSI values. I believe it makes sense for the default parameters to target the best detection rate and let those who want minimal power consumption/false positives tinker with the parameters. Any thoughts?

jgromes:
> Channel Activity Detection (CAD) is not supported for spreading factors 5 and 6

What leads you to believe this is the case?

> SX126x::setCad() writes over the spreadingFactor property

This is indeed incorrect, we should use a local variable for addressing the values in `detPeakValues` array.

Regarding your second point about the default settings, I'm open to reworking it - but in any case, `setCad` and associated methods expose all of the configurable CAD parameters, so the user always has the flexibility to use non-default values to fine tune for their specific use case.

christianh104:
Not sure why I thought CAD wouldn't work for SF5/6, I believe that was an initial assumption made when I saw the spreadingFactor clamped to 7 that stuck in my head. I have tested it and it does work. I'll go ahead with a PR if for no other reason than to make it more clear that the spreading factor is not actually being modified by setCad().

jgromes:
Resolved by merging #1527

PR #1527, opened 2025-06-19
Optimize default CAD parameters

christianh104:
This PR addresses #1526 by optimizing default CAD parameters according to DS.SX1261-2.W.APP rev. 1.1, page 92. This should improve detection rate at lower RSSI values, reduce false positives, and improve detection performance for SF5 and SF6 in particular. In addition, hard-coded values are eliminated for cadDetPeak.


jgromes:
All looks good thank you for the contribution!

PR #1532, opened 2025-06-21
[CI] Add uncrustify script and action

jgromes:
This PR adds script to perform uncrustify formatting check and fix. Result of the script is a `.patch` file which can be applied to fix the issues.

A GitHub action is also added to trigger this on every push to master and every PR.


.github/workflows/uncrustify.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Install uncrustify
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y uncrustify
```
Dazza0, line 22: @jgromes Might be a good idea to install a specific version of uncrustify here, as version mismatch between the `uncrustify.cfg` and the one installed on the runner is what is causing the following error:

```
 Option<UNUM>: at /home/runner/work/RadioLib/RadioLib/extras/uncrustify/uncrustify.cfg:2838: Expected unsigned number , for 'align_nl_cont'; got 'false'
```

Latest uncrustify [expects an integer](https://jeayne.github.io/Uncrustify-Documentation/index_name.html) for the `align_nl_cont` option. However, it seems like this option used to accept a boolean (at least up till version [0.76.0](https://github.com/uncrustify/uncrustify/blob/5e1d2eccffae9ef5ddc6311dc01b9647af620994/etc/mono.cfg#L68)).

jgromes: @Dazza0 indeed, I will need to fix this to a specific version. Though this CI fail is due to the fact that [uncrustify found some issues](https://github.com/jgromes/RadioLib/actions/runs/16125371765/job/45500896401?pr=1532#step:4:35981), as I have yet to set the configuration to correctly cover all the corner cases. So it is working exactly as "expected".

jgromes: Done in https://github.com/jgromes/RadioLib/pull/1532/commits/d2c876a74314e2f0a0712d5ec4bd2f88d5ba92e0

jgromes:
So after some trial and error I managed to get an uncrustify configuration that somewhat matches the existing code style. It did  find some inconsistencies, whitespaces at line ends etc., but overall I'm still not completely sure about this. I see the following problems:

1. It will require contributors to run the check, and do it with a specific uncrustify version. On Linux that's reasonably easy with the script in `extras/uncrustify/uncrustify.sh`, but I don't know how many contributors here use other OSes. Also, a lot of people contribute very simple fixes (simple meaning 5 changed lines or less) - especially when getting started. This automated check makes it somewhat harder and increases the barrier to entry.
2. The script does produce a `diff` file that can be applied with `git diff`, so technically it's not necessary to run uncrustify locally, however, it requires going to the build artifacts, downloading the diff file from there and applying it manually. Again though, that requires a bit more knowledge, so it has the potential of turning new contributors away.

What do you think @StevenCellist?

StevenCellist:
I'd very, _very_ much dislike having to do this. I'm a Windows user most of the time and it sounds terrible to have to do that every time. Let alone for those users that only want to commit a missing state assert and just learned to file an issue.

My personal preference would be to have it e.g. run as part of the release CI.

StevenCellist:
It's killing some alignment choices in places where we added some extra whitespace. Which is a minor inconvenience. And probably have it ignore the firmware images?

jgromes:
> My personal preference would be to have it e.g. run as part of the release CI.

That is indeed a solution, we can run this on release (and also occasionaly by manual trigger).

> It's killing some alignment choices in places where we added some extra whitespace

Could you point out where exactly? Like I said I wasn't able to fully match the existing style

> And probably have it ignore the firmware images?

I don't expect any changes there, so I actually don't really care how the firmware-only headers are formatted. Also there's no readability required there.

StevenCellist (commented):
Marked a couple of yours and one of mine as an example.

bitandpin:
FYI, there are discrepancies between uncrustify settings and whats stated in CONTRIBUTING.md.

eg `if(!pwmPin) {` vs `if (!pwmPin) {`

PR #1536, opened 2025-06-23
Use sdkconfig.defaults for ESP-IDF example

Dazza0:
This PR updates the ESP-IDF example so that it uses a `sdkconfig.defaults` file instead of `sdkconfig` (see [docs](https://docs.espressif.com/projects/esp-idf/en/v5.4.1/esp32/api-reference/kconfig.html#using-sdkconfig-defaults)). The main benefits of this are:

- Only needing to provide preset values for options that are relevant to the project, thus making it easier to maintain.
- No longer have unstaged git changes when building against newer versions of, as ESP-IDF will overwrite the existing `sdkconfig` file to fill in any missing config options or replace any options using deprecated names.

Changes:

- Ran `idf.py save-defconfig` on existing `sdkconfig` file to get options with non-default values. Saved relevant ones to `sdkconfig.defaults` (only found `CONFIG_FREERTOS_HZ` to be of relevance, see list below)
- Removed existing `sdkconfig`
- Added `sdkconfig` to `.gitignore` as ESP-IDF will automaticcally generate `sdkconfig` on each build

List of non-default config values found by `idf.py save-defconfig` against ESP-IDF `v5.4.2`

```
CONFIG_APP_RETRIEVE_LEN_ELF_SHA=16
CONFIG_ESP_PHY_REDUCE_TX_POWER=y
CONFIG_ESP_WIFI_GMAC_SUPPORT=n
CONFIG_FREERTOS_HZ=1000
CONFIG_LOG_COLORS=y
CONFIG_LWIP_TCP_SND_BUF_DEFAULT=5744
CONFIG_LWIP_TCP_WND_DEFAULT=5744
CONFIG_LWIP_HOOK_IP6_INPUT_NONE=y
```


jgromes:
My lack of experience with ESP-IDF shows, because I was not aware of this mechanism and it was always quite annoying to see a bunch of re-generated stuff in sdkconfig, so this is very useful indeed.

PR #1542, opened 2025-06-26
[LR11X0] Add missing description to 'offset' parameter in 'getPacketLength()'

Dazza0:
# Description

Adds missing Doxygen description for the `offset` parameter of `getPacketLength()` in the `LR11x0` class


jgromes:
Looking good - merged, thank you for the fix!

Issue #1558, opened 2025-07-17
Pico HAL interrupts trigger on both edges


koisto:
I've just started using the library to interface a RPi Pico to an RFM96W module. Thank you for providing such a great library, it has enabled me to make a quick start to my project. 

I have ported the interrupt driven receive example SX126x_Receive_Interrupt.ino over to the RPi Pico and I have observed that I am getting two receive interrupts for each packet received. 

Having investigated this issue, I believe that the problem lies in the HAL for the RPi PICO.

```
  void attachInterrupt(uint32_t interruptNum, void (*interruptCb)(void), uint32_t mode) override {
    if (interruptNum == RADIOLIB_NC) {
      return;
    }

    gpio_set_irq_enabled_with_callback(interruptNum, GPIO_IRQ_EDGE_RISE | GPIO_IRQ_EDGE_FALL, true, (gpio_irq_callback_t)interruptCb);
  }
```

Here the `mode` argument is not used and `gpio_set_irq_enabled_with_callback` function gets called to set the GPIO IRQ to trigger off of both the rising and falling edge. 

**Expected behavior**
Only one interrupt per received packet.

**Additional info :**
 - MCU: Raspberry Pi Pico (RP2040)
 - Wireless module type: RFM96W
 - Library version: 7.2.1

jgromes:
This is indeed incorrect - sicne the values for rising and falling edge have been already passed to `RadioLibHal` in the `PicoHal` constructor, it should be possible to just replace `GPIO_IRQ_EDGE_RISE | GPIO_IRQ_EDGE_FALL` with `mode` and it should work correctly - could you try it out? If it works, feel free to send a PR.

koisto:
I'll give it a try and raise a PR later today.

koisto:
I have now raised the [pull request](https://github.com/jgromes/RadioLib/pull/1560)

koisto:
Apologies, after some further testing I have realised that my change to the detatchInterrupt function was unnecessary and it actually introduces more problems as it doesn't clear any flags that may have been set previously. 

Line 80 in PicoHal.h should be as it was originally:

``` c
    gpio_set_irq_enabled_with_callback(interruptNum, GPIO_IRQ_EDGE_RISE | GPIO_IRQ_EDGE_FALL, false, NULL);
```

and not:

``` c
    gpio_set_irq_enabled_with_callback(interruptNum, 0, false, NULL);
```

[...]

PR #1560, opened 2025-07-18
Remove hard coded interrupt mode from PicoHAL

koisto:
Modified the `attachInterrupt` function in PicoHal.h so that the `mode` argument is passed into the events argument of `gpio_set_irq_enabled_with_callback`. Previously this used hard coded values that caused spurious (extra) interrupts.

This was tested using a Raspberry Pi Pico and an RFM96W module using code based on examples/SX127x/SX127x_Receive_Interrupt/SX127x_Receive_Interrupt.ino with a counter added into the receive interrupt callback.

``` c
volatile bool receivedFlag = false;
volatile int rx_count = 0;

void setFlag(void) {
  // we got a packet, set the flag
  receivedFlag = true;
  rx_count++;

}
```

The `receivedFlag` is serviced in the the `main` functions loop as follows:

``` c
        if (receivedFlag)
        {
            // reset flag
            receivedFlag = false;

            int numBytes = radio.getPacketLength();
            int rx_state = radio.readData(rx_buffer, numBytes);

            printf("[RFM96] Received packet %d %d %d %f %f %f\n", rx_count, numBytes, rx_state, radio.getRSSI(), radio.getSNR(), radio.getFrequencyError());
 
     }
```
This image shows the output after 5 packets are sent before the modification to the HAL. The interrupt routine fires twice for each received packet. 
Pasted image 20250718082405

This image shows the output after 5 packets are sent after the modification to the HAL. The interrupt routine fires only once for each received packet.
Pasted image 20250718081804

### DIFF
--- a/src/hal/RPiPico/PicoHal.h
+++ b/src/hal/RPiPico/PicoHal.h
@@ -69,7 +69,7 @@ public:
       return;
     }
 
-    gpio_set_irq_enabled_with_callback(interruptNum, GPIO_IRQ_EDGE_RISE | GPIO_IRQ_EDGE_FALL, true, (gpio_irq_callback_t)interruptCb);
+    gpio_set_irq_enabled_with_callback(interruptNum, mode, true, (gpio_irq_callback_t)interruptCb);
   }
 
   void detachInterrupt(uint32_t interruptNum) override {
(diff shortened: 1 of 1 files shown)


koisto:
This pull request is related to [this issue](https://github.com/jgromes/RadioLib/issues/1558)

jgromes:
@koisto thank you for the contribution. It all looks good, so I merged it, thanks!

PR #1573, opened 2025-08-14
[LoRaWAN] Channel improvements

StevenCellist:
This PR improves channel management for LoRaWAN. Also, I tried to improve the readability of the channel handling. I am not completely satisfied, but at least a whole bunch of weird choices is now removed or unified, resulting in a somewhat better reading experience. Even with the new features, the number of lines handling channels could be reduced measurably. Still, channel masks and a fair channel roulette is a monstrosity.

* [LoRaWAN] Fully support subband probing when no subband is configured.

Most notably, this PR lifts the restriction of a maximum number of 16 channels for fixed bands. RadioLib now supports the full 72 or 96 channels used in the fixed bands. With this change, I managed to implement the desired subband probing in case that no subband is configured. Behaviour before this PR is that a random channel is selected from each subband, which are then randomly cycled through. No tracking of used channels, no incremental subbank selection.

* [LoRaWAN] Retain join progress 'during deepsleep'

Behaviour before this PR is that the used channels are not tracked across 'deepsleep', requiring nodes to stay active until a JoinAccept is received. This is now solved, as 'partial' sessions can be restored without any notable difference.

As some collateral, a few more changes were made:
* [LoRaWAN] Remove the Join- / initial-datarate arguments from `activateX()` - now use `setDatarate()` before `activateX()` instead.

* [LoRaWAN] Use a pseudo-random RNG with a truly-random seed for channel roulette.

Behaviour before this PR is that each uplink is a truly-random selection.

* [LoRaWAN] Decrease Nonces buffer by one byte, breaking Nonces buffers from previous versions!

The session status byte has been moved to the session buffer which makes way more sense.

----

Changes are verified using LCTT for both US915 and EU868. 

While this PR will generate only a single changelog entry, I recommend adding some of the mentioned items above to the release notes whenever they will be generated.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.h
+++ b/src/protocols/LoRaWAN/LoRaWAN.h
@@ -320,7 +320,7 @@ constexpr LoRaWANPackage_t PackageTable[RADIOLIB_LORAWAN_NUM_SUPPORTED_PACKAGES]
 
 #define RADIOLIB_LORAWAN_PACKAGE_NONE { .packId = 0, .packFPort = 0, .isAppPack = false, .fixedFPort = false, .enabled = false, .callback = NULL }
 
-#define RADIOLIB_LORAWAN_NONCES_VERSION_VAL (0x0002)
+#define RADIOLIB_LORAWAN_NONCES_VERSION_VAL (0x0003)
 
 enum LoRaWANSchemeBase_t {
   RADIOLIB_LORAWAN_NONCES_START       = 0x00,


StevenCellist:
ESP32 is unable to keep up with the menacing LCTT during a deepsleep run - failing somewhere 200 uplinks into the run usually. But it passes all relevant tests one by one using deepsleep on EU868 and US915.

Modified the original message as overall behaviour is improved while still in line with original API.

StevenCellist:
@jgromes up to you whether you want to review. From my perspective all ready for merge.

StevenCellist:
> > Behaviour before this PR is that each uplink is a truly-random selection.
> 
> What is the motivation for the change? I'm guessing it mainly saves SPI transactions right?

With the new reservoir sampling using more random values, this indeed saves on SPI transactions. Especially when a node uses 64 or 96 channels.. ;)

StevenCellist:
It is only now that I see that `PhysicalLayer::random()` consumes 40ms! That's an optimization that shouldn't have gone unnoticed for this long :o
I thought it was 10ms, but it actually takes 4 bytes for a total of 40ms Rx time (plus quite some SPI overhead).

jgromes:
> It is only now that I see that PhysicalLayer::random() consumes 40ms!

That is indeed a significant amount of time. All looks good here, so merged, thank you! Is this change something that would warrant a release or should I hold on a bit longer for any additional changes in the pipeline?

StevenCellist:
No pressing reasons to release this at once. Let's wait for something more interesting to come up. And I'll have some time to fiddle with this version locally before release. LCTT isn't the definitive answer as I've seen before.

PR #1581, opened 2025-08-26
receive() method - improving and fixing bugs

MarusGradinaru:
I was experimenting with SX1262 LoRa modules using RadioLib and I found an issue with the receive() method.

The chip has a hardware timer that can be set in receive mode, and it raises a DIO1 IRQ on timeout if no packet is detected within the specified period. If a packet is detected in time, the timer is stopped and the payload reception continues. The problem is that in the RadioLib receive() method there is also a software safety timer, with the same timeout value (fixed to 100 LoRa symbols). However, this software timer does not stop when the packet is detected, but only after the packet is fully received. In some cases, when the payload is large, this causes the reception to stop too early. See detailed test info in the [receive tests.xlsx](https://github.com/user-attachments/files/21982828/receive.tests.xlsx) attached Excel file. 

The solution is to calculate the software timeout separately using getTimeOnAir, which accounts for the payload length. I implemented this, and although the calculated value matches the Semtech calculator, in practice the actual reception takes about 5–100 ms longer. I believe this comes from delays introduced by the code that prepares the reception (or possibly other factors). To compensate, I added a 160 ms safety margin.
Additionally, the hardware timer needs a minimum limit of 60 ms, because with some settings (e.g. 500 kHz, SF5) the calculated timeout can be very small (5 ms), which also causes failures. With these modifications, the receive() method now passes all of my tests.
Note: For the FSK mode the behavior is unchanged, as I didn't study it and I don't understand how it works. Maybe, in the future, someone will correct that too...

I also added an optional parameter (exTimeout) to the receive() method, allowing the user to extend the calculated timeout if they know that the transmitting module requires extra time to process the request and prepare a reply. This is very important in some use cases.
To add this parameter I had to modify receive() in PhysicalLayer and then implement overrides in all modules that provide this method. I successfully added exTimeout support in all modules except SX127x, because I do not fully understand its logic and did not want to break anything, so I left it unchanged. The code remains backward compatible, as the exTimeout parameter is optional.

While implementing this new parameter, I also found a bug in SX128x: the calculated timeout was in microseconds, but later it was compared with millis() (which returns milliseconds). I fixed this as well.

The modifications summary:
 - SX126x.cpp -> updated for user extra timeout / fixed bug: software timeout doesn't account for the full packet to be received
 - SX128x.cpp -> updated for user extra timeout / fixed bug: microseconds timeout compared with millis() in safety timer
 - LR11x0.cpp -> updated for user extra timeout 
 - Si443x.cpp -> updated for user extra timeout
 - CC1101.cpp -> updated for user extra timeout
 - RF69.cpp   -> updated for user extra timeout
 - nRF24.cpp  -> updated for user extra timeout
 - SX127x.cpp -> extra timeout not used / I don't understand the code logic for this chip

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -58,9 +58,9 @@ int16_t CC1101::transmit(const uint8_t* data, size_t len, uint8_t addr) {
   return(finishTransmit());
 }
 
-int16_t CC1101::receive(uint8_t* data, size_t len) {
-  // calculate timeout (500 ms + 400 full max-length packets at current bit rate)
-  RadioLibTime_t timeout = 500 + (1.0f/(this->bitRate))*(RADIOLIB_CC1101_MAX_PACKET_LENGTH*400.0f);
+int16_t CC1101::receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) {
+  // calculate timeout (500 ms + 400 full max-length packets at current bit rate + user extra timeout)
+  RadioLibTime_t timeout = 500 + (1.0f/(this->bitRate))*(RADIOLIB_CC1101_MAX_PACKET_LENGTH*400.0f) + exTimeout;
 
   // start reception
   int16_t state = startReceive();
--- a/src/modules/CC1101/CC1101.h
+++ b/src/modules/CC1101/CC1101.h
@@ -599,9 +599,11 @@ class CC1101: public PhysicalLayer {
       Overloads for string-based transmissions are implemented in PhysicalLayer.
       \param data Binary data to be sent.
       \param len Number of bytes to send.
+      \param exTimeout User extra timeout (ms). 
+      Used to allow the sender to prepare its reply, if needed.
       \returns \ref status_codes
     */
-    int16_t receive(uint8_t* data, size_t len) override;
+    int16_t receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) override;
 
     /*!
       \brief Sets the module to standby mode.
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -223,7 +223,7 @@ int16_t LR11x0::transmit(const uint8_t* data, size_t len, uint8_t addr) {
   return(finishTransmit());
 }
 
-int16_t LR11x0::receive(uint8_t* data, size_t len) {
+int16_t LR11x0::receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) {
   // set mode to standby
   int16_t state = standby();
   RADIOLIB_ASSERT(state);
@@ -256,6 +256,9 @@ int16_t LR11x0::receive(uint8_t* data, size_t len) {
     return(RADIOLIB_ERR_UNKNOWN);
   
   }
+
+  // adding user extra timeout
+  timeout += exTimeout;  
 
   RADIOLIB_DEBUG_BASIC_PRINTLN("Timeout in %lu ms", timeout);
 
--- a/src/modules/LR11x0/LR11x0.h
+++ b/src/modules/LR11x0/LR11x0.h
@@ -992,9 +992,11 @@ class LR11x0: public PhysicalLayer {
       Overloads for string-based transmissions are implemented in PhysicalLayer.
       \param data Binary data to be sent.
       \param len Number of bytes to send.
+      \param exTimeout User extra timeout (ms). 
+      Used to allow the sender to prepare its reply, if needed.
       \returns \ref status_codes
     */
-    int16_t receive(uint8_t* data, size_t len) override;
+    int16_t receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) override;
 
     /*!
       \brief Starts direct mode transmission.
--- a/src/modules/RF69/RF69.cpp
+++ b/src/modules/RF69/RF69.cpp
@@ -122,9 +122,9 @@ int16_t RF69::transmit(const uint8_t* data, size_t len, uint8_t addr) {
   return(finishTransmit());
 }
 
-int16_t RF69::receive(uint8_t* data, size_t len) {
-  // calculate timeout (500 ms + 400 full 64-byte packets at current bit rate)
-  RadioLibTime_t timeout = 500 + (1.0f/(this->bitRate))*(RADIOLIB_RF69_MAX_PACKET_LENGTH*400.0f);
+int16_t RF69::receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) {
+  // calculate timeout (500 ms + 400 full 64-byte packets at current bit rate + user extra timeout)
+  RadioLibTime_t timeout = 500 + (1.0f/(this->bitRate))*(RADIOLIB_RF69_MAX_PACKET_LENGTH*400.0f) + exTimeout;
 
   // start reception
   int16_t state = startReceive();
--- a/src/modules/RF69/RF69.h
+++ b/src/modules/RF69/RF69.h
@@ -530,9 +530,11 @@ class RF69: public PhysicalLayer {
       Overloads for string-based transmissions are implemented in PhysicalLayer.
       \param data Binary data to be sent.
       \param len Number of bytes to send.
+      \param exTimeout User extra timeout (ms). 
+      Used to allow the sender to prepare its reply, if needed.
       \returns \ref status_codes
     */
-    int16_t receive(uint8_t* data, size_t len) override;
+    int16_t receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) override;
 
     /*!
       \brief Sets the module to sleep mode.
--- a/src/modules/SX126x/SX126x.cpp
+++ b/src/modules/SX126x/SX126x.cpp
@@ -249,20 +249,25 @@ int16_t SX126x::transmit(const uint8_t* data, size_t len, uint8_t addr) {
   return(finishTransmit());
 }
 
-int16_t SX126x::receive(uint8_t* data, size_t len) {
+int16_t SX126x::receive(uint8_t* data, size_t len, RadioLibTime_t exTimeout) {
   // set mode to standby
   int16_t state = standby();
   RADIOLIB_ASSERT(state);
 
-  RadioLibTime_t timeout = 0;
+  RadioLibTime_t hwTimeout = 0;
+  RadioLibTime_t swTimeout = 0;
 
   // get currently active modem
   uint8_t modem = getPacketType();
   if(modem == RADIOLIB_SX126X_PACKET_TYPE_LORA) {
-    // calculate timeout (100 LoRa symbols, the default for SX127x series)
+    // calculate hardware timeout (100 LoRa symbols, the default for SX127x series)
     float symbolLength = (float)(uint32_t(1) << this->spreadingFactor) / (float)this->bandwidthKhz;
-    timeout = (RadioLibTime_t)(symbolLength * 100.0f);
-  
+    hwTimeout = (RadioLibTime_t)(symbolLength * 100.0f);
+    if (hwTimeout < 60) { hwTimeout = 60; }  // limit it to a resonable minimum of 60 ms
+ 
+    // calculate software timeout, taking into account the full packet length and +160 ms safety margin
+    swTimeout = (RadioLibTime_t)((getTimeOnAir(len) + 999ul) / 1000ul) + 160ul;
+
   } else if(modem == RADIOLIB_SX126X_PACKET_TYPE_GFSK) {
     // calculate timeout (500 % of expected time-one-air)
     size_t maxLen = len;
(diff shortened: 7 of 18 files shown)


jgromes:
I feel like there are few different things getting tangled in a single PR, so I will try to address them separately:

1. **SX126x hardware timeout vs RadioLib software timeout** - indeed, the current implementation may return a timeout while if it happens that packet reception has been started. There are a few ways to solve this, either checking IRQ bits, or as you have done, simply extending the software timeout duration. However, your approach is IMO a strange combination of accurate `getTimeOnAir` calculations, with added constants determined by experiment. There is no guarantee on different hardware your constants will still be valid. So why not just take the original value, double it and use that as the software timeout? It will be much simpler while fulfilling the exact same purpose - because with the multiple additions to the timeout duration, it will always be the radio raising the timeout, not the software (outside of failure scenarios, of course).

2. **Allowing user to manipulate the timeout duration using `exTimeout`** - I really don't like this change, because it has a very non-intuitive usage. It is added on top of the internally calculated timeout, which the user has no control over. If there is a use case that requires precise control of the timeout duration, `receive` is really not the method to use. `startReceive` and `readData` is the way to go, and if you want to implement some complicated timing mechanisms between those two calls, then feel free to do so. I really don't understand the use case here. An argument could be made that there should be a plain `timeout` argument for the `receive` method. But having part of the timeout calculated and another part user-provided is strange.

3. I am strongly opposed to do this only for some modules. The behavior really has to be consistent - `receive` on CC1101 has to perform the same functionality as it does on SX1278, that is one of the core design principles of this library.

4. **SX128x timeout units** - this might actually be a bug, I will double check. We could just use `micros()` timer for the comparison. But in any case, lumping it together into one PR (and even one commit) with other stuff is plain wrong, as it will now have to be solved here :)

@StevenCellist feel free to add your comments as well.

MarusGradinaru:
Ok, I understand...

> 2\. If there is a use case that requires precise control of the timeout duration, `receive` is really not the method to use. `startReceive` and `readData` is the way to go, and if you want to implement some complicated timing mechanisms between those two calls, then feel free to do so.

This is the first thing that I tried, to replicate externaly the code from the receive() method. But I didn't find a way to get the values of the used bandwidth and spread factor to calculate the hardware timeout, and neither to call the `fixImplicitTimeout()`, because it is private. Then I wanted to write my own `fixImplicitTimeout()` writing to the chip registers, but again, those functions are private too... 

It is ok if you don't like my solution, but you must fix somehow (in your own way) the bug from the SX1262.receive() method that stop the reception to early on big data packets, and the unpractical small hardware timeout values that results in some configurations, because as it is now, that method cannot be used.

StevenCellist:
I'd like to mention a few things as well:
1. If these changes are useful to you, all the power to you - use it as you wish, be happy with your work as you've got a reliable solution that fits your needs! Please note that the feedback is _on this pull request_, not your usecase. While we may have strong opinions on whether this PR suits a merge, that must not stop you from being happy with your solution!
2. I can imagine it took you quite a lot of time, head scratching, maths and testing to prepare these changes. But we're looking at rejecting this pull request, which may be sad for you as you've tried your best to contribute something useful from your perspective. My advise for future you, is that you always open up a little discussion first about your proposed changes, to hear whether that is something that the developers/maintainers think is good for the project. That will prevent mismatched expectations early in the process!
3. Having played with a bunch of different radios over the last few years, it's become quite obvious that there are some jarring differences between even the simplest things. One radio expects a timeout in symbols, the other in microseconds. Some IRQ values overlap, some are completely different. I've even 'forced' Mr. Gromes the Wise to expand the `PhysicalLayer` with non-trivial stuff to completely abstract those differences away. And essentially, you are proposing double work by adding those features to `receive()` as well. But they are incomplete as they only apply to LoRa, although understandably since you only used that.
4. There are a lot of extra features besides a simple `transmit()` and `receive()`, such as reading IRQs, RSSI, SNR etc. These are clearly not accessible when using `receive()`. The purpose of `receive()` is/was to allow the very very beginners to receive a demo message, nothing more. But `receive()` is/was never intended to be used as an actual production method, because you are wise to use more signals such as IRQs.
5. > although the calculated value matches the Semtech calculator, in practice the actual reception takes about 5–100 ms longer. I believe this comes from delays introduced by the code that prepares the reception (or possibly other factors). To compensate, I added a 160 ms safety margin.

   This is likely a user error or once again a limitation of `transmit()` and `receive()`, since I have successfully been able to receive messages with only 3 milliseconds of padding in front of and 3 milliseconds at the end of the reception window using `startTransmit()` and `startReceive()`. Solving this by slapping on additional radio time is a bit of an ostrich-solution, as we call it in Dutch :)
7. Please re-read point 1! (This makes it a proper feedback 'sandwich' ;) )

MarusGradinaru:
Indeed, I used a lot of time and effort to do all the tests and make `receive()` usable (because now is not). At first I was scared, because I tougth that if the PR is not accepted I will never be able to receice LoRa packets with timeouts, because in SX1262 is a hardware bug that needs `fixImplicitTimeout()` workaround called after reception, and that function is private in RadioLib. Of course, I could use my modified source code, but that means I will lose future RadioLib updates. Thats what I thougth. :) What I didn't know was that I can use my forked version with these modification and still update the main branch to be in sync with official RadioLib. Which saved me ! :)

> 3\. But they are incomplete as they only apply to LoRa, although understandably since you only used that.

It may be incomplete, but it's much better than `receive()` being completely useless. I can't build serious projects based on a function that only works under special conditions (small buffers). My logic was that I would make the code work for the LoRa SX1262 chip and leave the rest unchanged, and in the future someone who knows how will make it work for the other chips as well.

> 4\. because you are wise to use more signals such as IRQs.

Yes, but I need to access `fixImplicitTimeout()` from outside.

> 4\. There are a lot of extra features besides a simple `transmit()` and `receive()`, such as reading IRQs, RSSI, SNR etc. These are clearly not accessible when using `receive()`.

They are ! :) I allready use them in my project (I will upload it in my repo soon). You can get the RSSI, SNR, FreqError... after the `receive()` method has returned.

StevenCellist:
To be **very explicit**: anything you do with `receive()` can be done using `startReceive()`, and more. We **strongly** encourage you to use `startReceive()`, even though you spent your time making `receive()` more useful for your project. You have definitely learned a lot, so your time is not wasted, because now you know even more to be able to use `startReceive()`. 
If you want to know the required timeout for your big packet, you can use `getTimeOnAir()` which internally reads the bandwidth and spreading factor, so you don't need to have access to those hidden properties. 

> They are ... _after_ the receive() method has returned.

That is the difference, you have to wait until the reception has finished. When using `startReceive()`, you can even request them _during_ reception, which can be very useful. E.g.: if you started reception and expect a packet to begin its transmission within 20ms, you can request the IRQ after 50ms. If there is a RxTimeout IRQ, that means that there is no reception in progress, and can already stop reception, saving battery life. If there is no RxTimeout IRQ, then you wait until the full packet is received, up to a maximum of the TimeOnAir (plus a little bit extra such as your 60ms).

MarusGradinaru:
> **anything** you do with `receive()` can be done using `startReceive()`, and more.

Not "anything". I cannot call `fixImplicitTimeout()`, which is mandatory after a timeout receive.

StevenCellist:
~I've been using SX126x radios for two years extensively and never needed to call that using `startReceive()`.~ Edit: I don't use implicit mode, so can't comment on that with confidence actually.
As long as you supply `startReceive()` with a proper timeout value, you shouldn't run into any issues.

https://github.com/jgromes/RadioLib/blob/3d94858f368adea0759aebabbb58f044c228676c/src/protocols/PhysicalLayer/PhysicalLayer.h#L308-L318

Otherwise - wouldn't that be the actual issue to fix?

jgromes:
@MarusGradinaru 

>  I cannot call fixImplicitTimeout(), which is mandatory after a timeout receive.

In that case, we need to fix this real issue. If there is a use case for blocking receive, then we should talk about how to support it. On a side note, this fix only has to be called if you are using implicit header mode - otherwise it actually does nothing, and that is probably why other people did not run into the issue. So I disagree `receive` is broken in general ;)

To be perfectly honest I'm thinking about removing the `receive` API entirely. A lot of the time it just confuses people because it returns the timeout error. So let's discuss how to implement it using `startReceive`/`readData`, rewrite the examples and make sure it works.

MarusGradinaru:
@jgromes 
Blocking receive, is verry usefull ! Most of the time you need to read something before taking other actions, like I did [here](https://github.com/MarusGradinaru/LoRa-Test/blob/main/src/main.cpp#L545). There I send a packet with a command to the other module, and then I need to wait for the replay packet with the data before going further. Of course, you can struggle doing it with `startReceive()` and other stuff (when `fixImplicitTimeout()` will be made public), but why shouldn't there be a function that do all the low-level stuff ? 

> So let's discuss how to implement it using `startReceive`/`readData`, rewrite the examples and make sure it works.

Well, haven't I already done this ? :)
You said that you don't like the small compensation I made for the hardware timeout (60ms) and software timeout (160ms). These are just timeouts, they will not hurt anybody if the code waits a few milliseconds more for the data to arrive, especially since LoRa is not a verry high speed transfer protocol. Regarding the 5ms timeout that results for the hardware timer (with 500KHz and SF5), it is insanely small for the sending module to have time to respond. We can even set a one-second default timeout for all timers.
And if the module that needs to respond has to process some data first, the user must be able to add an extra timeout to give the other module time to do its job.

jgromes:
@MarusGradinaru first of all, it really brings me no great joy to type all of this, as you have clearly spent some time one the issue, but nevertheless ...

The example you have posted is actually a pretty good argument as to why the blocking `receive` method should not exist. You are transmitting a packet and expecting a reply from another device. However, the sucess of the operation depends on whether you can synchronize the two devices accurately enough. You send a packet, and then you have a brief window in which you can receive a reply. Protocols like this do exist (for example - LoRaWAN), but everything depends on the synchronization between the nodes and these kinds of protocols usually require agreed, accurate time intervals between the end of transmission and start of receive window. You work around this by adding extra margins around the blocking `receive` timeout, so in your case, it ends up working for now.

There's a bigger problem though - if the second device replies sooner than the first one has called `receive`, then you will miss the packet. There can be many reasons for this desynchronization, in your multi-threaded environment, the sending task might get blocked by something that has higher priority (since `transmit` is blocking as well).

This would be a lot simpler using interrupts. Just set the both devices to start receiving, send packet with `startTransmit` whenever needed (user input or a packet from another device - does not matter what is the source), then process the interrupt flag and decide whether it was caused by Tx or Rx. We even have [examples for this](https://github.com/jgromes/RadioLib/blob/master/examples/SX126x/SX126x_PingPong/SX126x_PingPong.ino). Your device will not miss packets and you don't have to mess with timing.

Speaking out of experience, this sort of chaining of blocking functions/methods together and hoping that the timing works out just never works reliably. Sure, you might be able to get it working for a specific setup, but as soon as some uncertainty is introduced (more threads, sensors, user input etc.), I can almost guarantee it will no longer work reliably. Had there been no `receive` method, you would probably be forced to implement this differently.

> You said that you don't like the small compensation I made

Not really, what I specifically don't like is to have a combination of calculated and user-provided timeout.

> These are just timeouts, they will not hurt anybody

To me it really hurts clarity of the interface. If I have a function that has a timeout argument, I expect that to be ... well, the timeout. I definitely do not expect that this is only a part of the timeout, which is added to an automatically calcualted part I (as the user) have no control over. It can be argued this can be documented, but experience from this project shows the amount of people that actually read the docs is well below 50% (despite much effort that went into them).

> if the module that needs to respond has to process some data first, the user must be able to add an extra timeout to give the other module time to do its job

Again, this is a design antipattern - you work around an unknown processing duration on a different device by adding more margins on the timeout instead of just letting the first device listen for a reply, and do other stuff in the meantime. So adding these timeouts really just encourages bad design practices.

### Bottom line
If you read all of that, thank you. The way I see this, there are two ways to handle this:

1. Add a timeout argument (in milliseconds) to the receive method, which is completely within user control. If set to 0, the timeout duration will be calculated automatically, with some lower bound. If set to non-zero, that will be the timeout and it is up to the user to set it. If they set it incorrectly, that's on them.

2. Doing away with the `receive` method entirely. We will rewrite the current blocking receive examples to use a combination of `startReceive`/`readData` with a blocking loop between those. We will also include some sort of `finishReceive` method that will take care of the cleanup (for example - calling `fixImplicitTimeout` on SX1262).

The second one is definitely a breaking change that would require a major release of the library (8.0.0). The first one is change in the public API so at least a new minor release will be needed.

MarusGradinaru:
@jgromes Thank you for taking the time to explain all this to me. This was my first PR ever, so it was expected that I would screw it up... :) Do you remember a few weeks ago when you told me to do a PR to fix a small bug I had found and I asked you, what is a PR ? :) I am verry new to all this GitHub stuff, but I learn fast...

Now I understand why you're saying to receive asynchronously. Until now, I created communication protocols only for TCP networks, that's way my mind was set to use blocking methods. I will try to rethink my project code to work as you recommended, with IRQ's instead of blocking. 

I don't know what to say about the blocking `receive()`... Maybe a combination of those two points you provide, will be better: Adding a single fixed timeout parameter in total user control (if you want I can modify this PR to do that), but definitely a `finishReceive()` which calls `fixImplicitTimeout()` should exists too ! 
But I dont't understand what you are saying here: "_If set to 0, the timeout duration will be calculated automatically, with some lower bound_". What is "lower bound" ? The way that timeout is now calculated ? Because if so, it will need adjustments, as I already told you that the sotware timeout does not take into account the full packet length. It is wrong to use the same timeout for software timer as the hardware time. If you want a single timeout, we can use TimeOnAir for both.

jgromes:
@MarusGradinaru I do remember, and I am actually very glad you followed up on this, as most people do not :)

> but definitely a finishReceive() which calls fixImplicitTimeout() should exists too !

I think you might be right, we have a `finishTransmit` method to clean up after transmission, so its Rx equivalent will be here as well, even if it will not do much most of the time.

> What is "lower bound" ?

I meant to limit the calculation to some reasonable minimum value, which i think is already included in your PR. 

To move this forward, I think the following steps are needed.

1. I will close this PR. As stated before, it contains multiple changes, which should be separated.
2. @MarusGradinaru feel free to submit a new PR for the actual bug here, i.e. the SX128x timeout units. Those should indeed be in microseconds, as the default timeotu period base we use is 15.625 us.
3. I will rework the `receive` API to make the timeout fully accessible to the user.

PR #1587, opened 2025-09-01
Add support for LoRa coding rate 4/4

GUVWAF:
As discussed in #1584, this adds support for coding rate 4/4, meaning no coding. A warning is added that this is not documented and not recommended without applying your own FEC.

Tested to be working on SX1278, SX1262 and LR1110.


jgromes:
Thanks for the contribution, it all looks good so merged!

PR #1588, opened 2025-09-01
Add methods to retrieve received LoRa header info

GUVWAF:
Related to #1584. Since in explicit header mode transmitter and receiver do not necessarily need to use the same coding rate and CRC enable status, it’s useful to know these from a received packet, e.g. for rate control or appropriate airtime calculations (PR to follow).

I used the same function name as existed for the LR11x0, but changed the parameters to explicitly return the CR and CRC.

Tested on SX1276, SX1262 and LR1110.

### DIFF
--- a/src/modules/SX126x/SX126x.h
+++ b/src/modules/SX126x/SX126x.h
@@ -120,7 +120,8 @@
 #define RADIOLIB_SX126X_REG_IQ_CONFIG                           0x0736
 #define RADIOLIB_SX126X_REG_LORA_SYNC_WORD_MSB                  0x0740
 #define RADIOLIB_SX126X_REG_LORA_SYNC_WORD_LSB                  0x0741
-#define RADIOLIB_SX126X_REG_FREQ_ERROR                          0x076B
+#define RADIOLIB_SX126X_REG_LORA_RX_CODING_RATE                 0x0749
+#define RADIOLIB_SX126X_REG_FREQ_ERROR_RX_CRC                   0x076B
 #define RADIOLIB_SX126X_REG_SPECTRAL_SCAN_STATUS                0x07CD
 #define RADIOLIB_SX126X_REG_RX_ADDR_PTR                         0x0803
 #define RADIOLIB_SX126X_REG_RANDOM_NUMBER_0                     0x0819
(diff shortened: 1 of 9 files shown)


src/modules/SX128x/SX128x.h
```diff
@@ -84,6 +84,7 @@
 #define RADIOLIB_SX128X_REG_FREQ_ERROR_CORRECTION               0x093C
 #define RADIOLIB_SX128X_REG_LORA_SYNC_WORD_MSB                  0x0944
 #define RADIOLIB_SX128X_REG_LORA_SYNC_WORD_LSB                  0x0945
+#define RADIOLIB_SX128X_REG_LORA_RX_CODING_RATE                 0x0950
```
jgromes, line 87: Similar queestion as for the SX126x driver - where is this adress from?

jgromes (commented):
Thank you for the contribution, I left some minor comments. Also, as the `getLoRaRxHeaderInfos` will be part of the public API, please add it to keywords.txt (the file uses true tabs, not spaces). Thanks!

And finally on a very nitpicky note, `getLoRaRxHeaderInfo` (without the -s) might be a better name, I originally named it `getLoRaRxHeaderInfos` to keep the naming from LR11x0 datasheet.

src/modules/SX128x/SX128x.h
GUVWAF: It’s from the latest version of the SX1280 datasheet (Rev. 3.3.), page 102.

GUVWAF:
Thank you, I will rename the function and add it to the keywords.txt later today.

src/modules/SX126x/SX126x.h
```diff
@@ -120,7 +120,8 @@
 #define RADIOLIB_SX126X_REG_IQ_CONFIG                           0x0736
 #define RADIOLIB_SX126X_REG_LORA_SYNC_WORD_MSB                  0x0740
 #define RADIOLIB_SX126X_REG_LORA_SYNC_WORD_LSB                  0x0741
-#define RADIOLIB_SX126X_REG_FREQ_ERROR                          0x076B
+#define RADIOLIB_SX126X_REG_LORA_RX_CODING_RATE                 0x0749
+#define RADIOLIB_SX126X_REG_FREQ_ERROR_RX_CRC                   0x076B
```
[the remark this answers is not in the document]
jgromes: Ah, thanks for the clarification, and for letting me know about the updated datasheet. It also looks like there are some other updates in it, but that's for future work.

GUVWAF:
I updated the PR with the requested changes. 

Regarding the `calculateTimeOnAir()` function: this is more work than anticipated and I'll need to write tests to ensure the behaviour of `getTimeOnAir()` is still the same, so this is going to take a while still.

jgromes:
Looking good, thank you! Merged.

On a side note, if you will be writing some tests, perhaps it would be a good idea to include them in the (still WIP) [unit tests](https://github.com/jgromes/RadioLib/tree/master/extras/test/unit/tests) we have, using the boost C++ library. `calculateTimeOnAir` function seems like something that would be very well suited to a unit test, requiring no interaction with the radio hardware.

PR #1590, opened 2025-09-05
Fixing SX128x.receive() timeout bug

MarusGradinaru:
I found a bug in `SX128x.receive()` method. The timeout is calculated microseconds (this is what `getTimeOnAir()` returns), but later, in the safety software time, it is compared with millis() (which returns milliseconds). The units should be the same, so I corrected the safety timer to use `micros()` instead of `millis()`. And I also took care of the debug message which outputs the timeout value in milliseconds (ceiling the micros).

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -363,7 +363,7 @@ int16_t SX128x::receive(uint8_t* data, size_t len) {
 
   // calculate timeout (1000% of expected time-on-air)
   RadioLibTime_t timeout = getTimeOnAir(len) * 10;
-  RADIOLIB_DEBUG_BASIC_PRINTLN("Timeout in %lu ms", timeout);
+  RADIOLIB_DEBUG_BASIC_PRINTLN("Timeout in %lu ms", (uint32_t)((timeout + 999) / 1000));
 
   // start reception
   uint32_t timeoutValue = (uint32_t)((float)timeout / 15.625f);
@@ -372,11 +372,11 @@ int16_t SX128x::receive(uint8_t* data, size_t len) {
 
   // wait for packet reception or timeout
   bool softTimeout = false;
-  RadioLibTime_t start = this->mod->hal->millis();
+  RadioLibTime_t start = this->mod->hal->micros();
   while(!this->mod->hal->digitalRead(this->mod->getIrq())) {
     this->mod->hal->yield();
     // safety check, the timeout should be done by the radio
-    if(this->mod->hal->millis() - start > timeout) {
+    if(this->mod->hal->micros() - start > timeout) {
       softTimeout = true;
       break;
     }


jgromes:
All looks good - merged, thank you for the contribution!

MarusGradinaru:
I'm glad I was helpful. 

I see here this message (don't know if it is visible for everybody):
"_You're all set — the SX128x.receive()-timeout-bug branch can be safely deleted. If you wish, you can also delete this fork of jgromes/RadioLib in the settings_"

So, I understand that I can delete the branch, and even the fork used for PR, from my repository, right ?
Can I leave the fork for a little while, as it may be needed to PR the other bug that I discovered recenly (the one with LDR optimisation) ?

jgromes:
You can safely delete the branch, it has been merged so there is no risk of losing any information from it. Regarding the fork, you can keep it up, or you can delete it and fork it again in the future, that's up to you. Keep in mind that if you choose to keep it and later want to contribute to RadioLib again, you will most likely need to merge the latest master to your fork to get it up-to-date. Otherwise, you will be sending a pull request "from the past", which can lead to merge conflicts.

lyusupov:
@jgromes @MarusGradinaru 

`startReceive( timeoutValue )` function call sets timeout _( in microseconds )_ to a vaule less than `getTimeOnAir() `

<br>

image

MarusGradinaru:
Yes, because the SX1262 chip wants the timeout value in its own clock time steps, which is 15.625 microseconds per step.

lyusupov:
1) 
> because the SX1262 chip want

This topic is about SX128x

2) startReceive( timeoutValue ...) is a part of `PhysicalLayer` API

image

MarusGradinaru:
1. Sorry about the confusion. I am working on a project with the SX1262 and my mind was stuck on it. :) I think SX1268 works the same way, if the timeout was divided in clock steps by the one who wrote the library.
2. Yes, and it is overriden in SX128x.cpp. I think each module interpret the timeout parameter in its own way...

jgromes:
@lyusupov the timeout argument of `PhysicalLayer::receive` is in radio-specific units, as per the documentation:

https://github.com/jgromes/RadioLib/blob/6c38520393e4714d2ba560245c2aca8ac6ed0487/src/protocols/PhysicalLayer/PhysicalLayer.h#L310-L318

This is also the case for the `ReceiveConfig_t` structure in `RadioModeConfig_t`, so I don't understand what is the question.

lyusupov:
>  it is overriden in SX128x.cpp

SX128x.h overrides startReceive() **with NO arguments**.
startReceive( ... ) **with arguments** is still expected to be processed by PhysicalLayer API.


image

jgromes:
@lyusupov I don't understand how those things are related to each other, or indeed this pull request in particular since it did not change the `timeoutValue` calculation or the call to `startReceive`. Not even talking about the fact that `startReceive` override without arguments uses the `RADIOLIB_SX128X_RX_TIMEOUT_INF` macro, so I really don't understand what you're trying to point out.

If you think there is another issue, please file a new issue report with all the necessary information (preferrably with text as opposed to screenshots of code).

PR #1592, opened 2025-09-07
User-controlled timeout for blocking receive

jgromes:
As discussed in #1581, this PR adds simple way for user to directly control the timeout of the blocking `receive` methods by adding a `timeout` argument (in milliseconds). If set to 0 (the default value), the timeout duration is calculated the same way as before.

I also added a `finishReceive` method to clean up after reception.

@StevenCellist I would be very glad for a review, thanks!

cc @MarusGradinaru

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -133,6 +133,7 @@ startTransmit	KEYWORD2
 finishTransmit	KEYWORD2
 startReceive	KEYWORD2
 readData	KEYWORD2
+finishReceive	KEYWORD2
 startChannelScan	KEYWORD2
 getChannelScanResult	KEYWORD2
 setBandwidth	KEYWORD2
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -75,8 +75,7 @@ int16_t CC1101::receive(uint8_t* data, size_t len, RadioLibTime_t timeout) {
     this->mod->hal->yield();
 
     if(this->mod->hal->millis() - start > timeout) {
-      standby();
-      SPIsendCommand(RADIOLIB_CC1101_CMD_FLUSH_RX);
+      (void)finishReceive();
       return(RADIOLIB_ERR_RX_TIMEOUT);
     }
   }
@@ -87,8 +86,7 @@ int16_t CC1101::receive(uint8_t* data, size_t len, RadioLibTime_t timeout) {
     this->mod->hal->yield();
 
     if(this->mod->hal->millis() - start > timeout) {
-      standby();
-      SPIsendCommand(RADIOLIB_CC1101_CMD_FLUSH_RX);
+      (void)finishReceive();
       return(RADIOLIB_ERR_RX_TIMEOUT);
     }
   }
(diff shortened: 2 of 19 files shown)


MarusGradinaru:
I have analyzed the changes for SX126x (the only one I worked with) and, as far as I understand, when no timeout is provided, the software timer still uses the timeout calculated for the hardware timer, which does not take into account the length of the packet. So, with default timeout, it will fail with big payloads, like before. It would be more correct for the timeout to be calculated with getTimeOnAir, as is already done for GFSK, and this value to be used for both the hardware and software timers. Of course, it still may fail occasionally, without a safe margin added, because small delays are introduced in user code when setup things, but it will happen much less often. 

Thanks for the `finishReceive()` ! 😃

jgromes:
@MarusGradinaru I think you might be right, theres no point keeping all of the sub-GHz LoRa radios tied to 100 symbols (which started as teh default timeout value on SX127x). So I went ahead and cahgned them all to 500% of expected ToA, it even simplifies the logic a bit which is always nice.

MarusGradinaru:
Indeed... A little more time, than expected TOA, for the receive timeout won't do any harm. At first, I thought that 5 x TOA is too much, because the maximum posible TOA is 197.58 seconds, but I think that nobody will go that far, because I don't even think it's legal to broadcast that much continuously. But then, the minimum TOA is 2 ms... so, 5x will be nice here. Better a longer timeout than a shorter one.

And one more thing...
Check out the datasheet for SX1262, page 75. If I understand correctly, the hardware timeout should be limited to 262 seconds (16768000 clock steps), even if the user requests more... It may be needed a `if(timeoutValue > 0xFFDC00) { timeoutValue = 0xFFDC00; }` before `startReceive()`, because if the `timeoutValue` happens to be 0xFFFFFF, it means infinite timeout.

screenshot 48

jgromes:
> If I understand correctly, the hardware timeout should be limited to 262 seconds

To be perfectly honest, I don't think this is ever going to be an issue. Indeed, to be absolutely accurate, we should limit the maximum calcualted raw value, but I don't think there is much chance someone will ever try to do that. Though if it does happen, we can always add it later.

jgromes:
To not block this feature from getting into master any further, I will go ahead and merge this PR. I have not tested it yet on some of the lesser-used modules (CC1101, Si4432 etc.), but there weren't that many changes in those; once my new testing hardware gets here, I will check it also on those.

MarusGradinaru:
@jgromes 
It seems that `finishReceive()` [here](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L292), clears the IRQ status [here](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L524), that is needed when reading the data [here](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L301), to check the CRC status [here](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L634)... 🤔

jgromes:
@MarusGradinaru you're right, the `finishReceive` call should be only in the timeout branch. And it also looks like I made the same mistake for SX127x - though the other modules looks good. I pushed a couple of commits addressing this problem.

PR #1600, opened 2025-09-16
[SX126x] Use a default of either 8 or 12 for minSymbols in startReceiveDutyCycleAuto

cleishm:
According to the Semtech datasheet (6.1.1.1, version 1.2), the minimum symbols to reliably latch a preamble is 12 for SF5-6. This PR updates the `startReceiveDutyCycleAuto(senderPreambleLength, minSymbols)` method to use a default of 12 for `minSymbols` when SF <= 6.


jgromes:
Thank you for the contribution! It looks OK, so I merged it.

PR #1602, opened 2025-09-16
[SX126x] For receiveDutyCycleAuto, check senderPreambleLength is shorter than configured preamble

cleishm:
Update documentation of `receiveDutyCycleAuto()` and add a check on the value of `senderPreambleLength`, returning RADIOLIB_ERR_INVALID_PREAMBLE_LENGTH if it exceeds the configured preamble length.

The check cause existing code to return an error instead, although arguably such code may already be causing unnoticed failures. The first commit only changes the documentation if preferred.


jgromes:
Thank you for the contribution - I think that checking this constraint makes sense, especially since the datasheet itself also calls for it. And even though it changes the current behavior, I don't think it is a major change. All the tests passed, so this can be merged, thank you!

PR #1603, opened 2025-09-16
[SX126x] No need to call standby() after a verified reset

cleishm:
Remove unnecessary call to `standby()` in `SX126x::modSetup()`.


jgromes:
Good catch - all looks, so merged, thank you for the contribution!

PR #1604, opened 2025-09-18
[SX126x] Remove extraneous reset after findChip

cleishm:
`SX126x::findChip` already resets and verifies the module, so there is no need to reset after. Removes that redundant reset in `modSetup`.


StevenCellist:
Someone's optimizing the start-up sequence! Love to see it :)

jgromes:
All looks good, so merged, thank you for the contribution!

On a side note, it is very likely the other modules also have similar redundancies.

PR #1606, opened 2025-09-19
Improve support for LoRa coding rate with Long Interleaver

GUVWAF:
As a last part related to #1584, this adds support for Long Interleaver coding rates for SX126x, as defined in the latest version of the datasheet. I added a note that it's not compatible with SX127x radios.

Furthermore, it adds some checks for the packet length, as the Long Interleaver does not support packets shorter than 8 bytes and longer than 253 (if CRC is enabled) according to the LR11x0 datasheet.

For SX128x, according to the datasheet, the only limitation is a maximum length of 253 when using CR 4/8 with the long interleaver and CRC enabled.


jgromes:
Thank you for the contribution! I have no comments so it can be merged as-is, and will be included in the release coming today or tomorrow ;)

PR #1607, opened 2025-09-19
[SX1262] Use named arguments for SX1262::begin* methods

cleishm:
Deprecates SX1262::begin* methods with long argument lists and provides replacements that take a struct of named arguments. Supports invocation using temporary named struct objects (C++14+), e.g.:

```
int state = radio.begin({
    .frequency = 434.0,
    .bandwidth = 125.0,
    .spreadingFactor = 9,
    .codingRate = 7,
    .preambleLength = 8
});
```

If there is support for this approach, I'm happy to extend this PR to apply the pattern in the other modules in RadioLib.

### DIFF
--- a/examples/SX126x/SX126x_Channel_Activity_Detection_Blocking/SX126x_Channel_Activity_Detection_Blocking.ino
+++ b/examples/SX126x/SX126x_Channel_Activity_Detection_Blocking/SX126x_Channel_Activity_Detection_Blocking.ino
@@ -43,7 +43,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Channel_Activity_Detection_Interrupt/SX126x_Channel_Activity_Detection_Interrupt.ino
+++ b/examples/SX126x/SX126x_Channel_Activity_Detection_Interrupt/SX126x_Channel_Activity_Detection_Interrupt.ino
@@ -53,7 +53,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Channel_Activity_Detection_Receive/SX126x_Channel_Activity_Detection_Receive.ino
+++ b/examples/SX126x/SX126x_Channel_Activity_Detection_Receive/SX126x_Channel_Activity_Detection_Receive.ino
@@ -58,7 +58,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_FSK_Modem/SX126x_FSK_Modem.ino
+++ b/examples/SX126x/SX126x_FSK_Modem/SX126x_FSK_Modem.ino
@@ -39,7 +39,7 @@ void setup() {
 
   // initialize SX1262 FSK modem with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.beginFSK();
+  int state = radio.beginFSK({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
@@ -50,8 +50,8 @@ void setup() {
 
   // if needed, you can switch between LoRa and FSK modes
   //
-  // radio.begin()       start LoRa mode (and disable FSK)
-  // radio.beginFSK()    start FSK mode (and disable LoRa)
+  // radio.begin({})       start LoRa mode (and disable FSK)
+  // radio.beginFSK({})    start FSK mode (and disable LoRa)
 
   // the following settings can also
   // be modified at run-time
--- a/examples/SX126x/SX126x_PingPong/SX126x_PingPong.ino
+++ b/examples/SX126x/SX126x_PingPong/SX126x_PingPong.ino
@@ -59,7 +59,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Receive_Blocking/SX126x_Receive_Blocking.ino
+++ b/examples/SX126x/SX126x_Receive_Blocking/SX126x_Receive_Blocking.ino
@@ -48,7 +48,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Receive_Interrupt/SX126x_Receive_Interrupt.ino
+++ b/examples/SX126x/SX126x_Receive_Interrupt/SX126x_Receive_Interrupt.ino
@@ -59,7 +59,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Spectrum_Scan/SX126x_Spectrum_Scan.ino
+++ b/examples/SX126x/SX126x_Spectrum_Scan/SX126x_Spectrum_Scan.ino
@@ -47,7 +47,7 @@ void setup() {
 
    // initialize SX1262 FSK modem with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.beginFSK();
+  int state = radio.beginFSK({});
   if(state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Transmit_Blocking/SX126x_Transmit_Blocking.ino
+++ b/examples/SX126x/SX126x_Transmit_Blocking/SX126x_Transmit_Blocking.ino
@@ -43,7 +43,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/examples/SX126x/SX126x_Transmit_Interrupt/SX126x_Transmit_Interrupt.ino
+++ b/examples/SX126x/SX126x_Transmit_Interrupt/SX126x_Transmit_Interrupt.ino
@@ -58,7 +58,7 @@ void setup() {
 
   // initialize SX1262 with default settings
   Serial.print(F("[SX1262] Initializing ... "));
-  int state = radio.begin();
+  int state = radio.begin({});
   if (state == RADIOLIB_ERR_NONE) {
     Serial.println(F("success!"));
   } else {
--- a/extras/test/SX126x/main.cpp
+++ b/extras/test/SX126x/main.cpp
@@ -13,7 +13,7 @@ SX1261 radio = new Module(hal, 7, 17, 22, RADIOLIB_NC);
 int main(int argc, char** argv) {
   int state = RADIOLIB_ERR_UNKNOWN;
 
-  state = radio.begin();
+  state = radio.begin({});
   printf("[SX1261] Test:begin() = %d\n", state);
   RADIOLIB_TEST_ASSERT(state);
 
--- a/src/modules/SX126x/SX1262.cpp
+++ b/src/modules/SX126x/SX1262.cpp
@@ -7,64 +7,115 @@ SX1262::SX1262(Module* mod) : SX126x(mod) {
   chipType = RADIOLIB_SX1262_CHIP_TYPE;
 }
 
+int16_t SX1262::begin(const Configuration_t& config) {
+  // execute common part
+  int16_t state = SX126x::begin(
+    config.codingRate, config.syncWord, config.preambleLength,
+    config.tcxoVoltage, config.useRegulatorLDO);
+  RADIOLIB_ASSERT(state);
+
+  // configure publicly accessible settings
+  state = setSpreadingFactor(config.spreadingFactor);
+  RADIOLIB_ASSERT(state);
+
+  state = setBandwidth(config.bandwidth);
+  RADIOLIB_ASSERT(state);
+
+  state = setFrequency(config.frequency);
+  RADIOLIB_ASSERT(state);
+
+  state = SX126x::fixPaClamping();
+  RADIOLIB_ASSERT(state);
+
+  state = setOutputPower(config.power);
+  RADIOLIB_ASSERT(state);
+
+  return(state);
+}
+
+// deprecated
 int16_t SX1262::begin(float freq, float bw, uint8_t sf, uint8_t cr, uint8_t syncWord, int8_t power, uint16_t preambleLength, float tcxoVoltage, bool useRegulatorLDO) {
+  // C++11 does not support designated initializers here, so we need to do this the old way
+  Configuration_t config;
+  config.frequency = freq;
+  config.bandwidth = bw;
+  config.spreadingFactor = sf;
+  config.codingRate = cr;
+  config.syncWord = syncWord;
+  config.power = power;
+  config.preambleLength = preambleLength;
+  config.tcxoVoltage = tcxoVoltage;
+  config.useRegulatorLDO = useRegulatorLDO;
+  return(begin(config));
+}
+
+int16_t SX1262::beginFSK(const ConfigurationFSK_t& config) {
   // execute common part
-  int16_t state = SX126x::begin(cr, syncWord, preambleLength, tcxoVoltage, useRegulatorLDO);
+  int16_t state = SX126x::beginFSK(
+    config.bitRate, config.frequencyDeviation, config.receiverBandwidth,
+    config.preambleLength, config.tcxoVoltage, config.useRegulatorLDO);
   RADIOLIB_ASSERT(state);
 
   // configure publicly accessible settings
-  state = setSpreadingFactor(sf);
-  RADIOLIB_ASSERT(state);
-
-  state = setBandwidth(bw);
-  RADIOLIB_ASSERT(state);
-
-  state = setFrequency(freq);
+  state = setFrequency(config.frequency);
   RADIOLIB_ASSERT(state);
 
   state = SX126x::fixPaClamping();
   RADIOLIB_ASSERT(state);
 
-  state = setOutputPower(power);
+  state = setOutputPower(config.power);
   RADIOLIB_ASSERT(state);
 
   return(state);
 }
 
+// deprecated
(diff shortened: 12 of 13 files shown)


cleishm:
Hmm... build checks currently fail. Works in my environment, but seems the build environment doesn't support brace-enclosed initializer lists. What C++ level does RadioLib target?

I could explicitly initialize the `SX1262::Configuration_t`, but that is less elegant a replacement to the long argument `::begin(...)` methods than allowing initialization inline.

cleishm:
Build checks now pass. Unfortunately, C++11 doesn't handle designated initializers for non-aggregate structs (C++14 and above does) and RadioLib is supporting C++11.

The code now uses C++11 syntax inside RadioLib, but I have left the examples using >=C++14 syntax.

jgromes:
The current `begin*` methods are indeed a bit clumsy. In one case, it even lead to having a dedicated member variable of the `SX126x` class just to control the presence/absence of the TCXO crystal which some radios use and must be configured correctly on startup, otherwise the radio will fail.

The main reason we are sticking to C++11 (or even trying to avoid more advanced parts of C++ in general) is that the build environments we target are ... varied, to say the least. The Arduino ecosystem is a bit of a mess in this, as it pretty much gives free reign to manufacturers of MCUs to have whatever C++ version they see fit - hence our CI which builds the library for all of them.

However, this is also a major change to the API, no doubts about that. I also worry somewhat about the performance hit: for the `SX126x_Transmit_Blocking` example, this increases the size from [39852 bytes of flash](https://github.com/jgromes/RadioLib/actions/runs/17864611897/job/50803786743#step:9:1312) to [40258 bytes](https://github.com/jgromes/RadioLib/actions/runs/17869443730/job/50819638171#step:9:1312), for other platform it might be more. We can run the CI for all of them to check this.

Overall, it is a good UX improvement, but it definitely needs to be rolled out gradually - so backwards compatible changes only and no change to the way examples are done. We can showcase this ability via the settings example only.

I would also be interested in @StevenCellist's and @GUVWAF's opinion on the matter.

GUVWAF:
On the one hand I agree with this. For example, a long time ago, after a RadioLib update we were accidentally setting `currentLimit` to the `preambleLength` parameter since the former was removed. Took us a while to find this out, and with this approach it would have been avoided.

On the other hand it’s indeed a rather big change and the increase in flash size is also a downside. I wonder if the latter can be decreased by passing the struct by reference? 

Furthermore, it might be nice to re-use the `DataRate_t` and `PacketConfig_t` structs (or actually, their modem-specific structs) inside the configuration.

cleishm:
Thanks for taking a look, and appreciate the thought and discussion. Glad to know that the concern that prompted this idea (the long arg list) is a shared one. And I'm happy to make any changes you'd like to suggest experimenting with.

I also appreciate the need to target c++11, it makes sense. Fortunately the designated-initializer syntax is just pretty sugar to avoid having to name the configuration structure, which makes for a cleaner looking invocation.

> For example, a long time ago, after a RadioLib update we were accidentally setting `currentLimit` to the `preambleLength` parameter since the former was removed.

Ouch! I was wondering how often that got confused. I admit to doing it a few times myself using the library (e.g. setting the power level instead of the preamble length!).

> I wonder if the latter can be decreased by passing the struct by reference?

I believe they are passed by reference already?

> this increases the size from [39852 bytes of flash](https://github.com/jgromes/RadioLib/actions/runs/17864611897/job/50803786743#step:9:1312) to [40258 bytes](https://github.com/jgromes/RadioLib/actions/runs/17869443730/job/50819638171#step:9:1312), for other platform it might be more

I hadn't considered the impact on flash size, and honestly I don't know how problematic a 408byte increase would be. I'm actually surprised it increases it by that much - the configuration structs aren't that big - and curious where it is going to. Would love some thoughts or insights if you have them.

GUVWAF:
> I believe they are passed by reference already?

Oh, yes, you're right.

> I admit to doing it a few times myself using the library (e.g. setting the power level instead of the preamble length!).

If you really want to be safe on this, you'd need to use designated initialization though, so it's still a problem if the user would use assigned initialization.

StevenCellist:
My main struggle with the proposed implementation here is that it is such a non-standard way of doing things. I started out with RadioLib a little over two years ago without any knowledge of radios and RadioLib, but it was well-written, well-documented and had a lot of defaults that meant I didn't need to understand everything to get going.
With the struct / named field list here, I think we raise the minimum skill threshold substantially, lose default values and in my opinion GCC 11 is the sweet spot for balance between ease of use, ease of understanding & being powerful.


> I'm actually surprised it increases it by that much - the configuration structs aren't that big - and curious where it is going to. Would love some thoughts or insights if you have them.

My endeavour last year creating 'DietLib' (RadioLib on a diet) showed that involving anything that is not plain variables adds 'a lot' of overhead as the compiler has difficulty optimizing it. The `PhysicalLayer` in RadioLib for instance prevents the compiler from optimizing away unused functions. Using additional structs also sucks (yesterday's PR for LoRaWAN showed exactly that). Meh.

If we analyse the usecase, I can only really see a need for this in ESP32-deepsleep world. Don't get me wrong, I'm a huge ESP32 lover and use it myself including deepsleep, but breaking the core function of the library for a specific niche is a bit rough.

The idea from @GUVWAF could be something. Somewhat of a balance between old and new. But maybe it could be an idea to look at the LoRaWAN implementation: have a fully working stack that expects to live in RAM, no ESP32-deepsleep. But provide an additional interface for such a usecase that leaves the rest of the library intact. An idea could be to have a sort of `export()` function that creates such a struct as you propose from all cached variables or read from registers. And then a sort of `import()` function that takes a struct and re-instates all variables without reconfiguring the radio etc. Up to the user to ensure that the radio is not modified in between. With such an implementation, `begin()` can be left as is, keeping it easy for the newbies around here.

Curious to hear from others!

cleishm:
Thanks for the thoughts, @StevenCellist!

> If we analyse the usecase, I can only really see a need for this in ESP32-deepsleep world.

I want to be clear though, this suggested change is independent of the deep sleep PR. I just felt that the long argument list was already clumsy, and when I added one more it felt too much. Even without adding one more though, I _feels_ like something that could be improved.

> But provide an additional interface for such a usecase that leaves the rest of the library intact. An idea could be to have a sort of export() function that creates such a struct as you propose from all cached variables or read from registers. And then a sort of import() function that takes a struct and re-instates all variables without reconfiguring the radio etc.

This is a discussion for the deep sleep PR, so I'll quote it over there. Of course, if you'd prefer to discuss here, then please bring it back again.

> With the struct / named field list here, I think we raise the minimum skill threshold substantially, lose default values and in my opinion GCC 11 is the sweet spot for balance between ease of use, ease of understanding & being powerful.

Fortunately, this code works with C++11 also, just more verbose syntax:

```
SX1262::Configuration_t config;
config.frequency = 915.0;
config.spreadingFactor = 10;
config.codingRate = 7;
config.preambleLength = 12;
int state = radio.begin(config);
```

This is clearly _way_ more verbose that the existing approach (`radio.begin(915.0, 125.0, 10, 7, 0x12, 10, 12);`) but has the advantage of being much clearer in intent, and allows the user to skip explicitly stating default values if they only want to change something later in the list. For example, setting preamble length to something different without having to specify the sync word and tx power level like I did just now.

And users who are able to opt into later language versions (e.g. most ESP32 Arduino code seems to default to C++17) get to use the simpler syntax.

cleishm:
> I hadn't considered the impact on flash size, and honestly I don't know how problematic a 408byte increase would be. I'm actually surprised it increases it by that much - the configuration structs aren't that big - and curious where it is going to.

Ok, I did a little investigating around flash size. I found that, for a simple example built for the esp32s3, the flash size went from 383447 to 383823, a 376 byte increase.

However, when I removed the current `::begin` methods and left just the versions that take a config struct, the flash dropped back down to 383455, *only an 8 bytes change*.

So almost all of the increase is coming from the duplication of the `::begin` methods. The older ones would be marked deprecated, but I can't image they could be removed anytime soon.

jgromes:
The flash increase is not that much of an issue, but it is something to keep in mind. It is actualy a bit less than I thought. I think the points raised here in favor of implementing this are all valid, the points against are mainly that it is a bit more advanced syntax, and that it does break the public API. The latter can be handled by gradual rollout - first a backwards compatible change, so the struct-based `begin` methods would exist alongside the original ones, then a depracation warning when the user tries to use the old one, and then finally removal of the original methods in a new major release (8.0.0). This would have to be communicated in advance to the major projects using RadioLib (e.g. Meshtastic).

Usage of more advanced C++ concepts is actually one that has been kind of baked into RadioLib from day one. Back then I started by drafting the public API as simple as possible (specifically instantiation of the various classes, e .g. `SX1262 radio = new Module(...);` and kept iterating until I found a way to implement it in C++, which turned out to be a converting constructor (which back then I had no idea was a thing). And as far as I recall, there were never any questions even from new users about this construction, I guess it just feels natural.

So to sum up, I think there's a couple of points to do here:

1. Backwards compatible changes - the original begin method has to remain for now, and we will have to take the penalty in the increased flash usage.
2. The configuration structures have to be common to all modules - all of them have [the same defaults](https://github.com/jgromes/RadioLib/wiki/Default-configuration), and we have to keep it like that without copy-pasting those structs all over the place. The module-specific configuration like TCXO voltage for SX126x has to be added in some way, but it seems solvable with inheritance.
3. Even post-8.0.0, I would like to keep the option of having `radio.begin()` without curly braces, simply because that's how most people use it.

StevenCellist:
> as far as I recall, there were never any questions even from new users about this construction, I guess it just feels natural.

@HeadBoffin and me have spent a fair few hours scratching our heads trying to find out how the module inheritance actually works! But visually it is easy to follow.

Regarding the argument list: is it an idea to decompose the arguments into three groups, namely Packet related (preamble, header, crc, cr etc), Modulation related (sf, bw, freq, iq, txpower etc) and Hardware related (tcxo, do a reset etc)? 
I'd personally prefer that over one big list as it helps learning all the different bells and whistles of a radio. But interested to hear what others prefer.

cleishm:
> So to sum up, I think there's a couple of points to do here:
> 
> 1. Backwards compatible changes - the original begin method has to remain for now, and we will have to take the penalty in the increased flash usage.
> 2. The configuration structures have to be common to all modules - all of them have [the same defaults](https://github.com/jgromes/RadioLib/wiki/Default-configuration), and we have to keep it like that without copy-pasting those structs all over the place. The module-specific configuration like TCXO voltage for SX126x has to be added in some way, but it seems solvable with inheritance.

I'm happy to experiment with that if you'd like me to.

> 3. Even post-8.0.0, I would like to keep the option of having `radio.begin()` without curly braces, simply because that's how most people use it.

Agreed. Easily solvable by adding a no-arg variant.

cleishm:
> simply because that's how most people use it.

Random aside - the defaults specify a specific frequency (434.0Mhz for most modules), but my reading suggests that almost all regions have a primary alternative that is not 434.0Mhz and the LoRa Alliance seems to be treating EU433 as legacy. So it does surprise me that most people use the library without specifying the frequency. Are you confident in that assessment? I'm not arguing against a no-arg constructor. Just curious.

jgromes:
@cleishm what I see in the issues/discussions is that a lot of people will call `begin` without arguments and then set the things they need via separate methods, e.g. `setFrequency` - but still, I would like it to remain visually the same. So definitely, people are not using it on the default frequency, just not using the `begin` method to set it - which is an argument in favor of reworking it.

Regarding the default frequency value - 434 MHz is an ISM band in EU, I had to choose one so I chose this (AFAIK there is no globally universal ISM band). The LoRaWAN stack configures the radio frequency based on LoRaWAN regional parameters, so this is opaque to the user.

HeadBoffin:
> @HeadBoffin and me have spent a fair few hours scratching our heads trying to find out how the module inheritance actually works! But visually it is easy to follow.

But not something most IDE's can unravel for you, so jumping to the final function to get an overview of what you are calling when debugging at user level isn't possible. CJH's LW lib used by Lacuna has taken that to an art form.

If I've missed some way of tracing where the call ended up, that would be super useful.

But to the matter at hand:

Splitting configuration out in to a struct adds clarity rather than multiple variants of the begin - which, as noted, is easiest used by calling the vanilla version and then using the setter functions. The struct also acts as an easy reference point for the user code to know what has been requested. [I'd put 'digital twin' in here, but I'm not that cool]

But how to handle the missing but required values? 

Some sorts of validation scheme that fills in the blanks? This becomes data hiding but as we call by reference, those missing values can be back-filled. 

Or my preference, an initialisation function that takes the struct and fills in sensible defaults (and why not 433MHz, I use it often enough) and then the user can set what they want. This would save having a block of code where the user has to setup every single struct member when many which they aren't familiar with. 

Maybe even a diff as well, React style, to set the ~DOM~ registers that have changed (kidding).


This should then assist with moving to not re-initialising the radio issue as there is a record of what state it should be in, the user code can do things like retrieve any received packets, travel hopeful with the radio state or just ask the driver to set the radio as per the struct.

The only thing I'd ask is that there are no nested structs - both Bosch & Analog use them for their drivers and it feels unnecessarily complicated to create & setup, even if they are all on the same screen/ page.

jgromes:
> The only thing I'd ask is that there are no nested structs

@HeadBoffin if not by composition, then by inheritance, but we will definitely need some mechanism in order to avoid duplicating the common arguments of `begin`. I imagine something along the lines of  the below - most of the building blocks like `DataRate_t` and `PacketConfig_t` are implemented already.

```c++
// PhysicalLayer.h:
struct RadioConfig_t {
  ModemType_t modem; // LoRa, FSK etc.
  float freq; // carrier frequency in MHz
  DataRate_t dr; // SF/BW/CR for LoRa
  PacketConfig_t pc; // preamble length etc.
  int16_t txPwr; // output power in dBm
};

// SX126x.h:
struct SX126xConfig: public RadioConfig_t {
  float tcxoVoltage; // and whichever other SX126x-only settings are needed
};

(...)

int16_t SX1262::begin(struct SX126xConfig * cfg) {
 if(cfg->modem == RADIOLIB_MODEM_LORA) {
    setCodingRate(cfg->dr.lora.codingRate);
    (...)
  }
}
```

HeadBoffin:
The issue / challenge I see is that you setup a struct with the values you need and then are required to stick that in to another struct which inevitably leads to sticking that in to another struct. And then users heads explode.

Where Analog & Bosch excel is in holding references to function callbacks and HAL implementation in their structs. A level of indirection similar to the RL Module system!

The ideal would be for the user-programmer to populate a fairly flat list of settings. Perhaps a helper lib can be used to then process that in to the structures for RL.

cleishm:
I did a little playing around, and while the idea of a hierarchy of configuration structures seems like it would help to avoid repetition, it makes the whole construct much more unwieldy. As soon as you create `struct SX126xConfig: public RadioConfig_t` (as an example) you no longer have control over the order of members and designated initialization doesn't work. The documentation also becomes much harder to follow, as the parameters a user may be interested in are scattered over the hierarchy.

I will also note that the duplication you're looking to avoid exists now. It's just hidden in the arguments to the begin methods, where they must be duplicated.

So I feel there's two choices. Accept duplication in these structs and look at them like an expanded alternative to the long argument lists - or just stick with the long argument lists and forget this idea :)

Actually, there is a third option, but it's probably a much bigger change. As @jgromes explained, a common way to work with the library is:

```
radio.begin();
radio.setFrequency(915.0);
radio.setCodingRate(7);
... etc
```

The downside to this is that the module is configured first with defaults, then updated. If we could support setting parameters before applying them to the module, that would provide a nice API and avoid that downside. E.g.:

```
radio.setFrequency(915.0);
radio.setCodingRate(7);
... etc
radio.begin();
```

cleishm:
> The downside to this is that the module is configured first with defaults, then updated.

@jgromes I admit I'm just speculating that this is a downside. Perhaps it's not really an issue at all.

In which case, the even easier option would be to simply deprecate the long argument list begin methods and encourage users to just do `begin()` then set the various options they want after (which you noted many do already).

jgromes:
Had some more time to try out different approaches, namely inheritance, composition and hiding the common stuff with macros (the last one being without a doubt the worst one of all, as it leads to a lot of obfuscation ... but wanted to try it out just for the sake of completeness).

<details><summary>Having the structure composed makes the call to begin convoluted to say the very least ...</summary>
<p>

```c++
// PhysicalLayer.h
struct RadioConfig_t {
  ModemType_t modem; // LoRa, FSK etc.
  float freq; // carrier frequency in MHz
  DataRate_t dr; // SF/BW/CR for LoRa
  PacketConfig_t pc; // preamble length etc.
  int16_t txPwr; // output power in dBm
};

// SX126x.h
struct SX126xConfig_t {
  RadioConfig_t phy;
  float tcxoVoltage; // and whichever other SX126x-only settings are needed
};

// SX1262.cpp
int16_t SX1262::begin(const SX126xConfig_t & cfg) {
  if(cfg.phy.modem == RADIOLIB_MODEM_LORA) {
    setCodingRate(cfg.phy.dr.lora.codingRate);
  }
  // etc.
  return(RADIOLIB_ERR_NONE);
}

// main.cpp
int state = radio.begin({
    .phy = {
      .modem = RADIOLIB_MODEM_LORA,
      .freq = 434.0f,
      .dr = {
        .lora = {
          .spreadingFactor = 9,
          .bandwidth = 125,
          .codingRate = 7,
        }
      },
      .pc = {
        .lora = {
          .preambleLength = 16,
          .implicitHeader = false,
          .crcEnabled = true,
          .ldrOptimize = true,
        }
      },
      .txPwr = 10,
    },
    .tcxoVoltage = 1.6,
  });
```

</p>
</details> 

But I feel that this approach can be adjusted to actually help us get rid of not just the massive list of `begin` arguments, but also the multiple variants of `beginXXX` methods - the information which modem to configure can be part of the configuration structure. And setting defaults can be done by initialization, either by dedicated function or by the constructor of the struct itself. And it lends itself to inheritance quite well:

```c++
// PhysicalLayer.h
struct RadioConfig_t {
  ModemType_t modem = RADIOLIB_MODEM_NONE; // LoRa, FSK etc.
  float frequency = 434; // carrier frequency in MHz
  DataRate_t dataRate = { 0 }; // SF/BW/CR for LoRa
  PacketConfig_t packetConfig = { 0 }; // preamble length etc.
  int16_t outputPower = 10; // output power in dBm

  RadioConfig_t(ModemType_t m) {
    switch(m) {
      case(RADIOLIB_MODEM_LORA):
        if(!this->dataRate.lora.spreadingFactor) { this->dataRate.lora.spreadingFactor = 12; };
    }
  };
};

// SX126x.h
struct SX126xConfig_t: public RadioConfig_t {
  float tcxoVoltage = 1.6; // and whichever other SX126x-only settings are needed
  SX126xConfig_t(ModemType_t m): RadioConfig_t(m) { };
};

// SX1262.cpp
int16_t SX1262::begin(const SX126xConfig_t* cfg) {
  if(cfg->modem == RADIOLIB_MODEM_LORA) {
    setSpreadingFactor(cfg->dataRate.lora.spreadingFactor);
  }
  // etc.
  return(RADIOLIB_ERR_NONE);
}

// main.cpp
SX126xConfig_t cfg(RADIOLIB_MODEM_LORA); // set everything to default values for LoRa
cfg.dataRate.lora.spreadingFactor = 9; // change to a non-default value
int state = radio.begin(&cfg);
```

Of course, this will make named arguments as @cleishm originally proposed them impossible, because extra logic is needed to configure the default values based on the selected modem. But to me it accomplishes the same goal, which I understand to be making the configuration more verbose, allowing the user to modify one of the parameters without having to write out all of them, and making it less prone to silently failing in future updates. Thoughts?

PS: I do realize I am overthinking what was originally a rather straightforward change. But since the cat of "breaking the basic part of the API" is out of the proverbial bag, let's start breaking ;)

cleishm:
> ```
> // main.cpp
> SX126xConfig_t cfg(RADIOLIB_MODEM_LORA); // set everything to default values for LoRa
> cfg.dataRate.lora.spreadingFactor = 9; // change to a non-default value
> int state = radio.begin(&cfg);
> ```

Nice, @jgromes. I must say, however, that the above is pretty confusing from the user's perspective. Now they have to know the structure of that SX126xConfig_t in order to change a common thing like SF.

Are you trying to make it so that `begin(cfg)` works for all modem types, e.g. LORA and FSK, vs having separate `begin(cfg)` and `beginFSK(cfg)` methods?

Also, we can probably close this PR and move this to an issue/discussion if you prefer. I don't think the code as it stands now is what we'd land on (it was just an initial proposal to start a discussion anyway).

jgromes:
@cleishm 

> the above is pretty confusing from the user's perspective. Now they have to know the structure of that SX126xConfig_t in order to change a common thing like SF

That is true, nevertheless, I think there are two different types of users we are trying to appeal to: the newbies who are just starting with radios or microcontrollers/electronics in general, and the more advanced users / big projects who can leverage the different abstractions we have. IMO the newbie can still use the dedicated `setSpreadingFactor` method, if they are intimidated by the multi-layered config. The advanced user can switch everything with a single method call.

> Are you trying to make it so that begin(cfg) works for all modem types

Exactly that. Getting rid of the extra begin methods will make it a lot easier to switch modems (currently done in the rather hacky `setModem` method), and might free up some memory (admittedly I have no numbers to support this claim yet).

> Also, we can probably close this PR and move this to an issue/discussion if you prefer

I agree, #1620 is the place.

PR #1611, opened 2025-09-22
[SX126x] Update doc of SX126x::getPacketLength

cleishm:
Add a note that the update parameter is not used for SX126x modules.


jgromes:
Indeed, a copy-paste from the SX127x. Merged, thank you for the contribution!

PR #1653, opened 2025-11-26
Fix `receive` `timeout` parameter docs

jrpear:
Hello,

The `timeout` parameter of `recieve` for the SX128x modules actually defines the timeout in **microseconds**, not milliseconds. See the implementation: https://github.com/jgromes/RadioLib/blob/4bba92f382aa7482e23d587b95ba6b4ecdf6e58e/src/modules/SX128x/SX128x.cpp#L298.

This PR fixes that in the function docstring.


jgromes:
Good catch - merged, thank you for the contribution!

PR #1661, opened 2025-12-17
Add a destructor function to RadioLibHal

Vinx911:
Add a destructor function to RadioLibHal

Solves this error:
error: deleting object of polymorphic class type 'Esp32S3Hal' which has non-virtual destructor might cause undefined behavior [-Werror=delete-non-virtual-dtor]

```
class Esp32S3Hal : public RadioLibHal {
    // ..................
}

Esp32S3Hal* hal = new Esp32S3Hal();
delete hal;
```


jgromes:
Looking good, and it even provoked a hidden bug in the unit test :) Merged, thank you for the contribution!

Issue #1628, opened 2025-10-17
Add PA Optimal Settings configuration option for SX1262?


Quency-D:
I saw in the manual that different optimal settings have different energy consumption, and that lower energy consumption options can be configured for certain areas where high output power isn't needed. However, the current code seems to only have one fixed setting. Is it possible to add all four options shown below?

```c
  state = SX126x::setPaConfig(0x04, RADIOLIB_SX126X_PA_CONFIG_SX1262);

```

<img width="1394" height="513" alt="Image" src="https://github.com/user-attachments/assets/d0bcecba-edd0-414a-a601-62f1fbfd9968" />

<img width="1322" height="660" alt="Image" src="https://github.com/user-attachments/assets/0fe2b6cf-7703-4d41-8ef1-923f2228593a" />

jgromes:
> the current code seems to only have one fixed setting

That is not really accurate - the `setPaConfig` method is public, so you can set any combination of parameters - with the caveat that it needs to be called after setting the Tx power, and that the method allows you complete control over the PA, so if you set incorrect values (not specified by datasheet) then you risk damaging the module.

Quency-D:
> That is not really accurate - the `setPaConfig` method is public, so you can set any combination of parameters

Sorry, that wasn't quite accurate, but what do you think of putting the four configuration options in the manual directly into the code and then choosing one from the four?

jgromes:
The problem is that the user manual does not specify how to set a different output power for optimized consumption, other than the values listed. For example, let's say the user wants to set  +9 dBm - the optimal settings in the datasheet do not include this output power level. When writing the SX126x driver, I chose to ensure accurate power level at the cost of efficiency, because I did not have a way to externally verify the output power.

The datasheet also isn't very transparent as to the impact the different settings have. So trying to find the most optimal solution for each supported power level seems a bit risky, since the library is deployed in extremely varied environments.

Overall it might be worth a bit of time to investigate though.

Quency-D:
OK, I initially thought there were only four fixed options. Now after reading the documentation carefully, I've discovered there are many other configuration options. I'll take a closer look.

jgromes:
[...]

PR #1662, opened 2025-12-17
[SX126x] Add optimized PA configuration table for SX1262 and SX1268

jgromes:
As described in #1628, this PR adds the following:

* Public method `SX126x::setOutputPower(int8_t power, uint8_t paDutyCycle, uint8_t hpMax, uint8_t deviceSel)` which gives users lower-level control over the PA
* Lookup tables with optimized PA configuration, saving 25 - 30% power consumption during transmission. The optimization is enabled by default, but can be disabled by setting the second argument of `setOutputPower` to false.

### DIFF
--- a/src/modules/SX126x/SX1262.cpp
+++ b/src/modules/SX126x/SX1262.cpp
@@ -111,24 +111,11 @@ int16_t SX1262::setOutputPower(int8_t power, bool optimize) {
   int16_t state = checkOutputPower(power, NULL);
   RADIOLIB_ASSERT(state);
 
-  // get current OCP configuration
-  uint8_t ocp = 0;
-  state = readRegister(RADIOLIB_SX126X_REG_OCP_CONFIGURATION, &ocp, 1);
-  RADIOLIB_ASSERT(state);
-
   // set PA config
+  int8_t paVal = optimize ? this->paOptTable[power + 9].paVal : power;
   uint8_t paDutyCycle = optimize ? this->paOptTable[power + 9].paDutyCycle : 0x04;
   uint8_t hpMax = optimize ? this->paOptTable[power + 9].hpMax : 0x07;
-  state = SX126x::setPaConfig(paDutyCycle, RADIOLIB_SX126X_PA_CONFIG_SX1262, hpMax);
-  RADIOLIB_ASSERT(state);
-
-  // set output power with default 200us ramp
-  int8_t paVal = optimize ? this->paOptTable[power + 9].paVal : power;
-  state = SX126x::setTxParams(paVal, RADIOLIB_SX126X_PA_RAMP_200U);
-  RADIOLIB_ASSERT(state);
-
-  // restore OCP configuration
-  return(writeRegister(RADIOLIB_SX126X_REG_OCP_CONFIGURATION, &ocp, 1));
+  return(SX126x::setOutputPower(paVal, paDutyCycle, hpMax, RADIOLIB_SX126X_PA_CONFIG_SX1262));
 }
 
 int16_t SX1262::checkOutputPower(int8_t power, int8_t* clipped) {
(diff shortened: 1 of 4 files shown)


serialrf433:
It is interesting to see how about every measured, optimized dBm value rised a small fraction while the power draw got reduced.

The unoptimized values are more or less linear. The optimized have some anomalies between -1dBm and 5dBm.

https://github.com/radiolib-org/power-tests/blob/main/logs/compare_sx1262/rlb_2025_12_14_084954.log

jgromes:
> optimized dBm value rised a small fraction while the power draw got reduced

@serialrf433 that is the result of the optimization algorithm: https://github.com/radiolib-org/power-tests/blob/71e2b9e361886192c0794365dfc3af42784a43b1/py/process.py#L78

I identify the "correct" output value not by the absolute level, but by comparing with the output of the unoptimized version for that level. This is to guarantee that our updated configuration always produces the same or slightly better result than the unoptimized version. For example, the real output for an unoptimized version of +15 dBm is approximately +14.84 dBm. The algorithm searches for settings that produce at least +14.84 dBm or more, and from those it selects the one with the smallest power draw.

> The optimized have some anomalies between -1dBm and 5dBm.

I noticed those too, but they seem to be aligned with the unoptmized samples. For example, notice how the sample below drops for both unoptimized and optimized version. Because of this I think it is some error in the measurement (and also why during the verification I measured each point 10 times, as can be seen in the sources).

       -1,                     -0.01,          3.31,       2.89,      28.94,     95.68                      0.20,          3.31,       2.03,      20.34,     67.29
       -1,                     -0.19,          3.31,       2.89,      28.92,     95.63                     -0.15,          3.31,       2.03,      20.34,     67.27
       -1,                     -0.00,          3.31,       2.89,      28.91,     95.58                      0.20,          3.31,       2.03,      20.36,     67.35

jgromes:
I have repeated the measurements for three additiona SX1262 devices, and the results are pretty good. There seems to be some impact of the optimization in terms of linearity and spread of the output power across different devices, but it is pretty small, and we correct for the spread by optimizing towards slighty higher output power anyway.

All data I have measured can be found in https://github.com/radiolib-org/power-tests, summary is in the charts below. Granted, 4 devices is not that many, but these tests are pretty time-consuming to run even when automated. Also since @StevenCellist reported no impact to performance in testing of his trackers, I will go ahead and call this good enough and unleash it upon the world.

Screenshot_175
Screenshot_176
Screenshot_177

serialrf433:
Thanks for presenting the data this way and even repeating the tests it again. This is great research about the PA settings on the SX126x.
Its sad that Semtech do not release such graphs on their own and have many known false information in their datasheets they do not seem to care to fix.

radiohound:
Yes, this is excellent information. Thank you!

jgromes:
Thanks! To be fair to Semtech, the PA configuration interface is shown in the datasheet, including some sample values. But it definitely could be improved a lot. The best thing would be to know what the different parameters (hpMax, paDutyCycle and deviceSel) actually do. But that would require at least a functional description of the PA, which might expose their intellectual property, so I somewhat see the logic of not including it in the documentation (though I don't agree with it).

The next best thing would be to show measurements and list the possible impacts. But since our data is available publicly in the above linked repository, I invite everyone to examine it and process further. And I will definitely be looking into implementing this optimization for other Semtech devices that allow PA configuration (such as LR11xx series).

serialrf433:
What about the SX127x/RFM9x the LoRa idea have start with? https://github.com/radiolib-org/power-tests could get a public documentation how the efficiency evolved over time at the Semtech products. I would be interested if and how far the SX127x/RM9 could be optimized compared to the SX126x

jgromes:
@serialrf433 the SX127x does not allow as much control over the PA as the newer SX126x, so there's nothing to optimize there unfortunately.

serialrf433:
Good to know. Your answer here would maybe help others with same question also not knowing that there is no optimization possible on the SX127x.
Looking forward then at LR11xx series and sometime later probably LoRa Gen4 series like the LR2021.

PR #1671, opened 2026-01-06
LR11x0 LoRa BWs per SWSD003 and NiceRF LoRa1121F33-xxx examples

UT2UH:
This PR adds all available LR11x0 LoRa BWs per SWSD003 both to sub-GHz and HF bands, add band dependent PA duty cycles and NiceRF LoRa1121F33-1G9/2G4 examples.

### DIFF
--- a/examples/LR11x0/LoRa1121F33-1G9_PingPong/LoRa1121F33-1G9_PingPong.ino
+++ b/examples/LR11x0/LoRa1121F33-1G9_PingPong/LoRa1121F33-1G9_PingPong.ino
@@ -1,0 +1,192 @@
+/*
+   RadioLib NiceRF LoRa1121F33-1G9 Ping-Pong Example
+
+   For the module description, see the manufacturer page
+   https://www.nicerf.com/lora-module/lora-1121f33-1g9.html
+
+   For default module settings, see the wiki page
+   https://github.com/jgromes/RadioLib/wiki/Default-configuration#lr11x0---lora-modem
+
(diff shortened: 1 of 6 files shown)


StevenCellist:
I'm not Mr. Gromes the Great so can't/won't comment on the driver part. 
What is the reason to include a very specific example for the NiceRF F33 module?

UT2UH:
@StevenCellist These modules were the only LR1121 modules I had to test the updated code so I included test examples just in case I was wrong and somebody else take care to fix them as well...

jgromes:
The examples are way too specific to be included here, sorry. If we had an example for every possible radio module we support, there would be thousands of them.

Next, regarding the bandwidths - the values below 62.5 kHz will be added as part of an ongoing work on supporting the LR2021 radio. So this will be reworked anyway in the near future.

And finally, the changes that were done to the PA configuration don't seem to match the LR1120 datasheet. Was this tested and the output power measured for all levels?

So overall this PR contains three unrelated changes; two of them seem unnecessary and one seems very strange/not well justified.

UT2UH:
@jgromes Hope to get LR2021 rework soon, thank you!

UT2UH:
@jgromes  PA Configuration changes were based on this recommendation - [https://github.com/Lora-net/SWDR001/blob/master/src/lr11xx_radio.h#L613](https://github.com/Lora-net/SWDR001/blob/master/src/lr11xx_radio.h#L613). Output power for all levels was not measured.

Issue #1667, opened 2025-12-21
RPi Pico HAL issue


smunaut:
The HAL provided for the RPi pico uses `gpio_set_irq_enabled_with_callback` to set callbacks.
But this sets a _shared_ GPIO handler for all the GPIO.

See https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_gpio_1ga6165f07f4b619dd08ea6dc97d069e78a : 

"""All GPIOs/events added in this way on the same core share the same callback;"""

So when you try to setup DIO0 and DIO1 action to different methods, this doesn't work and only the latest one you setup is called for both which is ... surprising.

I ran into that trying to do interrupt based CAD. In the end, I used a single call back and test the value of DIO0/DIO1 in the callback to differentiate but it's something that's not obvious.

StevenCellist:
That same documentation says that `gpio_add_raw_irq_handler()` is available to use independent callbacks on different GPIO. Have you tried using that instead? (Or can you try now?)

smunaut:
I didn't try no.

Because then it's the user responsibility to deal with acknowledging the IRQ which is another unexpected thing and it was easier for me to read the pin values in a platform independent way.

jgromes:
@smunaut I think what @StevenCellist meant was whether you could try to add this function to the call in `PicoHal::attachInterrupt` and test whether it resolves the issue, since we don't have the hardware and the set up, It would help us greatly. I agree this is most likely not intended behavior, but as I have nearly zero experience with RPi Pico, this slipped during review. Thanks!

smunaut:
@jgromes Yes, I understood.

However what I'm pointing out is that this won't work unless I also modify my callback functions to do something like:

```
    if (gpio_get_irq_event_mask(my_gpio_num) & my_gpio_event_mask) {
       gpio_acknowledge_irq(my_gpio_num, my_gpio_event_mask);
      // handle the IRQ
    }
```

inside my callbacks because the "raw" exception handler are expected to handle acknowledging the interrupt (else it's just called again ...)

And needing to do this in the callback is just as unexpected for a RadioLib callback as the current behavior and also prevents writing code that runs on different platform since you need platform specific stuff in the callback.

AFAICT the only way to properly handle it to expose the same kind of behavior as other platforms would be to have wrappers in the HAL for the interrupts. Those would handle the IRQ acknowledge and call the right user callback.

[...]

PR #1681, opened 2026-01-26
Re-Fix for #1667

smunaut:
The short story is that the raw handlers are not associated with
the mask given, the handler itself will be added to a linked link
by the generic IRQ code, so adding the same handler several time for
different interrupts yields not what you expect. The current code
also doesn't remove the handler before setting a new one which leads
to `assert` crashing the CPU.

The new strategy here is that once a pin is used with the HAL, we
always keep the shared HAL handler bound to it. Internally we keep
a list of what pin map to which user handler. When a new pin is used
we temporarely disable interrupt, unbind the HAL handler completely
and re-bind it with the new set of pin. And then re-enable the GPIO
interrupts. AFAICT from reading the SDK code, that's the only option.

To avoid having to do that whole thing too often (it's a rather large
block of code with interrupts disabled), we never remove our raw handler
once it's attached to a pin, we just remove the user handler from our
internal list. I don't think that's a problem since those pins where
the user would use the HAL are pins from the radio module and so are
managed by the HAL.

PS: Also a minor typo fix ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -356,7 +356,7 @@ jobs:
           mkdir -p ~/rpi-pico
           cd ~/rpi-pico
           git clone https://github.com/raspberrypi/pico-sdk.git
-          cd pico-sdk && git checkout 1.5.1
+          cd pico-sdk && git checkout 2.2.0
       
       - name: Build the example
         run: |


jgromes:
@smunaut thanks for the additional fix! However, the CI is crashing when trying to build it, my guess is that the SDK version in our pipeline (1.5.1) does not include the 64-bit versions of the functions. If you have a specific version that you have tested with, we can switch the CI to that one.

smunaut:
I'm using 2.2.0 here. I switched to the 64b version because the HAL was defining the max pico gpio to 48 so obviously 32b was not enough :sweat_smile:

jgromes:
OK, in that case you can change the checked-out version here: https://github.com/jgromes/RadioLib/blob/master/.github/workflows/main.yml#L359

jgromes:
@smunaut had some time so I fixed the CI and merged this. Thank you for the contribution!

PR #1684, opened 2026-01-31
[LR2021] add getDataRate() method missing

lyusupov:


jgromes:
Hmmm ... I'm thinking whether we need to keep those methods at all. To be honest I have not actually used them for the past couple of years. And if this information is something that the user really needs to have, they can always use `startTransmit` and measure it themselves. What do you think?

lyusupov:
`getDataRate()` is still a member of numerous RadioLib examples -

https://github.com/search?q=repo%3Ajgromes%2FRadioLib+Serial.print%28radio.getDataRate%28%29&type=code

jgromes:
That's true, but it can be easily removed from there - it's not adding much value to be honest.

lyusupov:
It is upon your discretion...

I may close this PR if the getDataRate() is scheduled by you for removal from other radio targets.

jgromes:
Yes, let's close this one and I will remove the method in the upstream. I actually wanted to do that for some time, but forgot - thanks for reminding me.

PR #1685, opened 2026-01-31
[LR2021] DIO5 can only do DIO_SLEEP_PULL_UP

lyusupov:
image


jgromes:
Ah yes, I did implement something similar to the initialization function since the hardware I have does actually happen to use DIO5 as the IRQ.

Merged, thank you!

PR #1690, opened 2026-02-01
[LR2021] set max limit of GFSK/OOK bit rate in accordance with datasheet

lyusupov:
Current value of the limit looks like a copy-paste from LR11x0.cpp

These are the limits specified in the datasheet -

image


jgromes:
Good catch - since a lot of the methods are copy pasted from LR11x0 class, there is bound to be a few more bugs like that.

Issue #1693, opened 2026-02-01
RAK3172 crashing at begin()


olivierouellet:
**Setup**
Module : RAK3172 (STM32WL) with TCXO
Protocol : LoRaWAN, US915 band
Library version : 7.5.0
PlatformIO ST STM32 version: 19.4.0

**Bug**
When I first started to use a RAK3172 module with RadioLib, I wasn't able to initialize my RAK3172 board using `begin()` . It would always crash.

<details><summary>Debug mode output</summary>
<p>

```
[0.025276] RLB_SPI: CMDW
[0.025420] RLB_SPI: SI
[0.025535] RLB_SPI: SO
[1.027007] RLB_DBG: GPIO pre-transfer timeout, is it connected?
[1.030733] RLB_DBG: -705 at .pio/libdeps/lora_release/RadioLib/src/Module.cpp:279
[2.036006] RLB_DBG: GPIO pre-transfer timeout, is it connected?
[2.037993] RLB_DBG: -705 at .pio/libdeps/lora_release/RadioLib/src/Module.cpp:251
[3.049006] RLB_DBG: GPIO pre-transfer timeout, is it connected?
[3.054743] RLB_DBG: -705 at .pio/libdeps/lora_release/RadioLib/src/Module.cpp:279
radio.begin() starting
[3.064488] RLB_DBG: 
RadioLib Info
Version:  "7.5.0.0"
Platform: "Arduino STM32 (official)"
"Feb  1 2026" "14:18:49"
[3.071049] RLB_SPI: CMDW        0
[3.077520] RLB_SPI: SI
[3.077669] RLB_SPI: SO  80
```

</p>
</details> 

Thinking it might be a clock drift problem, I (wrongly) measured my drift using 2 printf statements and it gave me 9 ms of drift for a period of 1000 ms (which is far away from the 2 ppm in the data sheet, but this is not the right way to measure a drift). So I added `-D RADIOLIB_CLOCK_DRIFT_MS=9` to platformio.ini and suddenly everything worked perfect. Perfect until [StevenCellist](https://github.com/StevenCellist) told me [I should not use it](https://github.com/jgromes/RadioLib/issues/1679#issuecomment-3800217998).

`RADIOLIB_CLOCK_DRIFT_MS` in **ArduinoHal.cpp** modifies :
- `ArduinoHal::delay`
- `ArduinoHal::delayMicroseconds`
- `ArduinoHal::millis`
- `ArduinoHal::micros`

[...]

PR #1696, opened 2026-02-05
Missing initialization in for loop of SPI debug

olivierouellet:
There are 2 inits missing in for() loops in the SPI debug part of `Module::SPItransferStream()` .

The second added init might be causing problems. It think it was caused by a copy and paste of the first one (line 426), so let's fix also this to prevent another error in the future. 

Related to #1693

### DIFF
--- a/src/Module.cpp
+++ b/src/Module.cpp
@@ -423,7 +423,7 @@ int16_t Module::SPItransferStream(const uint8_t* cmd, uint8_t cmdLen, bool write
       RADIOLIB_DEBUG_SPI_PRINT_NOTAG("R\t");
     }
     size_t n = 0;
-    for(; n < cmdLen; n++) {
+    for(n = 0; n < cmdLen; n++) {
       RADIOLIB_DEBUG_SPI_PRINT_NOTAG("%X\t", cmd[n]);
     }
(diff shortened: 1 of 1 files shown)


jgromes:
Both of these are intentional, though not really documented, so I see how that could lead to some confusion. In the first case, the variable is actually initialized a line above. That's just because it needs to be declared there as well, to have it's scope contain the subsequent for loops. It is then used as counter for printing out the outgoing data, which is composed of two parts - the command and the data bytes sent from MCU. We first print the command from the `cmd` buffer, and then keep on printing the data buffer. The reason we don't se the variable to 0 is that would repeat the command bytes which are present at the start of `buffOut`.

Doing it like this allows to separate command from the data in the debug printout, and it is a lot easier to read that way.

olivierouellet:
Ok, thanks for your answer. I thought the second init missing was not intentional and could cause problems.

Would you like me to add a comment for the second init missing (the first one is easy to understand). Otherwise I'll close this PR.

jgromes:
I will do that as the original author if this logic, and we can close the PR. Nevertheless, thanks for letting me know about this, so that we can clarify it!

PR #1697, opened 2026-02-05
Fix for float not supported on newlib-nano

olivierouellet:
newlib-nano library (used on some Arduino platforms) does not support %f format specifier in printf by default (as specified in [BuildOpt.h](https://github.com/jgromes/RadioLib/blob/4f3d2abcbcb234603a05f192ee0714e3ce363720/src/BuildOpt.h#L499) )

That's causing a bug (bunch of random characters and timing issues) on STM32 when using `RADIOLIB_DEBUG_PROTOCOL` . It may apply to other platforms too.

Related to #1693

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -2503,8 +2503,9 @@ bool LoRaWANNode::execMacCommand(uint8_t cid, uint8_t* optIn, uint8_t lenIn, uin
       uint8_t macDrMax = (optIn[4] & 0xF0) >> 4;
       uint8_t macDrMin = optIn[4] & 0x0F;
       
-      RADIOLIB_DEBUG_PROTOCOL_PRINTLN("NewChannelReq: index = %d, freq = %7.3f MHz, DR %d-%d", 
-                                      macChIndex, macFreq / 10000.0, macDrMin, macDrMax);
+      RADIOLIB_DEBUG_PROTOCOL_PRINT("NewChannelReq: index = %d, freq = ", macChIndex);
+      RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT_NOTAG((double)macFreq / 10000.0, 3);
+      RADIOLIB_DEBUG_PROTOCOL_PRINTLN_NOTAG(" MHz, DR %d-%d", macDrMin, macDrMax);
 
       uint8_t drAck = 0;
       uint8_t freqAck = 0;
@@ -2578,7 +2579,9 @@ bool LoRaWANNode::execMacCommand(uint8_t cid, uint8_t* optIn, uint8_t lenIn, uin
       // get the configuration
       uint8_t macChIndex = optIn[0];
       uint32_t macFreq = LoRaWANNode::ntoh<uint32_t>(&optIn[1], 3);
-      RADIOLIB_DEBUG_PROTOCOL_PRINTLN("DlChannelReq: index = %d, freq = %7.3f MHz", macChIndex, macFreq / 10000.0);
+      RADIOLIB_DEBUG_PROTOCOL_PRINT("DlChannelReq: index = %d, freq = ", macChIndex);
+      RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT_NOTAG((double)macFreq / 10000.0, 3);
+      RADIOLIB_DEBUG_PROTOCOL_PRINTLN_NOTAG("  MHz");
       uint8_t freqDlAck = 0;
       uint8_t freqUlAck = 0;
       


lyusupov:
> ... newlib-nano library (used on some Arduino platforms) does not support %f format ...

IMHO, this is **not an issue of the RadioLib**...

In most cases, even though **default** Arduino setting is to use nano-lib with no FP - there is typically an option for user to enable build with nanolib-FP through the Core menu or command line (rtlib=nanofp).

image

jgromes:
We actually have a macro for this, `RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT(VAL, DECIMALS)`, which on Arduino platforms will call `Serial.print(float, int)`, and form there it is up to the Arduino core how it handles printing the floats. So if using floats is absolutely necessary, it can be done that way. Though in this case I would say doing it via integers works too, we don't risk losing information. IIRC, LoRaWAN frequency channel resolution is limited in MAC commands.

StevenCellist:
Using floats is absolutely necessary for me to keep my sanity. I get an immediate stroke reading 868300000 Hz. *Faints from typing message.*

olivierouellet:
I can change it for this (untested, but will later if accepted) :
```
  float frequency = chnl->freq / 10000.0;
  int frequencyIntegral = (int) frequency;
  int frequencyDecimal = (int) ((frequency - frequencyIntegral) * 1000);
  (void) frequencyIntegral;
  (void) frequencyDecimal;
  
  RADIOLIB_DEBUG_PROTOCOL_PRINTLN_NOTAG("");
  RADIOLIB_DEBUG_PROTOCOL_PRINT("Frequency = ");
  RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT(frequency, frequencyDecimal)
  RADIOLIB_DEBUG_PROTOCOL_PRINTLN(" MHz, TX = %d dBm", pwr);
  state = this->phyLayer->setFrequency(frequency);
  RADIOLIB_ASSERT(state);
```
I'm opened to other suggestions as well. I see float are used in other places, so I'll fix these when we'll agree on the best way to handle it.

> Using floats is absolutely necessary for me to keep my sanity. I get an immediate stroke reading 868300000 Hz. Faints from typing message.

Ha ha. That's part of my daily job, so I'll bring you back. Help me with code and we'll try to avoid that!

StevenCellist:
We don't need to calculate the integral and decimal part. See [the reference](https://docs.arduino.cc/language-reference/en/functions/communication/serial/print/#:~:text=Serial.print(1.23456%2C%204,gives%20%221.2346%22 ) for printing floats the way Mr Gromes proposed.
So we would call `PRINT_FLOAT(frequency, 3)`.

olivierouellet:
I made a new version using `RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT`.

I had to make a few changes in BuildOpt.h
- Changed `RADIOLIB_DEBUG_PRINT_FLOAT` to `RADIOLIB_DEBUG_PRINT_FLOAT_LVL` to print floats with LEVEL (and kept `RADIOLIB_DEBUG_PRINT_FLOAT` to print floats without LEVEL) and corrected macros using it already
-  Added `RADIOLIB_DEBUG_BASIC_PRINT_FLOAT_NOTAG`, `RADIOLIB_DEBUG_PROTOCOL_PRINT_FLOAT_NOTAG` and `RADIOLIB_DEBUG_SPI_PRINT_FLOAT_NOTAG`
- It seems like I changed lots of thing but it's mostly because I realigned the macros

My serial output used to fill with random characters once in a while before the fix. Now it seems ok. I'll agree that code is less legible though (although it should't cause anybody to faint, hopefully).

I didn't check if _newlib-nano + float printf_ adds only a small amount of memory. I'm opened to gets rid of this commit, but I think it would be better to have a way to tell users they should enable _newlib-nano + float printf_ right from the logs when debugging.

Thanks for all your suggestions and feel free to ask me other changes.

olivierouellet:
I removed some spaces (added for alignment) in BuildOpt.h, so the changes are easier to see now.

It takes 8.4 KB to add float to newlib-nano on my setup (STM32WL using stm32duino). The test was done using `-D PIO_FRAMEWORK_ARDUINO_NANOLIB_FLOAT_PRINTF` in the build_flags of platformio.ini file.

I can check if there's floats used elsewhere in the rest of the code once we'll settle on something.

olivierouellet:
Floats are also printed is in [Si443x.cpp](https://github.com/jgromes/RadioLib/blob/35e9c3910ac137d795be513a20a429f6bdcff51e/src/modules/Si443x/Si443x.cpp#L804) but it's already using `RADIOLIB_DEBUG_BASIC_PRINT_FLOAT`.

And it's also used in [LoRaWAN_TS009.h](https://github.com/jgromes/RadioLib/blob/35e9c3910ac137d795be513a20a429f6bdcff51e/examples/LoRaWAN/LoRaWAN_TS_Packages/LoRaWAN_TS009.h#L186) but I don't know if that should be fixed (will certification have to be done again if there's any change?) .

I did not see any other floats printed in the rest of the code.

StevenCellist:
I am working on a full revamp of Technical Specifications and will include this when time comes. Since I don't expect many people to do certification by themselves, we can leave that for now.
@jgromes the rest is up to your preferences.

jgromes:
I think this can be merged - @olivierouellet thank you for the contribution!

On a side note I will most likely aim to rework the usage of floats in the future, since we are using them in palces where it is note strictly necessary.

PR #1698, opened 2026-02-05
Potential overflow fix for RADIOLIB_SPI_PARANOID

olivierouellet:
`checkInterval * 1000` could overflow on platforms using 16 bits int.

Casting to RadioLibTime_t prevents the overflow.


jgromes:
Good catch, thanks! Though this overflow is probably not present anywhere at the moment, as it would require calling this internal method with check interval set to 32 ms or higher, which I don't think is the case.

Still, good to fix it, so thank you for the contribution!

PR #1702, opened 2026-02-09
add setEncoding(RADIOLIB_ENCODING_MANCHESTER_INV) into ADS-B receive example

lyusupov:


jgromes:
@lyusupov once again, you are commiting into a branch that has a draft pull request attached to it. This means that the branch is work in progress and changes are to be expected - including this example. It is not final, at the moment I don't know what parts of the configuration will remain there and what parts will into the ADSBClient class or even to LR2021 class. In any case, merging this will only create a commit in the history that will later change. Hence, not merging.

And on a side note, have you run this? It shouldn't be needed, as encoding is set in `ADSBClient::begin`. In my testing setup, I am not explicitly setting encoding from the main file and it does receive the packets.

PR #1704, opened 2026-02-13
[lLoRaWAN] Fix incorrect DR calculation for fixed band in createSession()

Kodo-kakaku:
**Fix incorrect DR calculation for FIXED band in createSession()**

In LoRaWANNode::createSession(), the data rate calculation for
FIXED band type incorrectly uses drMin twice instead of drMin + drMax.

This causes the initial uplink data rate to be set to approximately
drMin instead of the intended average between drMin and drMax,
which is inconsistent with the DYNAMIC band calculation logic.

```
Changed:
  - (drMin + drMin + 1) / 2
To:
  - (drMin + drMax + 1) / 2
```

This ensures symmetric behavior between DYNAMIC and FIXED band types
and correct initial data rate selection.


StevenCellist:
Hi @Kodo-kakaku, thank you very much for the bug fix and contribution! This only surfaces in ABP which isn't commonly used and if no manual datarate is configured. But still very much a mistake, thanks for the PR!

PR #1707, opened 2026-02-17
[Crypto] Add streaming functionality to CMAC

StevenCellist:
Generating the AES128-CMAC of large chunks of data currently creates huge internal buffers, see:

https://github.com/jgromes/RadioLib/blob/1d68c7635eeb44b61179df5c62b46e7ca5c13fb6/src/utils/Cryptography.cpp#L58

Moreover, the current `generateCMAC()` requires a single contiguous buffer as input.

This PR adds functionality to stream the input in unrelated segments of arbitrary size. This is necessary for the implementation of LoRaWAN TS004 Fragmented Data Block Transport where the MIC has to be verified of data blocks that may reach hundreds of kilobytes. The first argument of the necessary CMAC calculation is an 'unrelated' block of 16 bytes:

image

Concatenating the `msg` to `B0` is virtually impossible, and writing `B0` to the 16 bytes before `msg` doesn't seem like a reasonable option to me.

### DIFF
--- a/src/TypeDef.h
+++ b/src/TypeDef.h
@@ -608,12 +608,6 @@
 */
 #define RADIOLIB_ERR_INVALID_MODE                               (-1121)
 
-/*!
-  \brief There was not enough data to continue with CMAC calculation.
-  Call the method again once RADIOLIB_AES128_BLOCK_SIZE or more bytes are available.
-*/
-#define RADIOLIB_ERR_CMAC_NOT_ENOUGH_DATA                        (-1122)
-
 // LR11x0-specific status codes
 
 /*!
--- a/src/utils/Cryptography.cpp
+++ b/src/utils/Cryptography.cpp
@@ -65,9 +65,9 @@ void RadioLibAES128::initCMAC(RadioLibCmacState* st) {
   st->subkeys_generated = false;
 }
 
-int16_t RadioLibAES128::updateCMAC(RadioLibCmacState* st, const uint8_t* data, size_t len) {
+void RadioLibAES128::updateCMAC(RadioLibCmacState* st, const uint8_t* data, size_t len) {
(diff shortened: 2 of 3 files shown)


jgromes:
Overall it looks good, though I do have two remarks:

1. I think the return type should be changed to `int16_t`. Previously in `generateCMAC`, there was only a single return path through the method, hence it not returning anything. But now there are checks of the global state, so we should return some information - at least for `updateCMAC` which may not perform an update if there is not enough input data.
2. I think we should add units tests here - yes I know they weren't here previously ... but we can improve it ;) It should be relatively simple to use the [test vectors from the RFC](https://www.rfc-editor.org/rfc/rfc4493.html#section-4) and test both the streamed methods, as well as the original `generateCMAC`.

StevenCellist:
Thanks for reviewing! Any chance you can integrate those points, please? I haven't touched the unit testing in RadioLib yet and all my brain cells are trying to handle the FUOTA specs.

jgromes:
@StevenCellist I added the test ... and it fails on two of the test vectors :D In master it passes*, so this could be some regression and we should investigate.

*(passes as in the vectors pass - there is a crash which is due to one of the test vectors being a zero length, which is not well handled in the master. But the vectors taht are affected in this branch pass there)

StevenCellist:
This is very useful, thanks for implementing! I'll have a look at it soon.

StevenCellist:
@jgromes the implementation on master does not do any null-pointer checks. If we remove those from the new implementation, we can make it void again. (Because I removed the return code about too little CMAC data which is not actually an error.) What's your preference?

jgromes:
Agree, I will make that change and then merge this (and then figure out why the CI is being so uncooperative ...)

PR #1714, opened 2026-02-26
[LR11x0] Expose instantaneous RSSI measurement

GUVWAF:
The function `getRssiInst()` is used for the LR11x0 to get an instantaneous RSSI measurement, but it is protected. This exposes it by overloading `getRSSI()` with the `bool packet` parameter added just like is done for SX127x, SX126x, SX128x and LR2021.

I also rephrased *Recorded* Signal Strength Indicator to *Received* Signal Strength Indicator as I believe that's the more common term.

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -1111,14 +1111,13 @@ float LR11x0::getRSSI() {
 
 float LR11x0::getRSSI(bool packet, bool skipReceive) {
   float val = 0;
-  int16_t state;
 
   // check if RSSI of packet is requested
   if (packet) {
     val = getRSSI();
   } else {
     if(!skipReceive) { (void)startReceive(); }
-    state = getRssiInst(&val);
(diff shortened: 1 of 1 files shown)


jgromes:
Thanks for the contribution! In testing, does this return the some reasonable numbers? I am asking because on the other devices (most recently on LR2021), I had to make sure Rx mode was active, otherwise the return value was stuck at -128 dBm.

GUVWAF:
Yes, I get reasonable numbers, but indeed I just checked that's only when we're in receive mode. Do you want me to add the  `bool skipReceive` parameter just like for LR2021 and let it go to receive mode if that value is false?

jgromes:
I think that would be the best option, that way we will keep the same interface for all the LoRa modules.

GUVWAF:
Added in the latest commit, following the same approach as the LR2021. 
I also slightly adjusted the description for this parameter for the LR2021.

jgromes:
Looking good now - merged, thank you for the contribution!

PR #1716, opened 2026-02-28
permit build with Arduino Core for Nordic Semi nRF54

lyusupov:


StevenCellist:
I don't think we should support a build platform that is currently in v0.1, targets a single board, exists for a week, has 2 stars and has such fundamental issues as you yourself reported in your issue on the repository. At that point we could "support" a thousand platforms and end up with so many complaints here.

lyusupov:
>  has such fundamental issues as you yourself reported in your issue on the repository

Yes, there is an issue with Strings component but it does not affect the **RadioLib** use case. I've built a few  of **RadioLib** examples successfully.

1) Nordic **nRF54**L15 SoC is a primary component of [**LR2021 LoRa Plus Evaluation Kit**](https://www.cnx-software.com/2026/02/16/lr2021-lora-plus-board-combines-semtech-lr2021-and-nordic-nrf54l15-for-high-speed-flrc-and-lora-connectivity/) developed by Seeed Studio in collaboration with Semtech Corp ;
    https://wiki.seeedstudio.com/semtech_lr2021_evk_getting_started/
2) the Core is light weight variant of another nRF54 Core for Zephyr OS https://github.com/lolren/NRF54L15-Arduino-core 
   The Zephyr Core for nRF54 is a bit more mature but needs 9 GB of disk space for deployment
3) even though the `Clean` Core looks new, it has one positive feature - it works on most recent Nordic nRF54 Series and it is the only one Core available for nRF54 at this moment ;
4) The BuildOpt.h contains a list of Cores of questionable quality and purpose, such as
 * Intel Curie
 * MegaCoreX
 * Apollo3
 * CubeCell

but as long as they all work with RadioLib and need no time to maintain - I believe it is Ok to keep them in the list,

jgromes:
I agree with @StevenCellist, listing official suport for this core seems premature. However, there is nothing specific for that core apart from the std lib IO header - so @lyusupov what's preventing you from using it with unmodified RadioLib? It will only raise a warning that this platform is not suported, that's it.

lyusupov:
> what's preventing you from using it with unmodified RadioLib?

image

<br>
<br>

missing 

```
 #include <cstdio>
```

breaks build of [ADSB_Receive](https://github.com/jgromes/RadioLib/tree/master/examples/ADSB/ADSB_Receive) example sketch with the nRF54 Core . 
This happens because the [ADSB.cpp](https://github.com/jgromes/RadioLib/blob/master/src/protocols/ADSB/ADSB.cpp) uses a call to `snprintf()` with no previous declaration.
https://github.com/jgromes/RadioLib/blob/master/src/protocols/ADSB/ADSB.cpp#L75

jgromes:
@lyusupov then the actual problem is missing `#include <stdio.h>` in ADSB.cpp - can you try that instead?

Issue #521, opened 2022-05-19
Receive sensitivity from low power devices is improved by a proximate higher power source


vtjballeng:
We have multiple devices using Dorji SX1262T modules. Here is an example of the problem we're seeing.

We'll use an example with 3 devices. Device 1 & 2 are located far apart. Device 1 is broadcasting a periodic message once a second. Device 2 will not receive this broadcast message from Device 1. If we turn on Device 3, which also sends a periodic broadcast message, then Device 2 will receive the message from Device 3 and also begin to receive messages from Device 1. 

FREQUENCY_CENTER_TRACK_STATE, 915, 
LORA_SF7, 
LORA_BW_500, 
LORA_CR_4_5, 
LDRO_AUTO

Messages are ~35bytes and we see no DIO1 pin activity when Device 3 is off but Device 1 is on at Device 2. All devices have Rx boosted gain active.

jgromes:
Do you have a record of the RSSI and SNR values? Especially the SNR would be quite interesting to know, since the 500 kHz/SF7 combination is quite fast (as LoRa goes), but the required SNR on the receiving end is somewhat higher because of it. If the SNR is at the limit (or close to it), then the demodulator might have trouble locking onto the preamble. Maybe it could be worthwile to try transmitting with SF8 (which works in lower SNR than SF7), at teh same distance, transmitted power, etc., to see if the problem persists.

vtjballeng:
@jgromes , we have some recordings but we might run the simulation again to check for SNR specifically. I'll say though that we aren't talking about a general reception problem. 

This is a mode we're seeing the transceivers fall into after some period of time where they stop seeing distant devices they could previously see. Like they've dropped some receive sensitivity. We can then poke them back to full sensitivity again with a high power proximate source message and they suddenly begin to see the distant device messages again.

With respect to the problem we're seeing, why would a higher power proximate source (device 3) message improve the preamble locking for the device that is further away (device 1)?

jgromes:
Higher power would help, but that would increase both RSSI and SNR. Whereas by using a higher SF, the signal strength at the receiver would remain the same, and only the SNR limit would be increased.

That's why it would be interesting to see the SNR/RSSI values throughout the time when it issue appears - it would be interesting to see if the values before and after this issue appears are different.

jgromes:
Possibly related: https://forum.lora-developers.semtech.com/t/sx1268-problem-in-rx-continuous-mode/1122

klemzi:
[...]

PR #1718, opened 2026-03-01
Implement proper AGC reset for SX126x, LR11x0 & SX1276

weebl2000:
Might be good to have this in RadioLib, based on these PRs:

- https://github.com/meshcore-dev/MeshCore/pull/1743
- https://github.com/meshtastic/firmware/pull/9705

For SX126x & LR11x0: reset analog registers, for SX1276: sleeping should be sufficient. I based this off the SX1262 datasheet & user manuals.

relates to #521

### DIFF
--- a/src/modules/LR11x0/LR11x0.h
+++ b/src/modules/LR11x0/LR11x0.h
@@ -931,7 +931,6 @@ class LR11x0: public LRxxxx {
 #endif
     uint8_t wifiScanMode = 0;
     bool gnss = false;
-
     int16_t modSetup(float tcxoVoltage, uint8_t modem);
     bool findChip(uint8_t ver);
     int16_t config(uint8_t modem);
--- a/src/modules/SX126x/SX126x.h
+++ b/src/modules/SX126x/SX126x.h
@@ -185,6 +185,22 @@ class SX126x: public PhysicalLayer {
     int16_t scanChannel(const ChannelScanConfig_t &config) override;
 
     /*!
+      \brief Reset the AGC gain state by performing a warm sleep, recalibration, and
+      image rejection calibration cycle. Re-applies DIO2 RF switch and RX boosted gain
+      settings if previously configured. Leaves the radio in standby mode.
+      \returns \ref status_codes
+    */
+    int16_t resetAGC();
+
+    /*!
+      \brief Perform calibration of the specified blocks.
+      \param params Calibration parameters - bitfield of RADIOLIB_SX126X_CALIBRATE_* values.
+      Use RADIOLIB_SX126X_CALIBRATE_ALL to calibrate all blocks.
+      \returns \ref status_codes
+    */
+    int16_t calibrate(uint8_t params);
+
+    /*!
       \brief Sets the module to sleep mode. To wake the device up, call standby().
       Overload with warm start enabled for PhysicalLayer compatibility.
       \returns \ref status_codes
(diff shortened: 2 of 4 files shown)


jgromes:
Thank you, this will be a very useful contribution!

I know the issue has been reported on SX126x - are there some reports this is also happening on LR11xx too, sicne this PR also contains reset method for those radios? And if so, is there a test that could be run on LR2021 to figure if it is happening there as well?

weebl2000:
> Thank you, this will be a very useful contribution!
> 
> I know the issue has been reported on SX126x - are there some reports this is also happening on LR11xx too, sicne this PR also contains reset method for those radios? And if so, is there a test that could be run on LR2021 to figure if it is happening there as well?

No reports on LR11x0 that his happens there too. I feel like the LR11x0 is a little more sophisticated with the AGC than the SX126x.

But I dove into the docs and LR1110 is largely done in the same way as the SX1262, so I added it for anyone that would like to try it. It could definitely help when an LR1110 is exposed to big interference say at like a cell site.

UPDATE: LR2021 is largely the same, implemented there too.

jgromes:
Thanks, I think should be safe to merge - however, before doing that, I will try to run some tests to see if this issue actually manifests on the other radios from Semtech, because we only have it confirmed on the SX126x series so far. So if it doesn't actually manifest there it doesn't make that much sense to be including the AGC reset methods for the other radios IMO. And if it does manifest, then we will know for a fact whether AGC reset resolves it or not.

weebl2000:
> Thanks, I think should be safe to merge - however, before doing that, I will try to run some tests to see if this issue actually manifests on the other radios from Semtech, because we only have it confirmed on the SX126x series so far. So if it doesn't actually manifest there it doesn't make that much sense to be including the AGC reset methods for the other radios IMO. And if it does manifest, then we will know for a fact whether AGC reset resolves it or not.

I'm fairly certain the LR1110 is prone to the same issue, but it's just seen/reported less because it's mainly integrated in more "mobile" applications where power and form factor are most important (hence sub-ghz, GNSS and WiFi in 1 chip) . Whereas the SX1262 is a workhorse in many boards that are also used as repeaters etc.

Most people aren't running repeaters on their LR1110. The calibration reset is valid though, would be good having it in I guess for those few users who might need it.

weebl2000:
Got my hands on a T1000E tracker with LR1110. Tried with periodic agc reset and can confirm it works fine. Not as dramatic as with Heltec v4 and SX1262 and a GC1109 LNA, but I can definitely see the AGC reset is triggered correctly. Noise floor bounces briefly, goes back to where it was fairly quickly.

jgromes:
After trying for a few days I was finally able to reproduce the issue with stuck AGC on SX1262. It seems to be heavily dependent on the output power level of the transmitter - I was only able to get the issue to manifest with good antennas, transmitting at +22 dBm. Frequency offset does not appear that critical, I was able to get it to break with as much as 8 MHz between transmitter and receiver.

The AGC does seem to work reliably though. I will try to also reproduce this on the other radios and test that as well.

weebl2000:
@jgromes good work. Yes, biggest issues are reported with SX1262 sitting near large cell sites. These will often have 800 - 900 MHz transmissions at 30-35 dBW EIRP. You can imagine the massive power bleeding on sidebands/harmonics if you're just a few feet away from those sources.

jgromes:
So I tried to do this with the other radios as well, and the SX126x is the only one where this problem manifests - all others (LR1120, LR2021 and SX1272) do not have this problem and after the transmitter is turned off the RSSI returns back to the original values.

In light of that I would propose to only implement this reset for SX126x. That will remove the necessity of having this in the `PhysicalLayer` abstraction.

weebl2000:
> So I tried to do this with the other radios as well, and the SX126x is the only one where this problem manifests - all others (LR1120, LR2021 and SX1272) do not have this problem and after the transmitter is turned off the RSSI returns back to the original values.
> 
> In light of that I would propose to only implement this reset for SX126x. That will remove the necessity of having this in the `PhysicalLayer` abstraction.

done

jgromes:
Looking good now - merged, thank you so much for this great contribution!

PR #1722, opened 2026-03-11
[LR2021] a fix for reverse order of bytes in OOK sync word

lyusupov:
This is a successor of previously applied PR #1686 that was relevant to GFSK code.
The same issue remains in OOK code.
This PR is intended to fix it.


jgromes:
Thanks for the contribution! Looking at the code, we should probably add a check that `syncWordLen <= RADIOLIB_LR2021_OOK_SYNC_WORD_LEN` as the `LR2021::setSyncWord` method will actually allow up to 8 bytes which is not valid for OOK.

lyusupov:
I am not certain of your intention, sorry.
Please, edit this PR on your discretion.

jgromes:
Added the check and merged - thanks for the contribution!

PR #1723, opened 2026-03-11
[LR2021] a fix for OOK packet's RSSI return value

lyusupov:
OOK getRSSI() currently works for instantaneous RSSI value only (when `packet` argument = false).
This PR gives an OOK getRSSI() value for the last received packet (`packet` = true).

### DIFF
--- a/src/modules/LR2021/LR2021.h
+++ b/src/modules/LR2021/LR2021.h
@@ -604,6 +604,7 @@ class LR2021: public LRxxxx {
     /*!
       \brief Gets RSSI (Received Signal Strength Indicator).
       \param packet Whether to read last packet RSSI, or the current value.
+      NOTE: With OOK modem, the "packet" RSSI value is the received power level of the high bits (digital 1).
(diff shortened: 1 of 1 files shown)


jgromes:
Thanks for the contribution! One minor thing, I think this is doing something very slightly different than what is currently written in the doxygen comment:

https://github.com/jgromes/RadioLib/blob/65084d0c5c34988b34cc75949cbe00022807bb2c/src/modules/LR2021/LR2021.h#L604-L611

For OOK, talking about "packet" does not really make sense in the context of RSSI because 0s are a lower level of signal (or nothing at all), so I would add a note to the `\param packet` paremeter that says something like "For OOK, the RSSI is the received power level of the high bits (digital 1)". That's also in line with the correction of the `getOokPacketStatus` method.

lyusupov:
I am not certain of your intention, sorry.
Please, edit this PR on your discretion.

jgromes:
Added the comment and merged - thank you for the contribution!

PR #1727, opened 2026-03-16
Updated CAD detection for LR 2021, resolving the freezing issue

SeeedZhangyh:
While using the `scanChannel()` function in lr2021, it was observed that the program would enter an infinite loop (suspected to be caused by the use of `setCad` without a configured timeout, resulting in the `CadDone` interrupt failing to trigger).

The original implementation of `scanChannel` in lr2021 relied on the `setCadParams` and `setCad` commands; 
however, the datasheet indicates that these are intended for "non-LoRa mode." 
In LoRa mode, the `SetLoraCadParams` and `SetLoraCAD` commands should be used instead.

This Pull Request modifies the specific implementation of `scanChannel` and adds support for default parameter configuration.
The code has been tested and verified to successfully complete the CAD process without freezing.

### DIFF
--- a/src/modules/LR2021/LR2021.cpp
+++ b/src/modules/LR2021/LR2021.cpp
@@ -550,7 +550,7 @@ int16_t LR2021::startChannelScan(const ChannelScanConfig_t &config) {
   RADIOLIB_ASSERT(state);
 
   // set mode to CAD
-  return(startCad(config.cad.symNum, config.cad.detPeak, this->pblAny, this->pnrDelta, config.cad.exitMode, config.cad.timeout));
+  return(startCad(config.cad.symNum, config.cad.detPeak, this->fastCad, config.cad.exitMode, config.cad.timeout));
 }
 
 int16_t LR2021::getChannelScanResult() {
@@ -721,7 +721,7 @@ int16_t LR2021::config(uint8_t modem) {
   return(state);
 }
 
-int16_t LR2021::startCad(uint8_t symbolNum, uint8_t detPeak, uint8_t pblAny, uint8_t pnrDelta, uint8_t exitMode, RadioLibTime_t timeout) {
+int16_t LR2021::startCad(uint8_t symbolNum, uint8_t detPeak, bool fast, uint8_t exitMode, RadioLibTime_t timeout) {
   // check active modem
   uint8_t type = RADIOLIB_LR2021_PACKET_TYPE_NONE;
   int16_t state = getPacketType(&type);
@@ -731,23 +731,21 @@ int16_t LR2021::startCad(uint8_t symbolNum, uint8_t detPeak, uint8_t pblAny, uin
   }
 
   // select CAD parameters
-  //! \TODO: [LR2021] the magic numbers for CAD are based on Semtech examples, this is probably suboptimal
   uint8_t num = symbolNum;
   if(num == RADIOLIB_LR2021_CAD_PARAM_DEFAULT) {
     num = 2;
   }
   
-  // reference values ​​from the datasheet, for num = 2.
+  // reference values ​​from the datasheet for 2 symbols
+  //! \TODO: [LR2021] allow CAD peak detection autoconfiguration
   const uint8_t detPeakValues[8] = { 56, 56, 56, 58, 58, 60, 64, 68 };
   uint8_t peak = detPeak;
   if(peak == RADIOLIB_LR2021_CAD_PARAM_DEFAULT) {
     peak = detPeakValues[this->spreadingFactor - 5];
   }
 
-  bool pbl = (pblAny >= 1 );
-
-  // default don't use acceleration.
-  uint8_t delta = pnrDelta; 
+  // in Fast CAD mode enable acceleration
+  uint8_t pnrDelta = fast ? RADIOLIB_LR2021_LORA_CAD_PNR_DELTA_FAST : RADIOLIB_LR2021_LORA_CAD_PNR_DELTA_STANDARD;
 
   uint8_t mode = exitMode; 
   if(mode == RADIOLIB_LR2021_CAD_PARAM_DEFAULT) {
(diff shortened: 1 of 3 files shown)


jgromes:
It is possible I did use the "wrong" CAD - the datasheet certainly does not make it easier as eveident by my confused comment in the code.

However, the change you have done in `PhysicalLayer` causes problems because you the extended definition of CAD parameters is only applicable to LR2021 - no other modules can use it and it also causes CI to fail/produce new warnings. So `PhysicalLayer` is not the right place for it. Do those extra arguments need to exist? If so, why?

SeeedZhangyh:
Two new parameters—`pblAny` and `pnrDelta` are used to configure `SetLoraCadParams` and relate to "Fast CAD" settings. While it is true that these parameters are not currently utilized by other LoRa modules, I anticipate that future LoRa modules will likely retain them.

Directly modifying the `PhysicalLayer` class is indeed not an ideal approach. I had initially assumed that partial parameter initialization would be supported when the structure is initialized within `startChannelScan`; however, it now appears that this approach presents certain issues. 

A viable alternative would be to directly store these two new configuration values ​​as member variables within the `lr2021` class, provide an interface to modify them, and then utilize them within `SetLoraCadParams`. Nevertheless, directly adding them to the `PhysicalLayer` class might prove to be the most convenient solution.

jgromes:
> directly adding them to the PhysicalLayer class might prove to be the most convenient solution.

I'm not convinced of that, whether those will or will not be part of any future LoRa radios is anybody's guess IMO. Having them as publicly accessible members of `LR2021` class would be more acceptable.

However, looking through the datasheet, I actually don't think these have to be configurable at all.

* `pbl_any` - this one should be able to distinguish between preamble and payload bytes. But the datasheet itself states that 
> even with pbl_any = 1, it is still very likely to detect a payload anyway. Even more so for a small number of symbols, and even more so at low SF, and at high SNR.

So if we expose this parameter to the user, we introduce uncertainty over what was actually detected which I would say defeats the purpose of having this parameter. I would much rather just hardcode this value to 0 and consider full packet detection the only suported feature (like it is on the older series).

* `pnr_delta` - as far as I understood the datasheet, the only possible values here are 0 (for "standard" CAD) and 8 (for "fast" CAD). So a `bool` flag would be enough. But I am not convinced of the utility of fast CAD - we can add it later once there are some real use cases.

So all in all I don't see much utility in having either one of those accessible to the user.

SeeedZhangyh:
Yes, you are right, the `pbl_any` is indeed unnecessary. Set 0 offers greater stability.

Regarding `pnr_delta`, the datasheet specifies a 4-bit configuration but provides only two reference values. Initially, I assumed that assigning different values ​​would result in faster operation and more pronounced effects; however, it now appears that this may have been a misunderstanding. It is possible that these bits are simply reserved by the manufacturer for future configuration purposes.
The datasheet notes:
> This optimization reduces average CAD duration from ~5 symbols to approximately 1.5 symbols when no LoRa signal is present, directly translating to ~70% power consumption reduction for Listen-Before-Talk (LBT) and Carrier Sense Multiple Access (CSMA) applications that perform frequent channel sensing.

so, this setting does play a certain role in reducing power consumption.though I must admit I have not yet tested this in practice.

> So all in all I don't see much utility in having either one of those accessible to the user.

All in all, I can proceed by making the minimal necessary adjustments based on this information.

SeeedZhangyh:
The simplest method is that I just need to modify：
`int16_t LR2021::startCad(uint8_t symbolNum, uint8_t detPeak, uint8_t pblAny, uint8_t pnrDelta, uint8_t exitMode, RadioLibTime_t timeout) `

like this:
```
  // set CAD parameters
  //! \TODO: [LR2021] add configurable exit mode and timeout
  state = setLoRaCadParams(num, 0, 0, mode, timeout_raw, peak);
  RADIOLIB_ASSERT(state);

  // start CAD
  return(setLoRaCad());

```

There is one other matter—unrelated to the CAD aspects here—regarding which I would like to propose adding an interface within LR2021. This would likely make it more convenient to integrate with Meshtastic.

like this:  in LR2021.cpp
```

int16_t LR2021::getLoRaRxHeaderInfo(uint8_t* cr, bool* hasCRC){
  int16_t state;
  state = this->getLoRaPacketStatus(cr, hasCRC, NULL, NULL, NULL, NULL);
  return (state);
}

```
Since Mashtastic utilizes template classes to initialize or invoke APIs, and rl2021 lacks a direct `getLoRaRxHeaderInfo` command, a compilation error would occur. While it is technically possible to handle the `rl2021` class separately via conditional logic, doing so would be overly cumbersome; adding an additional interface would be a much more convenient solution.
 would this be acceptable?
if ok, maybe I can submit along with this. thank you.

StevenCellist:
I would really like the FastCAD to be an option as I'm going to experiment with it for Relay. Given that CAD is about 40% or so of a Relay's power consumption, a fast exit could be a great win. Also untested yet, but really interested in trying this in a couple of weeks.

SeeedZhangyh:
Fine，good luck to you！








 ----------Reply to Message----------
 On Wed, Mar 18, 2026 18:42 PM ***@***.***&gt; wrote:

  StevenCellist left a comment (jgromes/RadioLib#1727)
 
I would really like the FastCAD to be an option as I'm going to experiment with it for Relay. Given that CAD is about 40% or so of a Relay's power consumption, a fast exit could be a great win. Also untested yet, but really interested in trying this in a couple of weeks.

jgromes:
To not block this PR any further, I have pushed the update so that fast CAD can be enabled by a simple flag. If in the future there are more radios that support it, we can add it to the `PhysicalLayer` interface.

@SeeedZhangyh regarding the `getLoRaRxHeaderInfo` method, it can be added, so if you can send it via a separate PR that would be great. I originally thought it was only implemented for `LR11x0` which was why I did not keep the same name for the LR2021 version. But it is also present for SX126x, so it makes sense to have an override.

SeeedZhangyh:
Okay, I already did that.

PR #1728, opened 2026-03-19
[LR2021] Add getLoRaRxHeaderInfo function

SeeedZhangyh:
### Add getLoRaRxHeaderInfo function

Since Mashtastic utilizes template classes to initialize or invoke APIs, and rl2021 lacks a direct getLoRaRxHeaderInfo command, a compilation error would occur. 

While it is technically possible to handle the rl2021 class separately via conditional logic, doing so would be overly cumbersome; adding an additional interface would be a much more convenient solution.


jgromes:
Looking good, thank you! Just a minor thing with the formatting but I will fix that after the merge.

Issue #1730, opened 2026-03-20
Cannot compile LoRaWAN examples on certain stm32 nucleo boards


krabrab:
**Describe the bug**
I am trying to adapt the LoRaWAN_Starter example for the STM32 NUCLEO-L073RZ board with a SX1272MB2xAS shield, connecting to the TTN. Besides adding my keys and regional data (tested both MSB and LSB formats, I am using the EU868 band), I have only changed line 8 of examples/LoRaWAN/LoRaWAN_Starter/config.h as so:
```c
SX1272 radio = new Module(D10, D2, A0, D3);
``` 

Attempting to compile the example using Arduino IDE results in the following error:

```/home/user/Arduino/libraries/RadioLib/src/hal/Arduino/ArduinoHal.cpp: In member function 'virtual void ArduinoHal::init()':
/home/user/Arduino/libraries/RadioLib/src/hal/Arduino/ArduinoHal.cpp:14:3: error: 'dwt_init' was not declared in this scope; did you mean 'uart_init'?
   14 |   dwt_init();
      |   ^~~~~~~~
      |   uart_init
exit status 1

Compilation error: exit status 1
```

Commenting out this line allows compilation. When attempting communication that way I am unable to activate over OTAA (err -1116, no downlink after uplink), ABP authentication sends the packet but doesn't receive a confirmation. No data gets received by the gateway / TTN console, according to it the device is not online. I have verified this behaviour in close proximity to a gateway, I deem hardware failure unlikely. I have verified my pinout matches the library requirements, the SPI header on the shield matches the default SPI interface on the board. I have mapped the Module's `irq` parameter to DIO0 - RxDone/TxDone on the transciever, Module's `gpio` to DIO1 on the transciever. 

This issue is likely related to #1693 and relevant to #549.

**To Reproduce**
Use the LoRaWAN/LoRaWAN_Starter example in the library.

**Expected behavior**
Library is compiled, radio performs the requested operation and the device is found by the network.

**Additional info (please complete):**
 - MCU: NUCLEO-L073RZ
 - Link to Arduino core: https://github.com/stm32duino/Arduino_Core_STM32, version 2.12.0
 - Wireless module type SX1272
 - Arduino IDE version 2.3.8
 - Library version 7.6.0

olivierouellet:
[...]

PR #1731, opened 2026-03-21
[HAL] Check if board is using DWT before calling dwt_init() on STM32

olivierouellet:
On STM32 boards using Cortex-M0, `micros()` is not based on DWT, so calling [dwt_init()](https://github.com/jgromes/RadioLib/blob/631e0910a220b6e69021b0a51bcc83e490b3237d/src/hal/Arduino/ArduinoHal.cpp#L13-L15) in `ArduinoHal::init()` creates a compile error.

Every STM32 board using DWT defines DWT_BASE, so adding a preprocessor check prevents the compile error.

Please note that dwt_init() is usually called by default on STM32duino in [premain()](https://github.com/stm32duino/Arduino_Core_STM32/blob/93d2c963d8cb2dd2953d838a75edc3ad34a665a4/cores/arduino/main.cpp#L25), but it seems it's not called if using your own main() function (linker issue?).

The path is premain() -> init() -> hw_config_init() -> dwt_init()

This PR fixes first part of #1730.

References:
- DWT_BASE is defined in [framework-cmsis](https://github.com/ARM-software/CMSIS_6)
- For Cortex-M3 : [core_cm3.h](https://github.com/ARM-software/CMSIS_6/blob/fdbbc52f24a14ba7ba06a8db9421721a7f604e48/CMSIS/Core/Include/core_cm3.h#L1376)
- For Cortex-M4 : [core_cm4.h](https://github.com/ARM-software/CMSIS_6/blob/fdbbc52f24a14ba7ba06a8db9421721a7f604e48/CMSIS/Core/Include/core_cm4.h#L1552 )

### DIFF
--- a/src/hal/Arduino/ArduinoHal.cpp
+++ b/src/hal/Arduino/ArduinoHal.cpp
@@ -10,7 +10,7 @@ void ArduinoHal::init() {
   if(initInterface) {
     spiBegin();
   }
-  #if defined(ARDUINO_ARCH_STM32)
+  #if defined(ARDUINO_ARCH_STM32) && defined(DWT_BASE)
     dwt_init();
   #endif
 }


jgromes:
Looks good and merged, thank you - most likely we will need a different implementation of `micros` to get these timers working on Cortex-M0 based STM32s.

krabrab:
Is there something stopping us from using the Core-defined micros() and delayMicroseconds() implementation in this case? Please correct me if I am incorrect but I tested the following code using the official STM32duino core on a Cortex M0+ board:
```
uint32_t t1 = micros();
delayMicroseconds(5000);
uint32_t t2 = micros();
Serial.println(t2-t1);
```

And I consistently got the result of 5006 (the difference is likely the execution time of the assignments anyways). If this is not sufficient, the common method seems to be to use a hardware timer interrupt on a high frequency and use it to increment an internal variable.

olivierouellet:
I think everything should be good and there's no need to reimplement anything.

From what I understand, `micros()` doesn't use DWT, but `delayMicroSeconds()` does. 

STM32duino [wiring_time.h](https://github.com/stm32duino/Arduino_Core_STM32/blob/93d2c963d8cb2dd2953d838a75edc3ad34a665a4/cores/arduino/wiring_time.h#L59)
```
static inline void delayMicroseconds(uint32_t) __attribute__((always_inline, unused));
static inline void delayMicroseconds(uint32_t us)
{
#if defined(DWT_BASE) && !defined(DWT_DELAY_DISABLED)
  int32_t start  = dwt_getCycles();
  int32_t cycles = us * (SystemCoreClock / 1000000);

  while ((int32_t)dwt_getCycles() - start < cycles);
#else
  __IO uint32_t currentTicks = SysTick->VAL;
  /* Number of ticks per millisecond */
  const uint32_t tickPerMs = SysTick->LOAD + 1;
  /* Number of ticks to count */
  const uint32_t nbTicks = ((us - ((us > 0) ? 1 : 0)) * tickPerMs) / 1000;
  /* Number of elapsed ticks */
  uint32_t elapsedTicks = 0;
  __IO uint32_t oldTicks = currentTicks;
  do {
    currentTicks = SysTick->VAL;
    elapsedTicks += (oldTicks < currentTicks) ? tickPerMs + oldTicks - currentTicks :
                    oldTicks - currentTicks;
    oldTicks = currentTicks;
  } while (nbTicks > elapsedTicks);
#endif
}
```
The reason I had a crash in #1693 was because I was using my own main function . `dwt_init()` was never called before `dwt_getCycles()`, causing the crash. #1701 fixed my problem, but it created a new compile error on boards not using DWT. 

To be honest, we could also remove #1701. Strictly speaking, if you go with STM32duino and don't override `main()`, my board should function ok (`premain()` will be called). But lots of people working on STM32s first start with SMT32CubeIDE to create the right   code for basic inits. It creates a `main()`, which does the same thing as STM32duino `main()`, except it never calls `dwt_init()`.
```
int main(void)
{
  HAL_Init();
  SystemClock_Config();

  MX_GPIO_Init();
  MX_IWDG_Init();
  MX_RTC_Init();
  //  [...]
}
```
There is no problem calling `dwt_init()` twice. The crash I had was hard to debug and might happen to others, so I would keep #1701 and this PR, but it's up to @jgromes. (I won't get offended, but I might stop sending you the maple syrup you use daily on these delicious palačinky.)

jgromes:
> Is there something stopping us from using the Core-defined micros() and delayMicroseconds() implementation in this case?

@krabrab there's not, I got confused - indeed it should be OK because the Arduino core takes care of the DWT-less STMs.

@olivierouellet `dwt_init` is being called from the HAL `init` method, which will only get called during module's `begin` method, that may happen at any point during runtime. So calling it again is the only way to be sure. If I was really paranoid, DWT might have even been disabled at some point after being enabled in `premain()` or `main()`...

> I might stop sending you the maple syrup you use daily on these delicious palačinky

So two points about this:

1. I don't know where you have been sending your maple syrup, for sure not here - you might want to get that looked into, as it could be that you are torturing some poor soul with large quantities of unsolicited syrup.
2. Excuse you sir, but I put apricot jam on my palačinky - as any civilized person should. Maple syrup??? Get this imperialist western nonsense out of here.

olivierouellet:
Ok, let's keep it the same. Thanks for your view on this. 

I also checked Arduino cores for other boards using Cortex processors. nRF boards should have no problem with [delayMicroseconds()](https://github.com/sandeepmistry/arduino-nRF5/blob/8d91cbb022af4f80de2f2131db275e08c9255321/cores/nRF5/delay.h#L61) (the function ends up calling [nrf_delay_us()](https://github.com/sandeepmistry/arduino-nRF5/blob/8d91cbb022af4f80de2f2131db275e08c9255321/cores/nRF5/SDK/components/drivers_nrf/delay/nrf_delay.h#L38)). Same thing for [SAMD21](https://github.com/arduino/ArduinoCore-samd/blob/993398cb7a23a4e0f821a73501ae98053773165b/cores/arduino/delay.c#L92), [SAMD51](https://github.com/ace22293/SAMD51/blob/8794281c907ac770ae9eca67912e15a7ab9884a6/cores/arduino/delay.h#L59) and [SAM3X](https://github.com/arduino/ArduinoCore-sam/blob/790ff2c852bf159787a9966bddee4d9f55352d15/cores/arduino/wiring.h#L62).

> I don't know where you have been sending your maple syrup

Hummm, that's a tough one. As with any reception issue, the first thing I would try to do is increase scanGuard. The other thing is I think the payload wasn't encrypted, so maybe a man-in-the-middle attack? I would first check if something happened during transit through Nederland. Maybe it got hijacked and ended up on a pannenkoeken?

PR #1735, opened 2026-03-23
Fail closed on SX126x receive metadata errors

robekl:
## Summary
- add internal error-aware SX126x helpers for packet type, packet status, and packet length metadata
- make `SX126x::readData()` fail closed when receive metadata retrieval fails
- stop packet RSSI/SNR helpers from fabricating values from failed metadata reads
- add regression tests covering SX126x receive-metadata command failures

## Problem
On SX126x, several receive-metadata helpers ignored `SPIreadStream(...)` return codes and treated zeroed or seeded buffers as valid chip data.

The highest-impact case is `SX126x::getPacketLength()`:
- it issued `GET_RX_BUFFER_STATUS`
- ignored the returned status
- and returned whatever remained in `rxBufStatus`

That meant a command/SPI failure could be interpreted as a valid packet length, including `0`.

This is a live issue for MeshCore's RadioLib integration. MeshCore's receive path asks RadioLib for packet length first and only calls `readData()` if the reported length is non-zero:
- `src/helpers/radiolib/RadioLibWrappers.cpp`

So on SX126x-family boards, a failed metadata read could be silently treated as “no packet”, or as a bogus packet length, instead of surfacing a receive error.

## Fix approach
I kept the public API stable and added internal error-aware helpers:
- `tryGetPacketType(...)`
- `tryGetPacketStatus(...)`
- `tryGetPacketLength(...)`

Those helpers return a real status code and only write outputs on success.

Then I switched the internal hot paths that need correctness to use them:
- `SX126x::readData()`
- packet RSSI and SNR helpers

For the legacy public `size_t getPacketLength()` API, there is no error channel because it comes from `PhysicalLayer`. To avoid silently collapsing a failed metadata read into “no packet”, the compatibility wrapper now returns a bounded non-zero sentinel (`maxPacketLength`) on failure instead of returning stale or zero metadata.

## Why this design
A more comprehensive long-term solution would be to add a public error-aware metadata API at the `PhysicalLayer` layer, for example a `tryGetPacketLength(...)`-style method that downstream users can adopt directly.

I did not do that in this PR because it would be a larger cross-library API change affecting many radios and downstream consumers.

This PR is intentionally narrower:
- fix the real live SX126x bug
- keep the existing public API intact
- create internal error-aware helpers that can support a broader future refactor

### New regression tests
Added `extras/test/unit/tests/TestSX126x.cpp` covering:
- `tryGetPacketLength()` propagates SX126x command timeout
- compatibility `getPacketLength()` no longer looks like “no packet” on metadata failure
- `readData()` fails closed when RX buffer status retrieval fails

### Proof against original code
I ran the new SX126x tests against an unmodified latest-master worktree.
The original code fails those tests because:
- `getPacketLength()` surfaces bogus packet-length data from a failed metadata read
- `readData()` returns success instead of propagating the metadata failure

### Fixed branch validation
Ran the full existing unit suite and the focused SX126x suite:
- `extras/test/unit/test.sh "-I/opt/homebrew/include -L/opt/homebrew/lib"`
- `./build/radiolib-unittest --run_test=suite_SX126x/* --log_level=test_suite`

Result:
- full suite passed
- SX126x regression suite passed

### DIFF
--- a/extras/test/unit/CMakeLists.txt
+++ b/extras/test/unit/CMakeLists.txt
@@ -13,6 +13,7 @@ file(GLOB_RECURSE TEST_SOURCES
   "tests/ModuleFixture.cpp"
   "tests/TestModule.cpp"
   "tests/TestCalculateTimeOnAir.cpp"
+  "tests/TestSX126x.cpp"
   "tests/TestPhyComplete.cpp"
   "tests/TestCrypto.cpp"
 )
--- a/extras/test/unit/tests/TestSX126x.cpp
+++ b/extras/test/unit/tests/TestSX126x.cpp
@@ -1,0 +1,131 @@
+#include <boost/test/unit_test.hpp>
+
+#include "ModuleFixture.hpp"
+
+#include "modules/SX126x/SX126x.h"
+
+class EmulatedSX126x : public EmulatedRadio {
+  public:
+    bool failRxBufferStatus = false;
+    bool failPacketType = false;
+    uint8_t packetType = RADIOLIB_SX126X_PACKET_TYPE_GFSK;
+    uint8_t rxBufferStatus[2] = {0x11, 0x22};
+
+    void HandleGPIO() override {
+      if(this->cs && this->cs->event && (this->cs->value == 0)) {
+        this->txnPos = 0;
+        this->currentCmd = 0x00;
+      }
+    }
+
+    uint8_t HandleSPI(uint8_t b) override {
+      if(this->txnPos == 0) {
+        this->currentCmd = b;
+        this->txnPos++;
+        return(0x00);
+      }
+
+      uint8_t out = 0x32;
+      if(this->txnPos == 1) {
+        if((this->currentCmd == RADIOLIB_SX126X_CMD_GET_RX_BUFFER_STATUS) && this->failRxBufferStatus) {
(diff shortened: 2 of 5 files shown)


jgromes:
While I agree the error handling could/should be improved, can you (or more likely your LLM of choice) enlighten me as to what exactly would change if instead of returning 0 on a failed read for the packet length we start returning maximum? That makes the situation even worse, because unlike 0, the maximum is a valid packet length. So the only thing that would change in meshcore is the below-linked condition:

https://github.com/meshcore-dev/MeshCore/blob/467959cc3bfc884e5f3425caac89453a450151b6/src/helpers/radiolib/RadioLibWrappers.cpp#L113

The test that this PR also adds seems pretty useless - it basically checks whether the test fixture was configured to reject the command by setting `failRxBufferStatus` to `true`. Seriously, I have no idea what is it supposed to be testing.

Bottom line - if there's a request from the user base to change the API so that basic getter methods like `getPacketLength` start returning status code and pass the actual packet length via reference, we can discuss it via an issue, let the major users know in advance and then make the breaking change in the next major release. But this PR does not help that at all.

robekl:
> But this PR does not help that at all

Understood. It does seem like a bigger issue than something for a quick workaround. I'll close this in preference of a better solution.

PR #1744, opened 2026-03-30
[LoRaWAN] Add Package Manager

StevenCellist:
This pull request revamps the interface for using packages as defined by Technical Specification documents. It provides a single Package Manager interface which supports adding multiple (many) packages with ease.

This pull request is fully verified using OTAA EU868 meaning it passes all corresponding tests in LCTT.

Some things may look a bit over-engineered for just the presence of TS009 Certification. However, the setup allows us to very easily expand with new packages in the future.

_Yes, I had lots of fun coming up with the PacMan short notation._


StevenCellist:
One CI is telling me to `#include <cstring>` but then the others throw errors... @jgromes huh?

jgromes:
That's because we are building everything using a whole bunch of different compiler versions. `#include <string.h>` should be compatible with everything.

src/protocols/LoRaWAN/LoRaWANPacMan.h
```diff
@@ -0,0 +1,202 @@
...
+    /*!
+      \brief Query the next scheduled task for this package
+      \returns `LoRaWANTaskInfo` containing task type and time
+    */
+    virtual LoRaWANTaskInfo hasTask();
+
+    /*!
+      \brief Perform an ACTION task (if any). Default no-op.
+    */
+    virtual void doAction();
```
jgromes: Would it make sense to have add a `void* arg` to allow the user to pass something to or from this method?

src/protocols/LoRaWAN/LoRaWANPacMan.h
```diff
@@ -0,0 +1,202 @@
...
+
+    // Package enablement methods
+    /*!
+      \brief Enable TS009 Certification Protocol package
+      \returns Status code
+    */
+    int16_t enableTS009(PhysicalLayer* radio, DelaySecondsCb_t delayCb, UplinkIntervalCb_t intervalCb, RebootCb_t rebootCb);
+
+    // TS009
+    bool getConfirmed();
```
jgromes: This also seems very tailored to TS009.

src/protocols/LoRaWAN/LoRaWANPacMan.cpp
```diff
@@ -0,0 +1,215 @@
...
+          RADIOLIB_DEBUG_PROTOCOL_PRINTLN("Package on FPort %d consumed 0 bytes", fPort);
+        }
+        // Package-specific answers remain in package buffer and will be
+        // exposed via hasTask()/getUplinkData(). Nothing more to do here.
+        return(RADIOLIB_ERR_NONE);
+      }
+    }
+    // No package matched this FPort; nothing to do.
+    RADIOLIB_DEBUG_PROTOCOL_PRINTLN("No package matched FPort %d, ignoring payload", fPort);
+    return(RADIOLIB_ERR_NONE);
```
jgromes: Nitpick, but this return is unnecessary.

src/protocols/LoRaWAN/LoRaWANPacMan.cpp
```diff
@@ -0,0 +1,215 @@
...
+  if(dataIn == NULL) {
+    return(RADIOLIB_ERR_NULL_POINTER);
+  }
+
+  // If the downlink arrived on a package-specific FPort (not the package
+  // manager's TS007 FPort), forward the entire payload to that package.
+  if(fPort != RADIOLIB_LORAWAN_FPORT_TS007) {
+    // Find package registered on this FPort
+    for(uint8_t i = 0; i < RADIOLIB_LORAWAN_NUM_PACKAGES; i++) {
+      if(!this->enabledPackages[i]) continue;
```
jgromes: I would strongly recommend to use braces even with one-liner conditions. For details what happens when brace-less conditions get out of hand, see [Apple goto fail bug](https://www.blackduck.com/blog/understanding-apple-goto-fail-vulnerability-2.html) ;)

jgromes (commented):
Thanks, I left some comments, it's mostly minor stuff but the thing about TS009 presence in pacman I really don't understand.

StevenCellist:
Taking this out of the review thread as it's important to discuss:

> why is it (TS009, red.) a part of the package manager itself, as opposed to having it in some dedicated test suite?

Certification testing is meant to be an available mode in the final software: you are not supposed to ship different firmware than what the test labs are getting. So it should not be a dedicated test suite.
TS009 also accounts for this by having a command to disable TS009 itself; this is the last step during certification testing. It is supposed to set a bit in NVS that tells the device not to turn on TS009 (unless e.g. a specific button is pressed, flipping that bit again). That last command is something that I have not implemented yet but this will follow very soon.

Hence it is regarded as a normal package. And as I'm looking to keep the user-interface as simple as possible, I'd rather not have the user retrieve the TS009 package in order to find the `getConfirmed()` function. Therefore, it is relayed through the package manager so that the user does not need to concern him- or herself with `LoRaWANPackageTS009` types and variables.

Does that clarify / satisfy your reservations, or do you have additional considerations?

src/protocols/LoRaWAN/LoRaWANPacMan.cpp
```diff
@@ -0,0 +1,215 @@
...
+      }
+    }
+  }
+
+  *lenOut = 0;
+  return(false);
+}
+
+bool LoRaWANPackageManager::doAction() {
+  RadioLibTime_t now = this->getSecondsCb();
```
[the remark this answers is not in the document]
StevenCellist: PacMan is guarded, I didn't guard the package implementations but I did guard the `enableTS009` function. Does that suffice?

jgromes:
Thanks for the clarification regarding TS009 usage. The main thing I was worried about was what is the impact on users that do not care about certification and just want to have a LoRaWAN node. But when trying to build some examples, with and without the methods in `LoRaWANPackageManager` that pull in TS009, it seems that there is no effect when not using the package manager, which I think will be the case for > 90% of LoRaWAN users. So we can leave it as is.

StevenCellist:
Yes, the goal with the package manager is to explicitly keep it away from the `LoRaWANNode`. They should be fully distinct features. The example would be much cleaner if we integrated PacMan into Node but then the compile size jumps by 10+kB once all packages are there, or you need additional compilation flags which I utterly despise (looking at you, LMIC).

PR #1753, opened 2026-04-14
[LoRaWAN] Add TS003 Application Layer Clock Synchronization package

StevenCellist:
This pull request adds the LoRaWAN TS003 Application Layer Clock Synchronization package to the Package Manager. Specification as per [doc](https://resources.lora-alliance.org/technical-specifications/ts003-2-0-0-application-layer-clock-synchronization). Feel free to read through it but it's verified using LCTT so personally not concerned about functionality. Code review is useful though.


jgromes (approved):
Thanks, I did not find anything that would require adjustments.

Issue #1757, opened 2026-04-19
SX128x setBandwidth() rejects valid approximate values — 203.0 fails instead of snapping to 203.125


rampageservices:
## Bug

`SX128x::setBandwidth()` returns `RADIOLIB_ERR_INVALID_BANDWIDTH` (-8) for inputs that are unambiguously close to a valid value.

**Example:** `begin(2400.0, 203.0, 9, 7, 20, 12)` fails with -8 even though the only LoRa BW near 203.0 kHz is 203.125 kHz.

## Root cause

The function uses exact-match with ±0.001 kHz tolerance:

```cpp
if(fabsf(bw - 203.125f) <= 0.001f) { ... }
```

`203.0f` misses by 0.125 kHz and hits the `return RADIOLIB_ERR_INVALID_BANDWIDTH` fallthrough. The prior `RADIOLIB_CHECK_RANGE(bw, 203.125f, ...)` also rejects values below 203.125, so `203.0` fails before the tolerance check even runs.

## Expected behavior

Since the four valid LoRa BW values (203.125, 406.25, 812.5, 1625.0 kHz) are discrete hardware enum entries, any reasonable input has an unambiguous nearest value. The function should snap to the nearest valid BW rather than hard-failing on common approximations.

## Affected hardware

SX1280 / SX1281 LoRa mode. Also affects Ranging mode (406.25–1625.0 kHz).

## Fix

PR #1756 replaces exact-match with midpoint-threshold snapping and sets `bandwidthKhz` to the canonical snapped value.

rampageservices:
Withdrawing — see my comment in #1763.

PR #1758, opened 2026-04-19
feat(SX128x): add per-step debug diagnostics to begin()

rampageservices:
## Problem

When `SX128x::begin()` fails, the caller receives only a numeric error code. With `RADIOLIB_DEBUG_BASIC` enabled the existing output identifies the chip-not-found case but gives no indication of *which configuration step* failed or *with what parameter*. This makes debugging parameter errors (wrong bandwidth, out-of-range power, etc.) unnecessarily difficult.

## Fix

Replace the silent `RADIOLIB_ASSERT(state)` calls in `SX128x::begin()` with explicit checks that emit a `RADIOLIB_DEBUG_BASIC_PRINTLN` message on failure, including the step name, the parameter value passed, and the error code:

```
SX128x LoRa begin() failed at setBandwidth(203.000 kHz) [-8]
SX128x LoRa begin() failed at setOutputPower(25 dBm) [-13]
SX128x LoRa begin() failed at setFrequency(9999.000 MHz) [-12]
```

Output is **gated behind `RADIOLIB_DEBUG_BASIC`** — release builds are completely unaffected. Error code semantics and return values are unchanged.

## Related

- #1757 — root cause that surfaced this gap (bandwidth parameter mismatch hard to diagnose)
- PR #1756 — BW snapping fix

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -33,28 +33,52 @@ int16_t SX128x::begin(float freq, float bw, uint8_t sf, uint8_t cr, uint8_t sync
 
   // set module properties and perform initial setup
   int16_t state = this->modSetup(RADIOLIB_SX128X_PACKET_TYPE_LORA);
+  if(state != RADIOLIB_ERR_NONE) {
+    RADIOLIB_DEBUG_BASIC_PRINTLN("SX128x LoRa begin() failed at modSetup [%d]", state);
+  }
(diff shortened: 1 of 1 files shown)


jgromes:
How is this better or different than the existing `RADIOLIB_VERBOSE_ASSERT` build-time option? See also the [wiki](https://github.com/jgromes/RadioLib/wiki/Debug-mode#assertions).

rampageservices:

PR #1760, opened 2026-04-19
docs(SX1278): document NRESET=NC crystal startup workaround in reset()

### LINKED ISSUE #1759: SX127x: NRESET=RADIOLIB_NC causes intermittent CHIP_NOT_FOUND when crystal is cold
## Summary

When `NRESET` is passed as `RADIOLIB_NC` (e.g. on designs where the SX127x reset line is tied to the board power-enable rail rather than a GPIO), `SX1278::reset()` becomes a no-op for the GPIO calls — the `ArduinoHal` skips `NC` pins. The chip is never actually reset, and the crystal oscillator state is not changed by the reset sequence.

## Root Cause

`SX127x::findChip()` calls `reset()` before every version register read (up to 10 attempts). The purpose of the reset is to:

1. Pulse NRESET LOW to force the chip to a known state
2. Release NRESET HIGH to let the chip exit reset  
3. Wait 5 ms for the crystal oscillator to reach stable amplitude

When `NRESET=RADIOLIB_NC`, steps 1–2 are skipped entirely. The 5 ms `delay()` still executes (delays are not gated on NC), but the crystal oscillator was never cycled.

The SX127x powers on in **FSK sleep mode** with the crystal stopped. `RadioLib::begin()` transitions the chip from sleep to standby via a register write, which takes ~20 µs. A cold crystal (32 MHz, typical XTAL) needs ~2–5 ms to reach the ~82% amplitude threshold required for the on-chip oscillator circuit to lock. SPI communication attempted before the crystal stabilizes results in:

- `RADIOLIB_ERR_CHIP_NOT_FOUND` (-2) — version register reads garbage
- Corrupted subsequent register reads — `begin()` proceeds but parameters are wrong
- Behavior varies boot-to-boot depending on how long the MCU took to reach `begin()`

## Affected Hardware

Any design where the SX127x NRESET is not connected to a dedicated GPIO. Common patterns:

- NRESET tied to ESP32 `CHIP_PU` (power enable)
- NRESET pulled to VCC with no MCU connection  
- NRESET left floating (internally pulled up on SX127x)

## Workaround

Before calling `begin()`, manually step the chip through FSK sleep → LoRa sleep → LoRa standby via raw SPI register writes (OpMode register 0x01):

```cpp
// Pre-warm SX127x crystal oscillator when NRESET is not connected to a GPIO.
// RadioLib's begin() transitions sleep -> standby in ~20 µs; crystal cold-start
// needs ~2 ms. Manually step FSK sleep -> LoRa sleep (5 ms crystal startup) ->
// LoRa standby so begin() sees a stable oscillator.
auto wr = [&](uint8_t val) {
    digitalWrite(NSS_PIN, LOW);
    SPI.transfer(0x81);  // write OpMode register
    SPI.transfer(val);
    digitalWrite(NSS_PIN, HIGH);
};
SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE0));
wr(0x08);  delay(2);  // FSK sleep     — crystal off
wr(0x88);  delay(5);  // LoRa sleep    — crystal starts
wr(0x89);  delay(2);  // LoRa standby  — crystal oscillating
SPI.endTransaction();
// Now safe to call radio.begin(...)
```

The crystal decay constant is τ ≈ 100 µs for a typical 32 MHz XTAL. After the 5 ms LoRa sleep step the crystal is at full amplitude. The ~20 µs `begin()` re-sleep drops amplitude to only ~82%, which restarts in < 100 µs — no second cold-start delay needed.

## Proposed Fix

At minimum, a documentation note in `SX1278::reset()` (and `SX1272::reset()`) explaining the behavior and the workaround. A longer-term fix could be to detect `RADIOLIB_NC` in `reset()` and insert a compensating delay to approximate crystal startup time, but this would be a behavioral change.

A small documentation PR will follow.

## References

- SX1276/77/78/79 datasheet §5.4.8 (Crystal oscillator startup)
- Semtech AN1200.24 (SX1276 crystal startup timing)

rampageservices:
## Summary

Adds a documentation comment to `SX1278::reset()` (and by inheritance `SX1276::reset()`) explaining the behavior and workaround when `NRESET=RADIOLIB_NC`.

Fixes #1759.

## What changed

`SX1278::reset()` — added a block comment explaining:

- When `NRESET=RADIOLIB_NC`, the `ArduinoHal` skips all GPIO operations, so no actual chip reset occurs
- The SX127x starts in FSK sleep with the crystal stopped; `begin()`'s sleep → standby transition is ~20 µs, too short for cold crystal startup (~2–5 ms)
- This causes intermittent `CHIP_NOT_FOUND` (-2) or corrupted register reads on affected hardware
- Workaround: manually step FSK sleep → LoRa sleep → LoRa standby via raw SPI before calling `begin()`

No behavior change — comment only.

## Test plan

- [ ] Verify no compilation errors (`SX1276`, `SX1278`, `SX1272` targets)
- [ ] Verify `SX1278::reset()` behavior is unchanged on hardware with NRESET connected

### DIFF
--- a/src/modules/SX127x/SX1278.cpp
+++ b/src/modules/SX127x/SX1278.cpp
@@ -74,6 +74,21 @@ int16_t SX1278::beginFSK(float freq, float br, float freqDev, float rxBw, int8_t
 }
 
 void SX1278::reset() {
+  // When NRESET is RADIOLIB_NC the GPIO calls below are no-ops (ArduinoHal
+  // skips NC pins) but the delays still execute. This occurs on designs where
+  // NRESET is tied to the board power-enable line (e.g. ESP32 CHIP_PU) rather
+  // than a dedicated GPIO. In that case no actual chip reset occurs and the
+  // crystal oscillator state is unchanged.
+  //
+  // Consequence for begin(): the SX127x powers on in FSK sleep mode with the
+  // crystal stopped. RadioLib's begin() transitions sleep -> standby in ~20 µs
+  // (a register write), which is too fast for a cold crystal start (~2-5 ms).
+  // Intermittent CHIP_NOT_FOUND or corrupted register reads can result.
+  //
+  // Workaround when NRESET=RADIOLIB_NC: before calling begin(), manually step
+  // the chip through FSK sleep -> LoRa sleep (5 ms) -> LoRa standby via raw
+  // SPI register writes. This pre-warms the crystal so begin() sees a stable
+  // oscillator. See SX1278 datasheet §5.4.8 and application note AN1200.24.
   Module* mod = this->getMod();
   mod->hal->pinMode(mod->getRst(), mod->hal->GpioModeOutput);
   mod->hal->digitalWrite(mod->getRst(), mod->hal->GpioLevelLow);


jgromes:
I am confused about the proposed workaround - why is the mode cycling needed? Is a simple delay if `NRESET == RADIOLIB_NC` not sufficient?

rampageservices:

PR #1762, opened 2026-04-19
fix(SX128x): send NOP wakeup in reset() before standby

### LINKED ISSUE #1761: SX128x: reset() deadlocks when chip is in sleep mode (BUSY stuck HIGH)
## Summary

`SX128x::reset()` deadlocks when the chip is already in sleep mode, causing `findChip()` to return `CHIP_NOT_FOUND` (-2) after a 3-second timeout.

## Root Cause

In sleep mode, the SX128x holds BUSY HIGH and ignores all SPI commands until it receives a NOP (opcode `0xC0`) as a wakeup signal. `reset()` polls `standby()` in a retry loop, but calls it without `wakeup=true`, so the NOP is never sent. Every `standby()` attempt waits for BUSY to go LOW, times out, and retries. After 3 seconds the loop exits and returns the timeout error.

```cpp
// reset() — current code
int16_t state = standby();   // wakeup=false — NOP never sent, chip stays asleep
```

The fix is already available in the `standby()` overload:

```cpp
int16_t standby(uint8_t mode, bool wakeup);
// when wakeup=true, sends NOP before SetStandby
```

## When This Occurs

- Any call to `begin()` or `reset()` when the chip was previously put to sleep via `sleep()`
- Power-on with NRESET tied to a power-enable line (RADIOLIB_NC): the hardware reset does not occur, and if the chip retained sleep state across a soft reboot it will be asleep when `begin()` is called
- Multi-radio designs that call `sleep()` on the SX128x before switching to another radio, then call `begin()` again later

## Fix

One-line change in `reset()`:

```cpp
// before
int16_t state = standby();

// after
int16_t state = standby(RADIOLIB_SX128X_STANDBY_RC, true);
```

A PR with this fix will follow.

rampageservices:
## Summary

Fixes #1761.

`reset()` called `standby()` without `wakeup=true`. When the chip is in sleep mode, BUSY is HIGH and the chip ignores all SPI commands until a NOP is received. Every `standby()` attempt timed out, causing `findChip()` to return `CHIP_NOT_FOUND` after the 3-second retry window.

## Change

`src/modules/SX128x/SX128x.cpp` — one line in `reset()`:

```cpp
// before
int16_t state = standby();

// after  
int16_t state = standby(RADIOLIB_SX128X_STANDBY_RC, true);
```

`standby(..., true)` sends a NOP command (0xC0) before SetStandby, waking the chip from sleep before attempting the standby transition. Sending NOP to an already-awake chip is a no-op per the datasheet.

## Test plan

- [ ] `begin()` succeeds after `sleep()` without requiring a hardware power cycle
- [ ] `begin()` still succeeds on fresh power-on (NOP to awake chip is harmless)
- [ ] No regression on SX1280/SX1281/SX1282 targets

### DIFF
--- a/src/modules/SX128x/SX128x.cpp
+++ b/src/modules/SX128x/SX128x.cpp
@@ -201,11 +201,12 @@ int16_t SX128x::reset(bool verify) {
     return(RADIOLIB_ERR_NONE);
   }
 
-  // set mode to standby
+  // set mode to standby - send NOP first to wake the chip if it is in sleep
+  // (in sleep BUSY is HIGH; without the NOP the standby command never gets through)
   RadioLibTime_t start = this->mod->hal->millis();
   while(true) {
     // try to set mode to standby
-    int16_t state = standby();
+    int16_t state = standby(RADIOLIB_SX128X_STANDBY_RC, true);
     if(state == RADIOLIB_ERR_NONE) {
       // standby command successful
       return(RADIOLIB_ERR_NONE);


jgromes:
Before I evaluate any of the issues and unsolicited PRs you have opened: have you read our AI usage policy in the [contributing guideline](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md#ai-use)?

rampageservices:
Yes, I read the policy. I used a local model to structure my words more clearly — this is how I work efficiently. I'm American, born and raised, but I've spent the past 10 years in China running my own business and raising a family, working across multiple languages daily. Code-switching between them in writing isn't trivial, and using a local model to tighten my English is just a tool — not a pipeline to a billion-dollar scraper.
The fix itself is real hardware work: I debugged this on an SX1281 board I'm building, hit the BUSY deadlock in sleep mode, and validated the NOP wakeup resolves it. Tested on fresh power-on to confirm no regression.
We're contributing this in free time because we hit the problem and solved it. If RadioLib has (or wants) a FUNDING.yml, we're happy to put money where our mouth is and support your work with actual dollars — I'll open a PR for that too if you want to get started.
On the "wall of text" point — no offense intended, and I hear the guideline. But this is a public forum, not a private review. Like Stack Overflow, the next person hitting CHIP_NOT_FOUND after sleep() is going to land on this PR via Google. The symptom (BUSY stuck HIGH, 3-second retry timeout) isn't obvious from the outside, and documenting the mechanism and fix means the next person doesn't burn hours on it. I'm not trying to lecture you on your own codebase — I'm scaffolding the writeup for the downstream reader. Happy to trim it if you'd prefer, but that's the reasoning.
That said, labeling real hardware debugging work as "AI slop" is frankly offensive. We spent actual time on this, not generating text. If the code change is solid, it's solid — the mechanism is sound per the datasheet. What's your read on the fix itself?

rampageservices:
Follow-up: opened #1763 to figure out the best way to support the project. Happy to put money behind RadioLib, not just PRs.

One pushback though. Calling this unsolicited doesn't quite sit right, regardless of what CONTRIBUTING.md says. This is a public repo, and that makes it a public forum. Drive-by fixes are normal in open source. The usual flow for a small fix is a draft PR alongside an issue so the patch and discussion live in one place. Requiring issue-first, wait, PR-later adds round-trips busy contributors can't afford, and it filters out exactly the kind of people who ship hardware and find bugs.

I should have marked this one as draft. That's on me. But a one-line fix shouldn't need a pre-approval round-trip.

jgromes:
First things first:

> What's your read on the fix itself?

It seems strange - why is the existing `standby()` call insufficient? It should default to RC standby with wakeup enabled, which is the same approach as is used on SX126x. So we should figure out why is that not happening.

Now for the rest:

>Yes, I read the policy.

That's great! Why are you not following it then? It clearly states that AI created content is to be marked as such. It is the very first point. You have also ignored the second one, and procedded to generate a mass amount of text in issue and a PR to describe ... a fix that adds one argument. Seriously, how is that easier than typing a simple description? I just do not understand the motivation.

>  I'm American

English is not my native language, but I would find it very weird if I required a machine to structure my own words. It is obvious you are using a model, just by the amount of inversions and em-dashes in your text alone. 

> If RadioLib has (or wants) a FUNDING.yml, we're happy to put money where our mouth is and support your work with actual dollars

Never had and never will. Keep the money, buy a book and read it.

> I'm scaffolding the writeup for the downstream reader

I think the main effect is increasing the amount of information they will have to go through. This could have been a one-line description. Now it's an issue, a PR and a discussion I seem to be having with you through a model.

> That said, labeling real hardware debugging work as "AI slop" is frankly offensive. We spent actual time on this

Unfortunately for you, there is no way for me to dinstinguish that as I only see the output of the model. Not the output of you.

> Requiring issue-first, wait, PR-later adds round-trips busy contributors can't afford, and it filters out exactly the kind of people who ship hardware and find bugs.

IMO there is nothing in contributing that requires this - point 3 does not apply here as it specifically calls out large, API-altering PRs. I will say though that firing off a whole bunch of issues/PRs feels overwhelming - that's the unsolicited bit.

rampageservices:
Closed 1762.
Also, I'm a real guy. SX1281 board on a bench in China. If you ever want to hop on a Zoom I'm happy to.
No urgency. Just wanted you to know there's a person on this end.

PR #1768, opened 2026-04-28
[LR2021] PA optimization

### LINKED ISSUE #1740: LR2021 sub-GHz TX power mismatch (configured vs measured)
Hi, I found a TX power mismatch issue on LR2021 in sub-GHz mode.

At 868 MHz, the configured power and measured output differ a lot (especially at higher power settings), beyond normal tolerance.

I have fixed it locally by updating LR2021 PA/TX configuration logic according to datasheet tables and half-dBm TX params.

Reference: https://github.com/hlym123/RadioLib/commit/f5a23d999a1d707938dbe24ae3e21a6558dda62b

If needed, I can submit this as a PR.

jgromes:
This PR adds PA configuration for LR2021 based on lookup table. It has been optimized the same way as SX126x series has been, measurement logs are available in https://github.com/radiolib-org/power-tests

So far only Sub-GHz PA has been measured, hence this is still in draft. Because the LR2021 does not seem to have a fixed default (like the SX126x did), this optimization will always be enabled.

We also need to figure out a simple mechanism to allow users to provide their own PA configuration table.

Fixes #1740

### DIFF
--- a/src/modules/LR2021/LR2021.h
+++ b/src/modules/LR2021/LR2021.h
@@ -68,6 +68,17 @@ class LR2021: public LRxxxx {
         MODE_TX_HF,
     };
 
+    /*!
+      \struct paTableEntry_t
+      \brief This structure is used as entry in the PA lookup table, 
+      to optimize PA configuration for minimum power consumption.
+    */
+    struct __attribute__((packed)) paTableEntry_t {
+      uint8_t paDutyCycle: 4;
+      uint8_t paSlices: 4;
+      int8_t paVal;
+    };
+
     // basic methods
 
     /*!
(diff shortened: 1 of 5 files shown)


jgromes:
I was not able to fully verify everything (especially the 2.4 GHz settings), but nevertheless this version is a lot better than the current master, so we should merge it earlier and fix further issues afterwards.

StevenCellist:
If you wish for me to try this with my LR2021, please be patient as I'm head over heels in STM32WL and my sensor boxes. If you just want to hear "I see no obvious weird stuff", then consider this my seal of approval ;)

PR #1774, opened 2026-05-05
[LR11x0][LR2021] Add support for duty cycle Rx

### LINKED ISSUE #1769: LR11x0 and LR2xxx RX Duty Cycle feature
**Is your feature request related to a problem? Please describe.**

On meshtastic we use 16 symbols preamble and this gives power savings on SX1262.
I would like to use the same feature on the LR11x0 and LR2xxx but it is not exposed:
https://grep.app/search?f.repo=jgromes%2FRadioLib&case=true&q=setRxDutyCycle

If you could copy the exact same API as the SX1262:
https://grep.app/search?f.repo=jgromes%2FRadioLib&case=true&q=startReceiveDutyCycle

it would be convenient for me since I already have the code.

Doing nothing.

**Additional context**

jgromes:
This PR adds support for duty-cycle Rx with the same interface that is present on SX126x. It also does a minor change for LR11x0 DIO pins, which are now being driven in sleep mode. It doesn't seem to have any impact on power consumption, but without it, DIO pins on LR11x0 would float in sleep mode, which caused spurious interrupts being detected in the host MCU.

Similar issue is also present on LR2021 and is handled there by enabling pull-downs by default. An important note is that DIO5 on LR2021 can only have a pullup in sleep mode, making it unusable as IRQ when duty cycle is used.

Two cases were tested, `startReceiveDutyCycleAuto(16, 6);` and `startReceiveDutyCycle(5000000, 1000000);`. The latter allowed to measure the sleep/Rx periods with oscilloscope (when DIO pins start to float after entering sleep mode).

Closes #1769

### DIFF
--- a/src/modules/LR11x0/LR11x0.cpp
+++ b/src/modules/LR11x0/LR11x0.cpp
@@ -416,6 +416,28 @@ int16_t LR11x0::startReceiveDutyCycle(uint32_t rxPeriod, uint32_t sleepPeriod, R
   return(this->setRxDutyCycle(rxPeriodRaw, sleepPeriodRaw, RADIOLIB_LR11X0_RX_DUTY_CYCLE_MODE_RX));
 }
 
+int16_t LR11x0::startReceiveDutyCycleAuto(uint16_t senderPreambleLength, uint16_t minSymbols, RadioLibIrqFlags_t irqFlags, RadioLibIrqFlags_t irqMask) {
+  // calculate the sleep and wake periods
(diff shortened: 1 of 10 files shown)


jgromes:
FYI @Jorropo - feedback on this and actual real world testing would be appreciated, thanks.

jgromes:
By measuring of the power consumption during Rx dtuy cycle I am reasonably confident this works, so merging. If something is broken, feel free to file a new issue report.

PR #1776, opened 2026-05-07
a fix for PiHal.h build with LGPIO V1

lyusupov:
LGPIO V1 had no support for pull up and pull down actions yet.

### DIFF
--- a/src/hal/RPi/PiHal.h
+++ b/src/hal/RPi/PiHal.h
@@ -225,6 +225,7 @@ class PiHal : public RadioLibHal {
       lgTxPwm(_gpioHandle, pin, 0, 0, 0, 0);
     }
 
+#if LGPIO_VERSION > 0x00010000
     void pullUpDown(uint32_t pin, bool enable, bool up) {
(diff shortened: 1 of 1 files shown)


jgromes:
Good catch, thanks! It might also be a good idea to add a warning for versions lower than the latest release `0x00020200` - AFAIK there should be no major reason to not updated if that outdated version is present.

lyusupov:
V2 LGPIO library works with V2 of Linux kernel GPIO API only.

There are reasons to use LPGIO V1 sometime:
1) Raspberry Pi distributions ( e.g. Raspbian ) before year 2022 have kernel GPIO driver with API 1.x
2) other SBCs, like based on Rockchip or AllWinner SoCs, may use V1 GPIO Linux kernel API till today...

jgromes:
Fair enough, but still, a warning would make it clear IMO ;) This way the method is just going to disappear/get replaced by the default implementation which does nothing.

lyusupov:
I have no objections against to add a warning. Just not sure what particular text should it contain.
Could you, please, add this warning on your discretion ?

jgromes:
Done now, and merged - thank you for the contribution!

PR #1780, opened 2026-05-12
[Crypto] Interface for hardware acceleration

### LINKED ISSUE #1778: Interface for hardware-crypto
**Is your feature request related to a problem? Please describe.**
I'm working with the STM32WLE5 chip which features hardware-accelerated cryptography. Which is like > 10× faster than software crypto, especially on low clock speeds.

An interface to customize or override specific Cryptography functions for a certain platform/chip. Something like the HAL stuff maybe. Preferably it doesn't _require_ re-implementing all Crypto functions.

Currently living off of a custom version of Cryptography.h and .cpp but that's not really ideal... It uses a whole bunch of #ifdefs for my platform but that's not going to expand nicely to other chips or platforms.

jgromes:
This PR adds support for implementing hardware-accelerated AES-128. This can be achieved by the following steps:

1. Setting a build-time option `-DRADIOLIB_CUSTOM_AES128=1` and 
2. Providing a new class that implements the pure virtual methods of `RadioLibAES128`
3. Creating an instance of this class and providing a pointer to this instance to the HAL via its member `RadioLibAES128* Hal::aes128` as shown in the snippets below.

```c++
class CustomAES128: public RadioLibAES128 {
  public:
    void init(uint8_t* key) override {
      // custom init here
    };

    size_t encryptECB(const uint8_t* in, size_t len, uint8_t* out) override {
      // custom encryption here
    };
    
    size_t decryptECB(const uint8_t* in, size_t len, uint8_t* out) override {
      // custom decryption here
    };

};

CustomAES128 myAes;

(...)

// in the init() method of your HAL:
this->aes128 = &myAes;
```

Closes #1778


StevenCellist (approved):
Works like a charm! This saves like 250ms of AES computation on every up/downlink at 16 MHz and I hope to drop to 8 MHz or even 4 MHz once I am comfortable enough removing the debug output.

jgromes:
Very good, thanks for the test!

PR #1784, opened 2026-05-16
Rework of begin methods

### LINKED ISSUE #1620: Rework of begin methods
Currently, all the `begin*` methods for all modules have a rather extensive list of arguments, which set basic radio properties like carrier frequency, LoRa spreading factor, FSK bit rate etc. This is unwieldy, leads to the default values being copy-pasted everywhere, is not futureproof as changing the order of arguments can cause silent failures, and overall just not being used as a lot of users will instead opt into calling the default no-argument version and then modifying whatever actually interests them.

I have written down the options discussed so far with examples of the API calls. Fruther proposals can be added. 

1. **Named arguments** - proposed by @cleishm in #1607. This approach uses a flat config structure as the only argument of the `begin` methods. Values for this structure can be provided by designated initialization directly in the `begin` method call, so from the perspective of most users, the only thing that changes is that now they can specify just a single argument in a very clean way.

    ```c++
    int state = radio.begin({.spreadingFactor = 9});
    ```
 
2. **Config structs with inheritance** - proposed by @jgromes. This approach is similar to the one above, except that the configuration structure is no longer flat and inherits from a base struct which contains all the common configuration. This limits the copy-paste, connects the configuration to existing config structures like `DataRate_t` and would allow us to get rid of the different variants of `begin`, like `beginFSK` or `beginLRFHSS`. It also allows more advanced users to easily switch between different modems and configurations. The downside is that the user API is more complicated.

    ```c++
    SX126xConfig_t cfg(RADIOLIB_MODEM_LORA);
    cfg.dataRate.lora.spreadingFactor = 9;
    int state = radio.begin(&cfg);
    ```

3. **Removing unnecessary configuration from begin** - most of the configuration arguments in `begin` calls are not actually needed to start the module. There are exceptions (for example TCXO on SX126x), but the list of arguments in `begin` can be actually removed most of the time. This would necessitate the user to configure the arguments themselves, which some already do anyway. The advantage here that this is the simplest solution, the clear disadvantage is that it pushes all the configuration onto the user, leaving the radio in a weird semi-configured state between the end of `begin` and completed initialization. It will also discourage users from checking all of the error codes, as every call to config method will need to be checked, which is currently done by the library.

    ```c++
    int state = radio.begin();
    RADIOLIB_ASSERT(state);
    state = radio.setFrequency(434);
    RADIOLIB_ASSERT(state);
    state = radio.setSpreadingFactor(9);
    ```

4. **Configure modem and use named arguments** - proposed by @StevenCellist. This proposal uses the existing `setModem()`  (possibly renamed) followed by approach 1 above with named arguments. The user must first (re)set the module to a specific modem (LoRa, FSK, ..) which does the usual `findChip()` stuff to make sure that the radio is connected properly. After that, the user calls `begin()` with named arguments with possible default values as in the proposal by @cleishm. This argument list is selectively interpreted based on the configured modem. 
As a side-effect, this also tackles the 'wake-up from deepsleeping ESP32'-problem as then the user can skip the call to `setModem()`.

    ```c++
    int state = radio.setModem(RADIOLIB_MODEM_LORA);
    RADIOLIB_ASSERT(state);
    state = radio.begin({.spreadingFactor = 9});
    ```

As this is a major breaking change, it will be rolled out gradually over the next library releases. The original list of arguments will be removed in 8.0.0, breaking backwards compatibility there.

jgromes:
As discussed in #1620


jgromes:
While implementing I also noticed that most modules reuse the exact same configuration, which leads to copy-pasted configuration structures. I think those should be moved into some common file, and all the configuration parameters that are not universal (e.g. `tcxoVoltage` for SX126x series) could be moved to publicly accessible member variables. Precedent in this already exists, as `SX126x::XTAL` is basically just a shortcut for `tcxoVoltage=0`.

StevenCellist (commented):
I'm of course not going to look through the whole diff.. my browser is also lagging big time typing and rendering this.

Most importantly, I think this needs a wiki page that is linked somewhere very very obvious or a high-up README section. The empty bracket notation `{}` that is used in most examples isn't obviously transformed into C++11 compliant code. And it's only in some specific example such as ADS-B that I see the two different methods 'explained'.

jgromes:
@StevenCellist thanks, a line-by-line review is definitely not requested here, I'm looking for exactly these high-level points ;)

I'm thinking how to solve this because the only reason the why the `radio.begin({});` calls build now in C++11 environments is because of the deprecated methods that will be removed in 8.0.0. So actually the examples shouldn't rely on those and should use a begin call with those configuration structures. At least showing how to set the freuqency since that is the first thing most users will need to adapt, based on their location.

Maybe we should also invert the logic of those ifdefs (show the config structure first and initializer list as optional) or even remove them completely, and just reference them in a comment: that for C++14 and newer, initializer list is an option.

Do you think that would make it more clear?

StevenCellist:
Yes to both, please!

jgromes:
@StevenCellist in the last few commits, I tried a more generic approach to the config structures (currently only implemented for SX126x) - let me know what you think.

StevenCellist:
That looks very welcome, because it means you just get to focus on the modulation, without mixing it up with the radio parameters, which took me a long time to understand when I started. Two questions with this idea:
1. Will this mean a deprecation of the `beginFSK()` and similar functions since this can now be an overload of `begin()` with `ConfigFSK_t`?
2. How would one set radio parameters such as the TCXO voltage? The previous implementation had that being part of e.g. `SX126x::ConfigLoRa_t`, but this new one cannot really support that without going back to having one huge struct across all radios.

jgromes:
@StevenCellist to answer the questions:

1. I was thinking about that as well, but it's not going to play nice with the argument lists, since `radio.begin({.frequency = 434});` will be ambiguous. It might still be a worthy trade-off though, it will just mean the users that want to use argument lists will have to specify more than just the common stuff like frequency. We will have to make sure there is no overlap between the different modulations.
2. Those I moved to member variables of their respective classes, e.g. `float SX126x::tcxoVoltage`. Like I wrote above, precedence for this exists already with `bool SX126x::XTAL`, sicne it was so annoying to have to write out all of the arguments for `begin` just to disable TCXO. I imagine those will be "hardware-related" configuration, specific to any given module

StevenCellist:
1. They would need to explicitly define the type of the argument, so you can't do a one-liner in that case indeed. Unless the modem argument is added.. but meh.
2. That works, the thing that I think of is that with the arguments being as they are now, you are somewhat forced (or at least get the pointers) to look at your hardware and wonder what it is using. You lose that a little if it is one of the one hundred possible functions in the Intellisense list. But then the same goes for RF switch and some other stuff...

jgromes:
> you are somewhat forced (or at least get the pointers) to look at your hardware and wonder what it is using

Unfortunately users have proven to be quite stubborn in the approach of "flash first, ask questions later". Even with the current approach, there was a lot of -706/-707 issues, until I added the automated XTAL detection. I agree it kind of obfuscates the hardware configuration though, so it probably deserves a dedicated wiki page.

StevenCellist:
That would be a good idea :)

jgromes:
@StevenCellist FYI I figured out how to generate a Doxygen page for the hardware configuration member variables, which is quite handy:

https://jgromes.github.io/RadioLib/group__module__config__vars.html

I also updated the Basics wiki page.

StevenCellist:
Ooh that is amazing! I hope people will find it easily enough but looks great to me :)

PR #1786, opened 2026-05-22
LR1121: override getVersionInfo to skip WiFi/GNSS reads

ionspark9000:
## Problem

`LR1121::begin()` returns `RADIOLIB_ERR_CHIP_NOT_FOUND (-2)` on real
LR1121 silicon even though the chip is alive and the `GetVersion`
command returns the correct device ID `0x03`.

Repro on a Lilygo T3S3 LR1121 board with this library at 6.5.0 and
6.6.0:

```
RLB_DBG: LR11x0 not found! (1 of 10 tries) RADIOLIB_LR11X0_CMD_GET_VERSION = 0x03
RLB_DBG: Expected: 0x03
... (10 attempts, every one reports the correct device byte)
RLB_DBG: No LR11x0 found!
```

## Root cause

`LR11x0::findChip(uint8_t ver)` calls `getVersionInfo(&info)` which
issues three SPI commands:

- `GetVersion` — works on every LR11x0 family member
- `WifiReadVersion` — implemented on LR1110/LR1120 only
- `GnssReadVersion` — implemented on LR1110/LR1120 only

On LR1121 the WiFi/GNSS commands return errors. `RADIOLIB_ASSERT`
propagates the error out of `getVersionInfo`, and `findChip` falls
into the not-found branch even though `info.device` is `0x03`.

This was anticipated — `LR1121.h` previously contained a `TODO` for
exactly this case:

```cpp
// TODO this is where overrides to disable GNSS+WiFi scanning methods on LR1121
// will be put once those are implemented
```

## Fix

Override `getVersionInfo()` in `LR1121` to call only `getVersion()`
and zero the LR1110/LR1120-only fields. Adds the `virtual` keyword
to `LR11x0::getVersionInfo()` so the override actually dispatches.

## Tested on

- Lilygo T3S3 with LR1121 module, sub-GHz HF, 869.85 MHz, EU868.
- Custom embedded firmware build. After the fix, `radio.begin()`
  returns `RADIOLIB_ERR_NONE`, the chip identifies as device `0x03`
  on the first attempt, and `setFrequency` / `setOutputPower` /
  `transmit` / `startReceive` all work as expected for a sub-GHz
  LoRa mesh.

## Notes

- No change to LR1110 or LR1120 behaviour — they continue using the
  inherited `LR11x0::getVersionInfo()` implementation.

### DIFF
--- a/src/modules/LR11x0/LR1121.cpp
+++ b/src/modules/LR11x0/LR1121.cpp
@@ -5,4 +5,26 @@ LR1121::LR1121(Module* mod) : LR1120(mod) {
   chipType = RADIOLIB_LR11X0_DEVICE_LR1121;
 }
 
+int16_t LR1121::getVersionInfo(LR11x0VersionInfo_t* info) {
+  if(!info) {
+    return(RADIOLIB_ERR_MEMORY_ALLOCATION_FAILED);
+  }
+
+  // LR1121 only implements GetVersion. WiFi-scanning and GNSS-scanning
+  // firmware blocks don't exist on this silicon, so the corresponding
+  // SPI commands return errors. Reading just the GetVersion fields is
+  // sufficient for findChip() and any caller that wants the device
+  // identifier or base firmware version.
+  int16_t state = this->getVersion(&info->hardware, &info->device,
+                                   &info->fwMajor, &info->fwMinor);
+  RADIOLIB_ASSERT(state);
+
+  // Zero the LR1110 / LR1120-only fields so callers don't see garbage.
+  info->fwMajorWiFi = 0;
+  info->fwMinorWiFi = 0;
+  info->fwGNSS      = 0;
+  info->almanacGNSS = 0;
+  return(RADIOLIB_ERR_NONE);
+}
+
 #endif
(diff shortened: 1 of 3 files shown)


jgromes:
Current version of `getVersionInfo` already catches this internally:

https://github.com/jgromes/RadioLib/blob/a90197d4661579aebfa076bbf31b224381d83d44/src/modules/LR11x0/LR11x0.cpp#L1384-L1391

> Repro on a Lilygo T3S3 LR1121 board with this library at 6.5.0 and 6.6.0:

That's great, but the current version is 7.6.0. Both versions you have mentioned are more than 2 years old, so the first thing I would try is an update.

ionspark9000:
> Current version of `getVersionInfo` already catches this internally:
> 
> https://github.com/jgromes/RadioLib/blob/a90197d4661579aebfa076bbf31b224381d83d44/src/modules/LR11x0/LR11x0.cpp#L1384-L1391
> 
> > Repro on a Lilygo T3S3 LR1121 board with this library at 6.5.0 and 6.6.0:
> 
> That's great, but the current version is 7.6.0. Both versions you have mentioned are more than 2 years old, so the first thing I would try is an update.

You're right — apologies, I was on 6.6.0 (what `espressif32@^6.5.0` 
currently pulls in via PlatformIO) and didn't think to check master. 
Confirmed the LR1121 branch at LR11x0.cpp:1384-1391 covers this case 
already. Closing this PR as redundant; will upgrade my local toolchain 
to a current release. Thanks for the quick reply and for maintaining 
RadioLib.

ionspark9000:
Late to party.

PR #1787, opened 2026-05-24
[STM32WL] Add Cube platform, use Arduino platform only when used by STM32duino

StevenCellist:
Up until now, the STM32WL support in RadioLib required using STM32duino. However, as I'm using STM32Cube, this of course does not work.

This PR hides the appropriate parts behind STM32duino macros and moves the `STM32WLx_Module` files to a designated Hal folder. With this, I can now supply an STM32Cube-specific HAL and use RadioLib without any further modifications.
Also, the STM32CubeWL platform is added to `BuildOpt.h`. Cube does not add its own macro equivalent to the `ARDUINO_ARCH_*` macros, so `STM32WLE5xx` is the best I can use.

Verified by compiling and running with Cube, and compiling for STM32duino.

### DIFF
--- a/src/BuildOpt.h
+++ b/src/BuildOpt.h
@@ -465,6 +465,26 @@
     #define RADIOLIB_ARDUINOHAL_INTERRUPT_MODE_CAST
   #endif
 
+#elif defined(STM32WLE5xx)  // Cube does not expose a more explicit macro
+  #define RADIOLIB_PLATFORM                           "STM32WLE5xx using STM32Cube"
+  #define STM32WLE5xx_CUBE                            // define a more explicit macro
+
+  #define RADIOLIB_NC                                 (0xFF)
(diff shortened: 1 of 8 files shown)


StevenCellist:
Since STM32 allows reading from flash directly, the error is not a real error I think but I'm not sure. Otherwise we would need some suppression maybe.

jgromes:
Cppcheck is just complaining that you are using C casting in the nonvolatile macros, instead of `reinterpret_cast` - feel free to to reuse the ones that are in the same file for the generic platform, that should work ;)

The only comment I have is whether it would make sense to also catch compilation for STM32WL55 - IMO the interface should be exactly the same as there is only a single STM32WL HAL driver (https://github.com/STMicroelectronics/stm32wlxx-hal-driver).

PR #1791, opened 2026-05-28
Implement SSDV

### LINKED ISSUE #129: Additon of SSDV protocol
A way to use ssdv with rf modules with the ease with which we are able to use AX.25, Morse, RTTY client etc.

Dave Akerman's lora ssdv

**Additional context**
reference :- https://github.com/fsphil/hadie

StevenCellist:
See the pictures at https://ssdv.habhub.org/ 
The included `image.h` is the 256×256 RadioLib logo.

Reception is not implemented. I have a file that receives the LoRa packets and parses the contents. It should be possible to reconstruct the image on-device but I have no incentive to implement this. But happy to add a function and example that returns a struct with the parsed info when supplied with a LoRa packet.

Please have a look at the licensing of the included files. And maybe Reed-Solomon should live in the `utils` directory?

Closes #129.


StevenCellist:
With the power of an LLM there is now also an example that receives and decodes the JPEGs. LLMs are stupid and had to correct some very weird stuff and reduce lotsss of comments. But hey, otherwise we wouldn't have had this example.

src/protocols/SSDV/SSDV.cpp
```diff
@@ -0,0 +1,418 @@
...
+    return(RADIOLIB_ERR_NULL_POINTER);
+  }
+
+  // Free any previously allocated decoder state.
+  endDecoder();
+
+  // Heap-allocate the ssdv_t decoder state.
+  // ssdv_t is a large struct (~2 KB) containing Huffman and quantisation
+  // table arrays; heap allocation avoids blowing the stack.
+  this->decoder = reinterpret_cast<ssdv_t*>(malloc(sizeof(ssdv_t)));
```
jgromes, line 318: C++ `new`/`delete` should really by used here and elsewehre in the file instead of C `malloc`/`free`

src/TypeDef.h
```diff
@@ -663,6 +663,87 @@
...
+  \brief JPEG encoding failure. This usually means the JPEG is not 
+  in a format accepted by the encoder (progressive DCT, wrong colour space, 
+  dimensions not multiples of 16, etc.).
+*/
+#define RADIOLIB_ERR_SSDV_ENCODE_FAILED         (-1505)
+
+/*!
+  \brief malloc() returned NULL while allocating the packet buffer.
+*/
+#define RADIOLIB_ERR_SSDV_ALLOC_FAILED          (-1506)
```
jgromes, line 698: We ahve a generic error code for failed memory allocations, `RADIOLIB_ERR_MEMORY_ALLOCATION_FAILED` - I think a new error code is not needed.

jgromes (commented):
Thank you, I only have two comments to the code. Also one general comment - I would really prefer to have the portions of AI-geenrated code marked in some way (e.g. as a comment or note in the doxygen method description).

StevenCellist:
@jgromes the `ssdv_enc` library is now merged into RadioLib's style/format with pretty much all of the AI stuff gone. I left a reference to the license on the original version, I'm not sure what would be best. Examples are verified with the included RadioLib logo.
Please check the use of `new` and `delete` and if static memory should be included, I have very little experience there as I'm almost always working with fixed buffer sizes myself.

src/TypeDef.h
```diff
@@ -663,6 +663,82 @@
 */
 #define RADIOLIB_ERR_ADSB_INVALID_CATEGORY                      (-1401)
 
+// ── SSDVClient-specific error codes ───────────────────────────────────────
+/*!
+  \brief Callsign is empty or longer than 6 characters.
+*/
+#define RADIOLIB_ERR_SSDV_CALLSIGN_TOO_LONG                     (-1501)
```
jgromes: Usually I start the error codes at index 0 (so -1500 in this case), is there a specific reason that's not the case here?

PR #1793, opened 2026-06-01
feat(hal): Add ESP-IDF HAL (EspHal) as a library HAL

x42en:
## Summary

This promotes the ESP-IDF HAL from the `examples/NonArduino/ESP-IDF` example
into a proper library HAL at `src/hal/ESP-IDF/EspHal.h`, alongside the existing
Arduino / RPi / RPiPico / Stm32duino HALs.

Currently ESP-IDF users have to copy `EspHal.h` out of the example folder into
their own project. With this change the HAL ships with the library and is
auto-included from `RadioLib.h` when building the generic (non-Arduino) variant
on an `ESP_PLATFORM` target:

```cpp
#if defined(RADIOLIB_BUILD_ARDUINO)
#include "hal/Arduino/ArduinoHal.h"
#elif defined(RADIOLIB_BUILD_GENERIC) && defined(ESP_PLATFORM)
#include "hal/ESP-IDF/EspHal.h"
#endif

### CODE REVIEW

**Comment by jgromes** [comment]
Thanks for this contribution! I have a couple of questions about it though:

1. How portable is this code across the different ESP32 (or even ESP8266?) variants, like ESP32-C3, ESP32-C6 etc.? I'm asking because so far, there always seemed to be some target-dependent configuration e.g. for the SPI (see [Arduino ESP32 library](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-spi.c#L747) for example). So I always assumed it would not be worth it go to the trouble of having a very portable ESP32 HAL in here. That's why it is only in the example (and why the readme states that some modification may be needed).
2. How was this tested?

**Comment by x42en** [comment]
Hi, and thanks for the thorough review — both are fair questions.

## 1. Portability across ESP32 variants

The key difference from the example HAL is that **this implementation does not touch the SPI peripheral registers at all**. The example in EspHal.h drives SPI bare-metal: it pokes `spi_dev_t` mapped at `DR_REG_SPI2_BASE`, enables the peripheral via `DPORT_*` registers, routes pins through the GPIO matrix with hard-coded signal indices (`HSPICLK_OUT_IDX`, `HSPIQ_OUT_IDX`, `HSPID_IN_IDX`), and computes the clock divider from the ESP32 clock-register layout. All of that is silicon-specific — the register structs, the GPIO-matrix indices and the clock tree differ on the S2/S3/C3/C6/H2 — which is exactly why that file carries the `#if CONFIG_IDF_TARGET_ESP32 == 0 → #error` guard and the README caveat. The link you posted (`esp32-hal-spi.c`) is the Arduino core doing the same per-target register work itself.

This HAL instead goes one layer up and uses the **ESP-IDF `spi_master` driver** (`driver/spi_master.h`):

- `spi_bus_initialize()` / `spi_bus_add_device()` for setup,
- `spi_device_polling_transmit()` for transfers,
- `driver/gpio.h`, `esp_timer.h`, `esp_rom_delay_us()` for GPIO/timing.

These are precisely the APIs Espressif maintains as the per-target abstraction, so the target-dependent SPI configuration you're referring to is handled **inside the driver**, not in this file. As a result there is **no `#error`, no `CONFIG_IDF_TARGET` branch, and no register/`DPORT`/`soc/*_struct.h` access anywhere** — the only platform guard is `#if defined(ESP_PLATFORM)`. The SPI host is a constructor parameter (`SPI2_HOST` by default, `SPI3_HOST` selectable), and the bus init tolerates an already-initialized bus (`ESP_ERR_INVALID_STATE`) so it can share a bus with other peripherals.

So in principle it should compile and run on any ESP-IDF target that exposes a general-purpose SPI host (ESP32, S2, S3, C3, C6, H2, …). **ESP8266 is out of scope**: it is not an ESP-IDF v5 target and has a different SPI driver/RTOS SDK, so this HAL is not intended for it.

The trade-off vs. the bare-metal example is a small amount of driver overhead per transfer, in exchange for being target-agnostic — which seems like the right balance for a library HAL rather than an example.

## 2. Testing

Honest disclosure on coverage:

- **Compiled and run on hardware**: ESP32 (original), ESP32-S3, and ESP32-C6, driving a CC1101 over SPI.
- **Used in production**: ESP32-S3 only.
- The other variants (S2, C3, H2) are expected to work by virtue of using only the portable `spi_master`/`gpio`/`esp_timer` APIs, but I have **not** tested them on hardware, so I can't claim verified support for those.

If you'd prefer, I'm happy to make the supported/tested targets explicit (e.g. a note in the header and README), or to keep an `ESP_PLATFORM`-only guard and document the tested set, rather than implying universal support. Let me know how you'd like to scope the claim.

File: src/hal/ESP-IDF/EspHal.h [hunk 1]
```diff
@@ -0,0 +1,396 @@
...
+      ESP_LOGE("EspHal", "Failed to remove SPI device: %s",
+               esp_err_to_name(ret));
+      return ret;
+    }
+
+    spiDevice = nullptr;
+    deviceAdded = false;
+    ESP_LOGD("EspHal", "SPI device removed");
+    return ESP_OK;
+  }
```
Steve-Tech: Adding `addSpiDevice` and `removeSpiDevice` would require changes to RadioLib and the rest of the other HAL's wouldn't it? I think it would be better to just add a device as part of `spiBegin` and leave the `spics_io_num` as `-1`.

Weirdly, I also started making an ESP-IDF HAL yesterday, and just saw this PR.

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,396 @@
...
+    bus_config.data4_io_num = -1;
+    bus_config.data5_io_num = -1;
+    bus_config.data6_io_num = -1;
+    bus_config.data7_io_num = -1;
+    bus_config.max_transfer_sz = 4096;
+    bus_config.flags = 0;
+    bus_config.isr_cpu_id = ESP_INTR_CPU_AFFINITY_AUTO;
+    bus_config.intr_flags = 0;
+
+    esp_err_t ret = spi_bus_initialize(spiHost, &bus_config, SPI_DMA_CH_AUTO);
```
Steve-Tech: For DMA you should allocate the buffers in DMA capable memory ([source](https://docs.espressif.com/projects/esp-idf/en/v6.0.1/esp32s3/api-reference/peripherals/spi_master.html#write-and-read-phases)), I don't see RadioLib doing that and I don't really see the point, so I think it's better to just set `SPI_DMA_DISABLED` and `max_transfer_sz = SOC_SPI_MAXIMUM_BUFFER_SIZE`.

x42en:
Hey @Steve-Tech great catches on both, and thanks for taking the time — you're right on both points!

**On `addSpiDevice` / `removeSpiDevice`:** you're correct, these are extra public methods outside the `RadioLibHal` interface, and RadioLib core never calls them — only the application does. They're not actually needed: RadioLib's `Module` already drives the CS pin in software (`hal->digitalWrite(csPin, …)` around each transaction), so configuring `spics_io_num = csPin` in the device is redundant. I'll rework it to add a single device inside `spiBegin()` with `spics_io_num = -1` and let RadioLib handle CS, which drops both extra methods and keeps the HAL aligned with the others. The shared-bus support (tolerating an already-initialized bus via `ESP_ERR_INVALID_STATE`) stays in `spiBegin()`, since that's the part that actually needs it.

**On DMA:** agreed. RadioLib doesn't allocate DMA-capable buffers, and the multi-byte `spiTransfer()` currently hands RadioLib's own buffers straight to the driver — which aren't guaranteed to be DMA-capable — so `SPI_DMA_CH_AUTO` is a latent footgun rather than a win. The transfers here are tiny (1–2 byte register access, ≤64-byte FIFO bursts), so there's nothing to gain from DMA anyway. I'll switch to `SPI_DMA_DISABLED` with `max_transfer_sz = SOC_SPI_MAXIMUM_BUFFER_SIZE`.

And no worries about the timing 😄 — this HAL actually started life as a fork back in January 2025; I just never got around to pushing it upstream until now. Funny that we both landed on an ESP-IDF HAL within a day of each other. Happy to compare notes if you'd like — if your version covers cases mine doesn't, I'd rather we converge on one good HAL than ship two.

Steve-Tech:
> these are extra public methods outside the RadioLibHal interface, and RadioLib core never calls them — only the application does.

Oh, I didn't think of the application calling things, but I wouldn't have liked that either as that strays from the examples and might confuse people. But the changes you've made now seem fine.

> Funny that we both landed on an ESP-IDF HAL within a day of each other.

Yeah, I know haha!

> if your version covers cases mine doesn't

Nah, mine was quite basic, and yours handles SPI bus sharing nicely, which mine didn't.

---

I do have a few minor nitpicks though:

* I'd prefer `this->digitalRead` instead of `digitalRead`
* `clockHz = 500000` seems quite slow, the existing example runs at 2MHz, I was just using `SPI_MASTER_FREQ_8M`. Is there a module supported by RadioLib that requires that?
* `spiTransferByte` isn't used by anything (though `SPI_TRANS_USE_TXDATA` is quite neat)
* `#define NOP()` isn't used either

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,370 @@
...
+    // ESP32 doesn't have a built-in tone() function
+    // This would require LEDC (LED PWM Controller) implementation
+    // For now, we provide a stub that does nothing
+    // Full implementation would use ledc_timer_config() and
+    // ledc_channel_config()
+    (void)frequency;
+    (void)duration;
+    ESP_LOGW("EspHal",
+             "tone() not fully implemented - requires LEDC peripheral setup");
+  }
```
Steve-Tech, line 228: I think you should copy the way the Arduino HAL for ESP32 works for Tone and noTone: https://github.com/jgromes/RadioLib/blob/034126ef3b5305394d1e4e14a5049482ec10c1c4/src/hal/Arduino/ArduinoHal.cpp#L126-L143

Basically initialises everything if it hasn't been called before.

It might also be a good idea to get `RADIOLIB_ESP32` and `RADIOLIB_TONE_ESP32_CHANNEL` defined, as they were undefined for me.

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,370 @@
...
+  void attachInterrupt(uint32_t interruptNum, void (*interruptCb)(void),
+                       uint32_t mode) override {
+    if (interruptNum == RADIOLIB_NC) {
+      return;
+    }
+
+    gpio_set_intr_type((gpio_num_t)interruptNum, (gpio_int_type_t)mode);
+
+    // Install ISR service only if not already installed (global service)
+    esp_err_t ret = gpio_install_isr_service(ESP_INTR_FLAG_IRAM);
```
Steve-Tech: > Install ISR service only if not already installed

This doesn't seem hard to add, just wanna make sure it doesn't get forgotten.

Currently, installing it multiple times causes an error to be logged: `gpio: gpio_install_isr_service(530): GPIO isr service already installed` (at least for ESP-IDF 5.5.3)

Steve-Tech:
Also, your replies look like they've been generated by an LLM, as per the [CONTRIBUTING.md](https://github.com/jgromes/RadioLib/blob/master/CONTRIBUTING.md#ai-use) you do need to disclose this, and I would like to know how involved it is in the creation of your commits.

x42en:
Hi, thanks a lot for your review which point several points! I've manually corrected the basics:
- fair point for digitalRead usage
- clockHz speed: I've bumped the default to 2 MHz to match the NonArduino example and kept it as a constructor parameter. But I didn't choose to default to SPI_MASTER_FREQ_8M because for exemple CC1101-class modules cap SCLK at 6.5 MHz, so 8 MHz would be out of spec
- Removed the spiTransferBytes and migrate the usage to my app

Then I used Claude 4.8, as an old and lazy developer, to generate the Tone and NoTone implementation mainly because that's not something I'm relying on, and because I'm not confident about my complete understanding about this. :wink:. Isr guard has been a suggestion of claude that I did not modify has it is close to something I could wrote but include comments better written that mines!

Concerning the usage of IA, sure I'm using LLM to help me in my everyday life, and open-source projects are a big part of it. I'm not a native speaker (you might I've seen that with this comment :smile: ) and Claude speaks far better english than me, it also generate nice markdown report whith bold stuff and many other enhancements that I don't take time to add.

Note: I found your way to point it to me a *bit severe*. I would have preferred something like `Hi, we have updated our CONTRIBUTING.md few month ago concerning IA usage, could you please double check it to comply with rules ?` ... I've been using RadioLib for years now in my projects I've to confess that I do not read every commit, and I didn't re-read CONTRIBUTING since a while :D I've done it now and we might have a problem here: my implementation is quite old and has probably been modified/contaminated by multiple IA pass around my codebase. I could not `mark the slop`, as you write, with accuracy. But be sure I know and fully-understand my code and I read and validate implementation before using it. Man I understand the frustration every member of the open-source community feels about having it's code stolen by billion-dollar companie to sell stuff, but it looks like we've lost this battle, period. Now it's time to use this kind of tools to enhance our projects, and help us going quicker for review and commit generation rather than being scared of using it. Github has been acquired by Microsoft for a reason: even your private repositories has been parsed, and this is no longer an open-source platform like in the early days spirit! By my side I'm now burning tokens, money and time to open-source projects just for fun and no-profit :grinning: We're living in a weird momentum!

Steve-Tech:
> Note: I found your way to point it to me a *bit severe*.

Oh, I'm sorry I came across that way. That's mostly just how I've seen it written/handled in academia.

I am not a contributor to RadioLib, I just want a good ESP-IDF HAL for my own RadioLib projects and wanted to offer some feedback. I saw that part in the CONTRIBUTING.md and thought I should bring it up. I also wanted to make sure you weren't feeding my comments straight in an LLM, because I can do that too if I really want; but I do believe that you understand what you're doing.

My interpretation of "Mark the slop", is it literally just needs to be disclosed somewhere, I didn't think you would need to mark exactly where AI was used.

> But I didn't choose to default to SPI_MASTER_FREQ_8M because for exemple CC1101-class modules cap SCLK at 6.5 MHz, so 8 MHz would be out of spec

Perfect! I was using an SX1262 which goes up to 16MHz when I wrote mine, but I'm glad you've checked.

Other than the `RADIOLIB_ESP32` comment, I am overall happy with this and don't have any more feedback.

Thanks for your work!

src/hal/ESP-IDF/EspHal.h
Steve-Tech: I was hoping you could add `|| defined(ESP_PLATFORM)` to BuildOpt.h here:
https://github.com/jgromes/RadioLib/blob/4ea54fe3cea720e2b49c90d92f284c2f3fc576df/src/BuildOpt.h#L235

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,381 @@
...
+    if (interruptNum == RADIOLIB_NC) {
+      return;
+    }
+
+    gpio_set_intr_type((gpio_num_t)interruptNum, (gpio_int_type_t)mode);
+
+    // The GPIO ISR service is global to the whole application; install it only
+    // once to avoid the driver logging "already installed" on repeated calls.
+    static bool isrServiceInstalled = false;
+    if (!isrServiceInstalled) {
```
jgromes: I don't understand why this needs to be guarded in this way - is it not safe to call `gpio_install_isr_service` repeatedly and checking its return code?

Additionally, this will only guard against repeated calls to `attachInterrupt`, not against user themselves calling `gpio_install_isr_service` manually.

EDIT: Which is exactly what happens in the first branch of the if. So this check is redundant.

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,381 @@
+// EspHal_spimaster.h
```
jgromes: What is `EspHal_spimaster.h`? I'm guessing the original name of this file, but that seems outdated now.

src/hal/ESP-IDF/EspHal.h
```diff
@@ -0,0 +1,381 @@
...
+// ESP-IDF HAL for RadioLib using spi_master driver
+// This avoids conflicts with other SPI devices on the same bus
+
+#ifndef ESP_HAL_SPIMASTER_H
```
jgromes: Maybe something like `RADIOLIB_ESP_IDF_HAL` would be better - I imagine `ESP_HAL_SPIMASTER_H` may be defined somewhere within ESP-IDF.

jgromes:
Thanks @Steve-Tech for stepping in for the review while I was busy with other stuff, for sure it is very good to have more eyes looking at this as the ESP32 is likely the most widespread platform running RadioLib.

@x42en your usage of LLMs is clear from your comments and personally, I would much rather read a comment with mistakes written by an actual human, than an overly long machine-generated statement. I am not a native English speaker either and yet here I am, typing these letters with meaty fingers.

The AI guideline is indeed new; that is self-evident, AIs are new(-ish) themselves. Before moving on with this PR, I'm going to need you to do the following:

1. I understood that only your original `tone`/`noTone` have been generated by an AI. Please confirm that is the case.
2. It should go without saying that the CI must pass; my guess is that it is failing due to the original HAL remaining in the example folder, but I have not checked that myself.
3. I left some comments, though mostly minor ones.

Thanks!

src/hal/ESP-IDF/EspHal.h
Steve-Tech: I found when using LoRaWAN with a SX1262 (and probably others), `receiveClassA`, `receiveClassC`, and `getDownlinkClassC` seem to call these functions:

https://github.com/jgromes/RadioLib/blob/4ea54fe3cea720e2b49c90d92f284c2f3fc576df/src/modules/SX126x/SX126x_config.cpp#L12-L26

Causing this error to be logged every Rx window:
`gpio: gpio_install_isr_service(530): GPIO isr service already installed`
https://github.com/espressif/esp-idf/blob/835c94776160692a48acc5f07dabdfb21f7616fe/components/esp_driver_gpio/src/gpio.c#L568

Like you are right that it is safe to call `gpio_install_isr_service` repeatedly, but I didn't like the error that it logs.

jgromes: @Steve-Tech the LoRaWAN stack should be calling `clearPacketReceivedAction` whnever the interrupt is no longer needed, and it assumes the HAL takes care of fully clearing the interrupts. `detachInterrupt` in here however does not uninstall the ISR service. I don't know if that is possible in ESP-IDF, but if not then I would move the ISR service install into `init` - it's not really possible to use the library without interrupts anyway (except for very basic blocking Tx).

Steve-Tech: > `detachInterrupt` in here however does not uninstall the ISR service. I don't know if that is possible in ESP-IDF

It is possible, using `gpio_uninstall_isr_service`.

> I would move the ISR service install into `init`

I prefer this though, in case interrupts are already being used.

x42en:
> Thanks @Steve-Tech for stepping in for the review while I was busy with other stuff, for sure it is very good to have more eyes looking at this as the ESP32 is likely the most widespread platform running RadioLib.
> 
> @x42en your usage of LLMs is clear from your comments and personally, I would much rather read a comment with mistakes written by an actual human, than an overly long machine-generated statement. I am not a native English speaker either and yet here I am, typing these letters with meaty fingers.
> 
> The AI guideline is indeed new; that is self-evident, AIs are new(-ish) themselves. Before moving on with this PR, I'm going to need you to do the following:
> 
> 1. I understood that only your original `tone`/`noTone` have been generated by an AI. Please confirm that is the case.
> 2. It should go without saying that the CI must pass; my guess is that it is failing due to the original HAL remaining in the example folder, but I have not checked that myself.
> 3. I left some comments, though mostly minor ones.
> 
> Thanks!

Hi, sorry for the delay I was AFK!

Regarding your questions:
1. Yes I do confirm the AI generation of `tone`/`noTone` code part
2. I've tried the CI on my fork, except the multiple warning about Node v20 deprecation it seems good (at least `esp-build` run successfully by my side)
3. I think I've covered them all

PS: I'm using RadioLib in a project where multiple radio modules are used on shared SPI bus with other components, actually your suggestion simplified a lot my implementation! I don't know why I didn't came to it before... the power of review :wink:

jgromes:
It looks like all the comments have been addressed - so merged, thank you @x42en for the contribution!

LoopAxel:
To start of with, this may not be the correct place to post this, sorry if that is the case I don't do this often. Though I spent way to much time on this (bug?) to not report it. 
The problem is in the EspHal.h implementation of delay(), vTaskDelay(...); is used. vTaskDelay delays the time in freertos ticks, and default configuration is 1 tick = 10ms which makes the accuaracy of the delay terrible. Which in my case forced me to use a higher scanGuard when using LoRaWAN.

I solved this using esp_timer instead of vTaskDelays and so far it seems to work as intended. Here is the new implementation of delay:

  void delay(unsigned long ms) override {
    if (ms == 0) return;
    if (!_delayTimer) {
      vTaskDelay(pdMS_TO_TICKS(ms) + 1);
      return;
    }
    _delayTaskHandle = xTaskGetCurrentTaskHandle();
    ulTaskNotifyTake(pdTRUE, 0);           
    esp_timer_stop(_delayTimer);           
    esp_timer_start_once(_delayTimer, (uint64_t)ms * 1000ULL);
    ulTaskNotifyTake(pdTRUE, pdMS_TO_TICKS(ms + 500));
    _delayTaskHandle = nullptr;
  }

"vTaskDelay(pdMS_TO_TICKS(ms) + 1);" Adding one tick here is needed because calling pdMS_TO_TICKS(...) with a value less than 10ms returns 0, which for me caused greater problems then waiting longer than asked for. Ideally this code only is used if initialization of esp_timer fails, which never should happen (?).

This may not be the best solution, and it certainly isn't the prettiest. There is some more initialization code provided in the attached file, I'm sure someone else can make a better job than me implementing this. Also, I'd guess that changing CONFIG_FREERTOS_HZ to 1000 instead of default 100 would fix the problem, though I don't really see that as a viable soloution. 

Didn't manage to attach the file as plain text. Therefor a zip.
[EspHal.zip](https://github.com/user-attachments/files/29285096/EspHal.zip)

Steve-Tech:
I'm not too experienced with FreeRTOS, but wouldn't `ulTaskNotifyTake` just block the task until the next kernel tick? Making that approach not any better.

I think really our only option would be something like:
```cpp
    void delay(unsigned long ms) override {
        if (pdMS_TO_TICKS(ms) > 0) {
            vTaskDelay(pdMS_TO_TICKS(ms));
        } else {
            this->delayMicroseconds(ms * 1000UL);
        }
    }
```
But maybe with a more efficient way to delay.

Also, you should bring this up in #1812.

PR #1798, opened 2026-06-08
[ESP-IDF] Move EspHal in-tree and modernize on esp_driver_{gpio,spi}

bitandpin:
- Move EspHal from `examples/NonArduino/ESP-IDF/main/EspHal.h` to `src/hal/ESP-IDF/EspHal.{h,cpp}`, matching the Arduino / RPi / RPiPico HAL layout.
- Rewrite the implementation on top of `driver/spi_master.h` + `driver/gpio.h`. Drops the `DPORT_*` / `spi_dev_t` register pokes and the `CONFIG_IDF_TARGET_ESP32 == 0` hard-fail — HAL now   works on any ESP32 variant.
- Extend the `esp-build` CI job to a matrix over `esp32` / `esp32s3` / `esp32c3`.

Built locally on esp32, esp32s3, esp32c3 with ESP-IDF v5.5.3


bitandpin:
I see there is a huge overlap with existing PR targeting exaclty the same thing - improving support for ESP-IDF in RadioLib. I guess I was harboring this change for too long...

jgromes:
Indeed, there is a currently open #1793 - then again, that one is still undergoing review, so it might happen we return to this one ;)

Issue #1799, opened 2026-06-08
LR2021: expose detector index from GET_LORA_PACKET_STATUS


dasmensch-su:
The LR2021 supports multi-SF reception via side detectors configured through `setSideDetector()`. When a packet is received, the chip reports which detector (main or side 1/2/3) caught it in the response to `GET_LORA_PACKET_STATUS`. This information is currently read into `buff[5]` but only the RSSI LSBs (bits 1:0) are extracted -- the detector field in bits 5:2 is discarded.

Exposing this field would complete the multi-SF receive path: users can configure side detectors, receive packets on different spreading factors, but currently have no way to determine which detector/SF a given packet arrived on.

### Proposed change

Add an optional 7th parameter `uint8_t* detector = NULL` to `getLoRaPacketStatus()`:

**LR2021.h:**
int16_t getLoRaPacketStatus(uint8_t* cr, bool* crc, uint8_t* packetLen = NULL,
    float* snrPacket = NULL, float* rssiPacket = NULL,
    float* rssiSignalPacket = NULL, uint8_t* detector = NULL);

**LR2021_cmds_lora.cpp** (after the rssiSignalPacket block, before `return(state)`):
if(detector) { *detector = (buff[5] >> 2) & 0x0F; }

The value is a one-hot bitmask per the datasheet:
- `0x01` -- Main detector
- `0x02` -- Side detector 1
- `0x04` -- Side detector 2
- `0x08` -- Side detector 3

### Notes

- Fully backward-compatible: the new parameter defaults to `NULL`, so existing callers are unaffected.
- The data is already in `buff[5]` from the existing SPI read -- no additional chip communication needed.
- Follows the same optional-out-pointer pattern used by the other parameters in this function.
- We have already implemented this patch in our project and verified it is functional on hardware with multi-SF reception across 4 simultaneous spreading factors: https://github.com/dasmensch-su/Meshtastic-nRF54L15-LR2021-Multi-SF

jgromes:
I think this can be added - if you want, feel free to submit a pull request. The only thing I would change is to translate the one-hot encoding to 0 - 3 value.

dasmensch-su:
Thanks. Closing this since PR has been submitted.

PR #1800, opened 2026-06-09
[LR2021] Expose detector index from getLoRaPacketStatus

dasmensch-su:
Adds optional 7th parameter to report which detector (0 = main, 1-3 = side) received the packet. One-hot value from buff[5] bits 5:2 is translated to a 0-3 index. Defaults to NULL for backward compatibility.

Code has been tested with Meshtastic compiled to run on Semtech LR2021 LoRa Plus Evaluation Kit-US915 ShortFast Ch 15  with Cross-SF communication between it and a stock Meshtastic 2.7.16 running on Heltec WiFi LoRa 32 v4  LongFast Ch 15 and a stock Meshtastic 2.7.15 on Muzi Duo MediumFast Ch 15.

PR for feature discussed in https://github.com/jgromes/RadioLib/issues/1799 

Code was created with assistance from Claude AI.

### DIFF
--- a/src/modules/LR2021/LR2021.h
+++ b/src/modules/LR2021/LR2021.h
@@ -803,9 +803,10 @@ class LR2021: public LRxxxx {
       \param snrPacket SNR of the last received packet in dB
       \param rssiPacket RSSI of the last received packet in dBm
       \param rssiSignalPacket Estimation of the RSSI of LoRa signal after despreading in dBm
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks for the contribution! It looks OK (though I did spend some time thinking about a better way to convert from the one-hot encoded value to the position), it's nto really worth it going into detail about. 

Thanks for also being transparent about the AI usage, though it seems to me rather funny that the commit mesage probably cost more tokens than the actual code change :)

Merged, thanks!

Issue #1591, opened 2025-09-05
SX126x.getTimeOnAir - LDR is always applyed, even if the user disable it


MarusGradinaru:
Hi !
I think there may be a problem here: [SX126x.getTimeOnAir source code](https://github.com/jgromes/RadioLib/blob/master/src/modules/SX126x/SX126x.cpp#L1369)

The SX1262 chip has a "Low Data Rate" optimization, which is recomended to be applyed when symbol length is greater than 16ms. The RadioLib take care of this too, but it seems that, when the symbol length is >16, the optimization is always applyed, event is the user used `forceLDRO(false)` to disabele it. Shouldn't we check the `this->ldroAuto` and `this->ldrOptimize` flags first ?

jgromes:
In the current implementation of `getTimeOnAir`, this is indeed a bug that will cause some slight difference for cases with forced-disabled LDRO - good catch!

However, as @GUVWAF mentioned in #1584, the current implementation of `getTimeOnAir` will be split between two methods, `calculateTimeOnAir` and `getTimeOnAir` to allow us to calcualte ToA for different modulation properties than what is currently configured. So I think it would be best to wait for this and make sure the LDRO is covered in this new approach. This is probably not currently causing major problems, so we can live with it for a while.

MarusGradinaru:
Ok, I'll watch it...

GUVWAF:
I’ll make sure to take this into account when I get to it.

PR #1801, opened 2026-06-11
fix(SX127x): update ldroEnabled in setSpreadingFactor so getTimeOnAir is correct at high SF

Jaimi5:
> Note: this PR was written with AI assistance (the fix and description were reviewed by a human).

### Problem

On SX127x, `getTimeOnAir()` underestimates LoRa airtime by one code block at SF11/SF12. For example at SF12 / BW125 / CR4-7 / explicit header / CRC on with a 24-byte payload it returns ~1582 ms instead of the correct ~1810 ms.

### Cause

`setBandwidth()` updates both the LDRO register and the cached `this->ldroEnabled` member. `setSpreadingFactor()` only writes the register — it never updates the member. `getTimeOnAir()` reads the member (`SX127x.cpp`: `pc.lora.ldrOptimize = this->ldroEnabled;`).

Since `begin()` calls `setBandwidth()` (while SF is still the default 9, so LDRO computes as off) *before* `setSpreadingFactor()`, the member stays `false` even though `setSpreadingFactor()` correctly turns the LDRO register on for SF11/SF12. The radio transmits fine; only the airtime estimate is wrong.

### Fix

Set `this->ldroEnabled` in `setSpreadingFactor()` next to the register write, mirroring what `setBandwidth()` already does. Applied to both `SX1278` (SX1276/77/78/79) and `SX1272` (SX1272/73). No transmission behavior changes — only the cached value `getTimeOnAir()` reads. The existing workaround (re-calling `setBandwidth()` after `begin()`) is no longer needed.

SX126x is unaffected: there both setters route through `setModulationParams()`, which recomputes `ldrOptimize` from current SF+BW each call. (#1591 was a separate prior LDRO fix on that family.)

### Repro

```cpp
SX1278 radio = new Module(/* pins */);
radio.begin(434.0, 125.0, 12, 7);       // SF12, BW125, CR4-7
Serial.println(radio.getTimeOnAir(24));  // before: ~1582000 us, after: ~1810000 us
```

### DIFF
--- a/extras/test/unit/tests/TestCalculateTimeOnAir.cpp
+++ b/extras/test/unit/tests/TestCalculateTimeOnAir.cpp
@@ -23,6 +23,7 @@ static std::vector<RadioConfig> allConfigs = {
 
     { "SX127x", RADIOLIB_MODEM_LORA, {.lora={6,125,6}}, {.lora={8,false,true,false}}, {7,23,98,156}, {23000,39000,115000,174000} }, // 20.61, 39.04, 115.84, 174.21 
     { "SX127x", RADIOLIB_MODEM_LORA, {.lora={8,250,8}}, {.lora={32,true,true,false}}, {10,20,80,160}, {70000,87000,210000,373000} }, // 69.89, 86.27, 209.15, 372.99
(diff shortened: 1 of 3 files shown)


jgromes:
Thank you, the fix itself looks OK to be merged, however, I think it would also be worth it to add this case into the ToA calculation unit test: https://github.com/jgromes/RadioLib/blob/master/extras/test/unit/tests/TestCalculateTimeOnAir.cpp

jgromes:
@Jaimi5 thanks for the addition - merged!

PR #1805, opened 2026-06-17
[HAL] Change SPI bus initialization to use DMA channel auto

vshymanskyy:
Updated SPI bus initialization to use SPI_DMA_CH_AUTO for larger transactions.

Just tried to migrate to the new ESP-IDF HAL with my test project, and noticed it does not transmit payloads of more than roughly 60 bytes. I use payloads of 120 and 240 bytes in my tests and noticed it immediately. The proposed change fixes the issue and is tested on ESP32 and ESP32-C3 platforms.


vshymanskyy:
@jgromes @x42en

jgromes:
Very good catch, thanks! No comments on the change itself, so merged.

PR #1809, opened 2026-06-19
[LR11x0] Fix begin method for high band

vshymanskyy:
On LR1120, begin() fails for frequencies > 1000MHz (bandwidth validation fails).


jgromes:
Thank you for the contribution! It looks good - the other LR11x0 variants don't need this, so merged.

PR #1810, opened 2026-06-20
[LoRaWAN] Improve persistence and session handling (breaking) (#1789)

### LINKED ISSUE #1789: [LoRaWAN] Nonces signature not working as intended
The LoRaWAN nonces buffer contains most importantly the DevNonce and JoinNonce. A session is linked to a specific combination of Nonces using the signature from the Nonces buffer.
In a recent issue on the persistence repository it became apparent that this signature doesn't fulfill its purpose.

Upon each successful join, the DevNonce and JoinNonce both increase by one. In 50% of the cases, this will mean the same bit is flipped for both of these values. And as the JoinNonce in the nonces buffer is two bytes over from the DevNonce, and the signature uses a 16-bit XOR checksum, this results in an identical signature for a new session. 
Most people use the examples as intended which don't run into this issue, but I've seen setups that will try to restore a session from a previous set of nonces with new the nonces which may therefore falsely succeed.

There's a bunch of possible resolutions: use a differently sized checksum, apply some bitwise logic to one or more fields, use something else than a simple XOR, do not use the signature at all but reference the Nonces directly in the session buffer... I care little what it is as long as it is safe and is easy to understand.

This will break the Nonces (and Session) buffer but so does #1770 and we're heading towards a new major version anyway.

StevenCellist:
This PR proposes a rework of the persistence buffer (previously nonces buffer) including some collateral.

There are two main changes:
1. Modified the contents of the persistence buffer as outlined in fix #1789.
2. Moved to a callback system. 

Some packages need persistent values (TS004 and TS009), and having a separate buffer again for the package manager adds double work and overhead. Moreover, these packages - as well as LoRaWAN v1.2 to be released - may update persistent values while a session is running. 
With the current logic, the user would have to check continuously whether something changed and only then write, to prevent wearing NVS. This is very prone to mistakes and wear. With callbacks, the stack can dictate when to save in order to minimize the number of writes.

Passes user tests on a Heltec v3 with TTN; ready for review of the callback system, keeping as draft until verified by LCTT.

Note: the persistence repository must be updated when merging `8.0.0-staging` into `8.0.0`.  
Note2: implement a porting function to update existing persistence buffers to the new format.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -299,10 +299,13 @@ int16_t LoRaWANNode::sendReceive(const uint8_t* dataUp, size_t lenUp, uint8_t fP
 }
 
 void LoRaWANNode::clearPersistence() {
-  // clear & set all the device credentials
+  // clear all the persistent values
   memset(this->bufferPersist, 0, RADIOLIB_LORAWAN_PERSISTENCE_BUF_SIZE);
+  // reset TS004's fragmentation nonces to 0xFFFF to accept initial value 0
   memset(&this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS004], 0xFF, 8);
-  memset(&this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS009], 0x01, 1);
+  // allow enabling of TS009's test mode by setting to 1
+  this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS009] = 0x01;
+
   this->keyCheckSum = 0;
   this->devNonce = 0;
   this->joinNonce = 0;
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -203,6 +203,7 @@ int16_t LoRaWANNode::sendReceive(const uint8_t* dataUp, size_t lenUp, uint8_t fP
       // sometimes, a spurious error can occur even though the uplink was transmitted
       // therefore, just to be safe, increase frame counter by one for the next uplink
       this->fCntUp += 1;
+      this->saveSessionBuffer();
 
       #if !RADIOLIB_STATIC_ONLY
       delete[] uplinkMsg;
@@ -265,6 +266,7 @@ int16_t LoRaWANNode::sendReceive(const uint8_t* dataUp, size_t lenUp, uint8_t fP
 
   // if a hardware error occurred, return
   if(state < RADIOLIB_ERR_NONE) {
+    this->saveSessionBuffer();
     return(state);
   }
 
@@ -278,11 +280,16 @@ int16_t LoRaWANNode::sendReceive(const uint8_t* dataUp, size_t lenUp, uint8_t fP
     }
     // remove only non-persistent MAC commands, the other commands should be re-sent until downlink is received
     LoRaWANNode::clearMacCommands(this->fOptsUp, &this->fOptsUpLen, RADIOLIB_LORAWAN_UPLINK);
+    this->saveSessionBuffer();
     return(rxWindow);
   }
   
+  // parse the contents - if there is a parsing error, silently drop it
   state = this->parseDownlink(dataDown, lenDown, rxWindow, eventDown);
   RADIOLIB_ASSERT(state);
+
+  // parsed all nicely, save the buffer
+  this->saveSessionBuffer();
 
   // open RxC window (this returns if not applicable)
   this->receiveClassC();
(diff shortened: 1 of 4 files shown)


StevenCellist:
Ermm, actually, I have more changes lined up (among which tackling the other issues) and I am not a git master. I will verify with LCTT later, let's get this merged now first please :D

src/protocols/LoRaWAN/LoRaWAN.cpp
```diff
@@ -278,76 +280,97 @@ int16_t LoRaWANNode::sendReceive(const uint8_t* dataUp, size_t lenUp, uint8_t fP
...
   return(rxWindow);
 }
 
-void LoRaWANNode::clearNonces() {
+void LoRaWANNode::clearPersistence() {
   // clear & set all the device credentials
-  memset(this->bufferNonces, 0, RADIOLIB_LORAWAN_NONCES_BUF_SIZE);
+  memset(this->bufferPersist, 0, RADIOLIB_LORAWAN_PERSISTENCE_BUF_SIZE);
+  memset(&this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS004], 0xFF, 8);
+  memset(&this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS009], 0x01, 1);
```
jgromes: Why are nono-zero defaults needed for the packages? Also memset of 1 has probably a lot more overhead than just `this->bufferPersist[RADIOLIB_LORAWAN_PERSISTENCE_TS009] = 0x01;` ;)

src/protocols/LoRaWAN/LoRaWAN.h
```diff
@@ -1289,4 +1334,20 @@ void LoRaWANNode::hton(uint8_t* buff, T val, size_t size) {
   }
 }
 
+template<typename T>
```
jgromes, line 1337: Is this really needed as a template? So far only `uint32_t` version seems used, and also `checkBufferCommon` makes an assumption that the checksum is 32 bits.

Usually checksums have a known bit width, or just a part of a longer checksum could be used (e.g. last 16 bits of a 32-bit checksum)

src/protocols/LoRaWAN/LoRaWAN.h
```diff
@@ -846,6 +860,18 @@ class LoRaWANNode {
...
+      NOTE: Do NOT use, for Packages only!
+    */
+    void getPersistencePackage(uint8_t pIndex, uint8_t* buff);
+
+    /*!
+      \brief Set the persistent buffer for a certain package.
+      NOTE: Do NOT use, for Packages only!
+    */
+    void setPersistencePackage(uint8_t pIndex, uint8_t* buff);
+
```
jgromes, line 874: If these are not to be exposed publicly, then these methods should be `private` and the class that needs it could be added to `LoRaWANNode` as `friend` class.

jgromes (commented):
Overall just minor comments from my side regarding the code.

I have an additional one regarding testing though, since the persistent buffer is provided by the user (via the callbacks, but still), I think it would  be good to also cover this in the fuzzing. I don't see much logicx flow that depends on the contents of the buffer, but there is still some, and some could be added in the future.

src/protocols/LoRaWAN/LoRaWAN.cpp
StevenCellist: Meh, I like style. But can change that of course.
Will add comments.

src/protocols/LoRaWAN/LoRaWAN.h
StevenCellist: True that we use only one, but I thought it would be nice to provide a templated type anyway since it doesn't add any code or complexity. What do you prefer?

src/protocols/LoRaWAN/LoRaWAN.h
StevenCellist: ... which requires to pull in all the package classes resulting in additional kilobytes of unused stuff.

StevenCellist:
I think we can't do better than trusting the buffer signature (checksum) here, honestly. There's not much else I can do to verify contents of the buffer, as they all could be pretty much random at some point in the lifetime of a device.
Maybe I'm missing something, let me know if you see something that should be tested.

src/protocols/LoRaWAN/LoRaWAN.h
jgromes: Ah, didn't realize but yeah - never mind then.

src/protocols/LoRaWAN/LoRaWAN.h
jgromes: If it doesn't increase the binary size (dramatically) then we can keep it as a template.

jgromes:
Added the fuzzing test and executed manually - it passes so now this is good to merged. Thanks!

PR #1812, opened 2026-06-22
[ESP-IDF] Follow-up improvements to EspHal

bitandpin:
These changes are AI augmented follow up to the EspHal merged in #1793, originating from late #1798.

These are mostly polishing and streamlining changes, to follow the formats from other HALs. File split + reorderng might be too much, but I still decided to go for it. CI is a real value.

- Include guard `RADIOLIB_ESP_IDF_HAL` to `_RADIOLIB_ESP_IDF_HAL_H` to match the leading-underscore convention used by `Hal.h` and `ArduinoHal.h`.
- Drop file-scope `LOW`/`HIGH`/`INPUT`/`OUTPUT`/`RISING`/`FALLING` macros. They were defined as aliases for the IDF gpio enums but, because `EspHal.h` is reachable from `RadioLib.h` on ESP-IDF builds, the names leaked into every TU and could silently substitute user identifiers with the same name. Pass the IDF enums directly to the `RadioLibHal` base constructor, the same way `PicoHal` does.
- Implement `yield()` with `taskYIELD()`. The default `RadioLibHal::yield()` is an empty no-op, so on ESP-IDF the long polling loops (LoRa receive, channel-activity-detection) never relinquish the CPU. With the default sdkconfig the task watchdog watches IDLE, which can be starved and reset the chip.
- Split `EspHal.h` into `.h` + `.cpp` to match the `ArduinoHal` / `PicoHal` convention. Keeps `driver/ledc.h`, `esp_log.h`, `freertos/*.h`, etc. out of every TU that includes `RadioLib.h` on ESP-IDF builds. Wires the new `.cpp` into the IDF component build and adds `esp_driver_ledc` to `REQUIRES`.
- Align method order with the `RadioLibHal` base class. Reorder so pure-virtual overrides come first (in `Hal.h` order), then the virtual-with-default overrides — matching how `ArduinoHal.h` is laid out. Pure reordering, no logic change.
- CI: extend the `esp-build` job to a matrix over `esp32`, `esp32s3` and `esp32c3`. The merged HAL claims portability across ESP32 variants; this proves it on every PR. `fail-fast: false` so one chip failing does not cancel the others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,16 @@ if(ESP_PLATFORM)
     REQUIRES esp_driver_gpio esp_driver_spi esp_driver_ledc esp_timer
   )
 
+  # EspHal::delay() uses vTaskDelay(pdMS_TO_TICKS(ms)); below 1 kHz tick rate
+  # small delays round to zero (delay(5) -> vTaskDelay(0)). arduino-esp32
+  # makes this a hard error; we warn so users can opt in to lower tick rates
+  # if they really know what they are doing.
+  if(DEFINED CONFIG_FREERTOS_HZ AND CONFIG_FREERTOS_HZ LESS 1000)
+    message(WARNING
+      "RadioLib EspHal: CONFIG_FREERTOS_HZ is ${CONFIG_FREERTOS_HZ}; "
+      "set it to 1000 for accurate delay() resolution.")
+  endif()
+
   return()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,9 +13,14 @@ if(ESP_PLATFORM)
   # required because ESP-IDF runs cmake in script mode
   # and needs idf_component_register()
 
+  # add the ESP-IDF HAL implementation, which was excluded by the filter above
+  file(GLOB RADIOLIB_ESP_IDF_HAL_SOURCES "src/hal/ESP-IDF/*.cpp")
+  list(APPEND RADIOLIB_SOURCES ${RADIOLIB_ESP_IDF_HAL_SOURCES})
+
   idf_component_register(
     SRCS ${RADIOLIB_SOURCES}
-    INCLUDE_DIRS . src 
+    INCLUDE_DIRS . src
+    REQUIRES esp_driver_gpio esp_driver_spi esp_driver_ledc esp_timer
   )
 
   return()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,16 +24,6 @@ if(ESP_PLATFORM)
     REQUIRES esp_driver_gpio esp_driver_spi esp_driver_ledc esp_timer
   )
 
-  # EspHal::delay() uses vTaskDelay(pdMS_TO_TICKS(ms)); below 1 kHz tick rate
-  # small delays round to zero (delay(5) -> vTaskDelay(0)). arduino-esp32
-  # makes this a hard error; we warn so users can opt in to lower tick rates
-  # if they really know what they are doing.
-  if(DEFINED CONFIG_FREERTOS_HZ AND CONFIG_FREERTOS_HZ LESS 1000)
-    message(WARNING
-      "RadioLib EspHal: CONFIG_FREERTOS_HZ is ${CONFIG_FREERTOS_HZ}; "
-      "set it to 1000 for accurate delay() resolution.")
-  endif()
-
   return()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,12 +13,13 @@ if(ESP_PLATFORM)
   # required because ESP-IDF runs cmake in script mode
   # and needs idf_component_register()
 
-  # add the ESP-IDF HAL implementation, which was excluded by the filter above
+  # the ESP-IDF HAL was excluded from RADIOLIB_SOURCES by the filter above;
+  # pass it directly to idf_component_register so RADIOLIB_SOURCES stays
+  # HAL-free, the same way other build environments handle it
   file(GLOB RADIOLIB_ESP_IDF_HAL_SOURCES "src/hal/ESP-IDF/*.cpp")
-  list(APPEND RADIOLIB_SOURCES ${RADIOLIB_ESP_IDF_HAL_SOURCES})
 
   idf_component_register(
-    SRCS ${RADIOLIB_SOURCES}
+    SRCS ${RADIOLIB_SOURCES} ${RADIOLIB_ESP_IDF_HAL_SOURCES}
     INCLUDE_DIRS . src
     REQUIRES esp_driver_gpio esp_driver_spi esp_driver_ledc esp_timer
   )
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -243,10 +243,14 @@ jobs:
 
   esp-build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [esp32, esp32s3, esp32c3]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      
+
       - name: Install dependencies
         run: |
           sudo apt-get update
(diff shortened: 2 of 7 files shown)


bitandpin:
@jgromes @x42en @Steve-Tech @vshymanskyy

Steve-Tech:
LGTM!

Although, and this is a massive nitpick, when I asked @x42en to add [`ESP_PLATFORM` in `BuildOpt.h`](https://github.com/jgromes/RadioLib/pull/1793#discussion_r3354821238), I realised after the PR was merged that that section of code was guarded by [`RADIOLIB_BUILD_ARDUINO`](https://github.com/jgromes/RadioLib/blob/217f8cf4bd8f7bc803ba5e9f6db0235fa37f0f9b/src/BuildOpt.h#L152) anyway, so it did nothing for ESP-IDF. I was wondering if you were willing to add a new section similar to [`STM32WLE5xx`](https://github.com/jgromes/RadioLib/blob/217f8cf4bd8f7bc803ba5e9f6db0235fa37f0f9b/src/BuildOpt.h#L468-L488) for `ESP_PLATFORM`.

CMakeLists.txt
```diff
@@ -13,9 +13,14 @@ if(ESP_PLATFORM)
   # required because ESP-IDF runs cmake in script mode
   # and needs idf_component_register()
 
+  # add the ESP-IDF HAL implementation, which was excluded by the filter above
+  file(GLOB RADIOLIB_ESP_IDF_HAL_SOURCES "src/hal/ESP-IDF/*.cpp")
+  list(APPEND RADIOLIB_SOURCES ${RADIOLIB_ESP_IDF_HAL_SOURCES})
```
jgromes: I'm guessing that the `EspHal.cpp` file is needed to be passed to `idf_component_register` - if that is the case I would prefer to pass it some other way than re-adding it back to `RADIOLIB_SOURCES`, as other build envornments do not care about that file. Sure, it's compliation is prevented by a macro guard but having it back there would be a different approach than all the other HAL source files.

jgromes (commented):
@bitandpin thanks for revisiting this!

The changes are mostly cosmetic (I really don't care much about the format of the macro guard), though I did leave one comment regarding the CMake. The CI extension is a good addition - and obviously, before merging, cppcheck has to pass.

LoopAxel:
Posted a comment about delay not working as intented in #1793 moving it to here instead. Original: https://github.com/jgromes/RadioLib/pull/1793#issuecomment-4787439625

@Steve-Tech Brought up a valid consern, that ulTaskNotifyTake won't wake up a task until a new tick anyways. To be honest, I'm not sure. I guess one solution to that would be to call portYIELD_FROM_ISR(xHigherPriorityTaskWoken); after xTaskNotifyGive.

Anyway I like the idea @Steve-Tech proposed better:
> ```c++
>     void delay(unsigned long ms) override {
>         if (pdMS_TO_TICKS(ms) > 0) {
>             vTaskDelay(pdMS_TO_TICKS(ms));
>         } else {
>             this->delayMicroseconds(ms * 1000UL);
>         }
>     }
> ```

Though that exact implementation won't work when vTaskDelay is called between ticks (which is almost always will be) the delay will be of by the amount of time until the next tick.  
I changed it up abit and came up with this, did a quick test and for what I can see it works fine. Altough it can force buzy waiting upto 10ms, say what you want about that.
```c++
    void delay(unsigned long ms) override {
        uint64_t start_us = esp_timer_get_time();
        uint64_t target_us = start_us + (ms * 1000ULL);
    
        if (ms > portTICK_PERIOD_MS) {
            vTaskDelay(pdMS_TO_TICKS(ms) - 1);
        }
 
        int64_t remaining_us = (int64_t)target_us - (int64_t)esp_timer_get_time();
        if (remaining_us > 0) {
            esp_rom_delay_us((uint32_t)remaining_us);
        }
    }
```

jgromes:
@LoopAxel I'm not sure the timer is strictly needed there; you can wait for `(ms % 10) * 1000` us and be reasonably accurate I think. The remark about busy wait is true, but that will always be the case for waiting periods below the tick resolution.

Is the assumption of tick being always 10 ms accurate though? I am asking because ESP32 Arduino core seems to implement delay very simply: https://github.com/espressif/arduino-esp32/blob/6aa7d51a89039b4ae497415f85db08052018477e/cores/esp32/esp32-hal-misc.c#L212 I wasn't able to find out in that repo how fast the tick is, but it would seem strange for it to be 10 ms as that would cause the entire Arduino ESP32 to have this issue, which seems pretty strange for such a widespread Arduino core ...?

LoopAxel:
https://docs.espressif.com/projects/esp-idf/en/v4.3.2/esp32/api-reference/kconfig.html

> CONFIG_FREERTOS_HZ
> Tick rate (Hz)
> Select the tick rate at which FreeRTOS does pre-emptive context switching.
> Range:
> from 1 to 1000
> Default value:
> 100

https://github.com/espressif/arduino-esp32/blob/master/CMakeLists.txt

> if(NOT CONFIG_FREERTOS_HZ EQUAL 1000 AND NOT "$ENV{ARDUINO_SKIP_TICK_CHECK}")
>     # See delay() in cores/esp32/esp32-hal-misc.c.
>    message(FATAL_ERROR "esp32-arduino requires CONFIG_FREERTOS_HZ=1000 "
>                       "(currently ${CONFIG_FREERTOS_HZ})")
> endif()

@jgromes You are correct. Tick update rate is 1000Hz when using Arduino, which makes it work for Arduino. 
I just want to make it clear that I have no intrest in having my code in the repo, I'm only reporting this because it took me 10 hours yesterday to get this working :-) So the solution may be to add a documentation in the esp-idf example that CONFIG_FREERTOS_HZ needs to be set to 1000 for it to work. <- Which may be prefered anyway

jgromes:
@LoopAxel thanks for pointing us to that CMake snippet - I think it would be worth it to use that to let the user know via an error or warning that 1 kHz tickrate is vastly preferred. And then we don't have to mess with any weird combination of delays (which could break again if the tickrate changes).

bitandpin:
I will add the warning, but I dont agree it should be there. CONFIG_FREERTOS_HZ default value being 100 and its consequences (like min delay being 10ms) is well documented and widely known to the ESP developers. In ESP IDF v5.5.4 examples, there are 15 examples with CONFIG_FREERTOS_HZ set to 1000, and 22 examples with CONFIG_FREERTOS_HZ set to default 100.

Precise timings are simply not expected with default CONFIG_FREERTOS_HZ value. For those who actually care about precise timings, thats a firmware-wide decision, and all their firmware code will be developed in a manner that expects it.

I believe bringing up a build warning about SDK default value is an overkill. A comment on a delay function might be better.

StevenCellist:
In the LoRaWAN world, 10ms is the difference between always receiving a message and never receiving a message. Not receiving a message is OK if it happens once, bad luck if it happens twice, but catastrophic if it never happens. 
Given that radio comms more often than not rely on battery power and therefore are optimized down to milliseconds, I think this deserves a huge fat yellow warning.

bitandpin:
Also, I do agree we should pursue precise timings to the max possible extent, even under the framework limitations, so this https://github.com/jgromes/RadioLib/pull/1812#issuecomment-4789699789 is reasonable to add if it offers better precision than simple vTaskDelay(pdMS_TO_TICKS(ms)).

CMakeLists.txt
```diff
@@ -13,11 +13,27 @@ if(ESP_PLATFORM)
...
 
+  # EspHal::delay() uses vTaskDelay(pdMS_TO_TICKS(ms)); below 1 kHz tick rate
+  # small delays round to zero (delay(5) -> vTaskDelay(0)). arduino-esp32
+  # makes this a hard error; we warn so users can opt in to lower tick rates
+  # if they really know what they are doing.
+  if(DEFINED CONFIG_FREERTOS_HZ AND CONFIG_FREERTOS_HZ LESS 1000)
+    message(WARNING
+      "RadioLib EspHal: CONFIG_FREERTOS_HZ is ${CONFIG_FREERTOS_HZ}; "
+      "set it to 1000 for accurate delay() resolution.")
+  endif()
```
vshymanskyy: I think it can be (more naturally) represented as a preprocessor check in `EspHal.cpp`:
```c
#include "sdkconfig.h"

#if defined(CONFIG_FREERTOS_HZ) && (CONFIG_FREERTOS_HZ < 1000)
  #if !defined(RADIOLIB_RELAX_RTOS_TICK) || (RADIOLIB_RELAX_RTOS_TICK != 1)
    #error "RadioLib EspHal: CONFIG_FREERTOS_HZ is " RADIOLIB_VALUE(CONFIG_FREERTOS_HZ) "; set it to 1000 for accurate delay() resolution, or define RADIOLIB_RELAX_RTOS_TICK=1 to bypass this check."
  #endif
#endif
```

This would make it impossible to miss, and at the same time gives a way to unblock the build.

jgromes:
I went through the changes and it all looks good, so merged - thank you @bitandpin for the contribution! And of courese thanks to @vshymanskyy, @Steve-Tech and @LoopAxel for the reviews and suggestions.

PR #1813, opened 2026-06-23
[LR2021] Fix setRfSwitchTable to allow non-contiguous DIO pins

oltaco:
This one liner fixes a bug in LR2021::setRfSwitchTable that makes it impossible to use non-contiguous DIO pins in the table.

The RF switch mode bitmap is built using ``dioConfigs[dioNum]`` but it's applied using ``dioConfigs[i]``. For non-contiguous DIO assignments this causes the configuration to be read from the wrong dioConfigs[] entries, resulting in incorrect (zeroed) RF switch configuration being applied to the DIO pins which are after the gap.

For example if you want to use an RF switch table with DIO5,6,9,10,11 (skipping DIO7 and 8) then the config for DIO9,10,11 would not be applied correctly.

https://github.com/jgromes/RadioLib/blob/217f8cf4bd8f7bc803ba5e9f6db0235fa37f0f9b/src/modules/LR2021/LR2021_config.cpp#L211-L223

Tested and confimed it works.


jgromes:
Thank you for the contribution! The LR2021 design I am using is switchless, but I did confirm the logic is indeed fixed using debugger. No comments, merged!

oltaco:
Thanks for all the hard work on RadioLib 👍

PR #1818, opened 2026-06-28
[LoRaWAN] Add runtime (sub)band selection

### LINKED ISSUE #1770: Support runtime LoRaWAN region and sub-band selection
**Related problem**
Currently, the LoRaWAN region is selected when constructing the node:

```cpp
const LoRaWANBand_t Region = EU868;
const uint8_t subBand = 0;

LoRaWANNode node(&radio, &Region, subBand);
```

This works well when the region is fixed at compile time, which is probably the common case. However, some applications may need to select the region after compilation, for example from stored configuration or as a user setting.

The applicability is limited because many sensors/devices are only suitable for one region due to RF front-end, antenna, and certification constraints. Still, runtime switching should be feasible and useful for compatible regions such as `US915`, `AU915`, and `AS923`, and especially for selecting/changing their sub-bands (see [Australia](https://www.thethingsnetwork.org/forum/t/future-strategies-for-au915-and-as923-in-australia/44218?utm_source=chatgpt.com))

**Proposed solution**
It would be useful to have an officially supported way to recreate or reconfigure a `LoRaWANNode` with a different region and sub-band at runtime, for example:

```cpp
node.setRegion(&US915, 2);
```

or an officially documented pattern that safely invalidates the old LoRaWAN session and reinitializes the node.

A possible workaround today appears to be manually reconstructing the node in statically allocated memory:

```cpp
LoRaWANNode* node = nullptr;

void createNode(const LoRaWANBand_t* region, uint8_t subBand) {
  if(node != nullptr) {
    node->~LoRaWANNode();
  }

  static alignas(LoRaWANNode) uint8_t nodeStorage[sizeof(LoRaWANNode)];
  node = new (nodeStorage) LoRaWANNode(&radio, region, subBand);
}
```
Note that the destructor is not yet implemented. 
Then switching region would require clearing/invalidating the old session and joining again:

```cpp
createNode(&US915, 2);
node->beginOTAA(joinEUI, devEUI, nwkKey, appKey);
node->activateOTAA();
```

It would be helpful if RadioLib either supported this directly or documented which internal state must be reset when changing region.

**Alternatives considered**
The current alternative is to compile separate firmware builds per region, or to choose the region at compile time:

```cpp
const LoRaWANBand_t Region = AU915;
LoRaWANNode node(&radio, &Region, 2);
```

Another workaround is the placement-new reconstruction pattern shown above. However, this is not very discoverable and may be unsafe if users do not know exactly which LoRaWAN session, channel, MAC, and timing state must be invalidated.

**Additional context**
Some points that may need discussion before implementing this:

* All region configs that can be selected at runtime must remain in the compiled binary. Should this be the default, even though it may increase flash size, or should there be a way to define a list of enabled regions at compile time?
* How should existing OTAA/ABP session state be handled when the region changes?
* Should nonces and identity/root credentials be preserved while session/channel/MAC state is cleared?
* Should this be exposed as `setRegion(...)`, as a documented reconstruction helper, or left as an advanced user pattern?
* For `US915`/`AU915`, should changing only the sub-band be supported separately from changing the full region?

StevenCellist:
Closes #1770 (yes I know it is being merged into staging but it's not a necessary bug but instead an enhancement).

### DIFF
--- a/src/TypeDef.h
+++ b/src/TypeDef.h
@@ -613,6 +613,11 @@
 */
 #define RADIOLIB_ERR_INVALID_MULTICAST_GROUP                    (-1122)
 
+/*!
+  \brief The requested subband is invalid for this band (or there is no subband).
+*/
+#define RADIOLIB_ERR_INVALID_SUBBAND                            (-1123)
+
 // LR11x0-specific status codes
 
 /*!
--- a/src/protocols/LoRaWAN/LoRaWAN.cpp
+++ b/src/protocols/LoRaWAN/LoRaWAN.cpp
@@ -588,6 +588,11 @@ int16_t LoRaWANNode::loadSessionBuffer() {
     return(RADIOLIB_ERR_SESSION_DISCARDED);
   }
 
+  // check the band plan
+  if(buff[RADIOLIB_LORAWAN_SESSION_BAND] != this->band->bandNum) {
+    return(RADIOLIB_ERR_SESSION_DISCARDED);
+  }
+
   // copy the whole buffer over
   memcpy(this->bufferSession, buff, RADIOLIB_LORAWAN_SESSION_BUF_SIZE);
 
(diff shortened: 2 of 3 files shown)


jgromes:
Looks OK from my side - merged, thanks!

PR #1824, opened 2026-07-01
[LoRaWAN] Rework package handling, add TS007

StevenCellist:
This PR adds TS007 - which admittedly is not in use anywhere to my knowledge, but it means that the package manager is fully functional.

TS003 and TS009 are reworked to fully integrate with the package manager, instead of being a hybrid between their old standalone state and the later added package manager.

Disclaimer: some of the (doxygen) comments are autocompleted by an LLM.

One point for review: the package manager needs access to the HAL. It cannot access the HAL from PHY currently so I added it to the initializer, however this requires a manual instance of ArduinoHal on the Arduino platforms, where it instead is typically automatically constructed within the Module constructor. I am happy to force the user to instantiate the HAL and supply it to the Module too because any use of the Package Manager should be done by a decently experienced user, but it is a point for review nonetheless.

Example to be updated soon - I have the certification sketch ready but will review all examples anyway.

TS003, TS007 and TS009 all pass certification testing on an ESP32-S3 and an STM32WLE5 using LCTT.

### DIFF
--- a/src/protocols/LoRaWAN/LoRaWANPacMan.cpp
+++ b/src/protocols/LoRaWAN/LoRaWANPacMan.cpp
@@ -6,49 +6,46 @@
 #include <string.h>
 
 // LoRaWANPackage implementation
-LoRaWANPackage::LoRaWANPackage(uint8_t ts, LoRaWANPackageManager* pacMan, LoRaWANNode* node, GetSecondsCb_t secondsCb)
-  : packageIdentifier(ts), lenUp(0), pacMan(pacMan), lorawanNode(node), getSeconds_cb(secondsCb) {
+LoRaWANPackage::LoRaWANPackage(uint8_t ts, LoRaWANPackageManager* pacMan, RadioLibHal* hal, LoRaWANNode* node, GetSecondsCb_t secondsCb)
+  : packageIdentifier(ts), pacMan(pacMan), hal(hal), lorawanNode(node), getSeconds_cb(secondsCb) {
   this->packageVersion = 0;
-  memset(this->dataUp, 0, sizeof(this->dataUp));
 }
 
-size_t LoRaWANPackage::processData(const uint8_t* data, size_t len, LoRaWANEvent_t* event) {
-  // Default implementation: assume the provided buffer is entirely consumed.
-  // Derived classes should override and return actual bytes consumed.
-  (void)data;
+size_t LoRaWANPackage::processData(const uint8_t* dataIn, size_t lenIn, uint8_t* dataOut, size_t* lenOut, LoRaWANEvent_t* event) {
+  // Default implementation: consume nothing and produce no answer.
+  // Derived classes override and return actual bytes consumed.
+  (void)dataIn;
+  (void)lenIn;
+  (void)dataOut;
   (void)event;
-  return(len);
+  *lenOut = 0;
+  return(0);
 }
 
-void LoRaWANPackage::getUplinkData(uint8_t* dataOut, size_t* lenOut) {
-  if(this->lenUp > 0) {
-    memcpy(dataOut, this->dataUp, this->lenUp);
-    *lenOut = (uint8_t)this->lenUp;
-    this->lenUp = 0;
-  } else {
-    *lenOut = 0;
-  }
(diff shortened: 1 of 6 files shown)


src/protocols/LoRaWAN/LoRaWANPackageTS003.cpp
```diff
@@ -2,45 +2,45 @@
...
 void LoRaWANPackageTS003::setSecondsCb(SetSecondsCb_t cb) {
   this->setSeconds = cb;
 }
 
-LoRaWANTaskInfo LoRaWANPackageTS003::hasTask() {
-  LoRaWANTaskInfo task;
-  task.type = RADIOLIB_LORAWAN_TASK_NONE;
-  task.time = 0;
+bool LoRaWANPackageTS003::handleTask(RadioLibTime_t* tNext, bool* uplinkDue) {
+  *uplinkDue = false;
```
jgromes, line 16: I don't remember whether/which packages are accessible to the user, but a good practice here and elsewhere would be to check the pointers are set at least in all methods that are public in each class.

src/protocols/LoRaWAN/LoRaWANPackageTS009.cpp
```diff
@@ -190,42 +208,47 @@ size_t LoRaWANPackageTS009::processData(const uint8_t* dataDown, size_t lenDown,
...
 
-        this->radioModule->setFrequency(freq);
-        this->radioModule->setOutputPower(txPower);
-        this->radioModule->transmitDirect();
+        this->radio->setFrequency(freq);
+        this->radio->setOutputPower(txPower);
+        this->radio->transmitDirect();
         this->delaySecondsCallback(timeout);
-        this->radioModule->standby();
+        this->radio->standby();
```
jgromes, line 215: If we assume these calls cannot fail then it would be better to explicitly cast to void, e.g. `(void)this->radio->setFrequency(freq);`. But passing at least errors to debug might be a better idea.

src/protocols/LoRaWAN/LoRaWANPacMan.cpp
```diff
@@ -6,49 +6,46 @@
...
-  // default: no action
-}
+// #if defined(RADIOLIB_BUILD_ARDUINO)
+// // LoRaWANPackageManager implementation
+// LoRaWANPackageManager::LoRaWANPackageManager(LoRaWANNode* node, GetSecondsCb_t secondsCb) {
+//   RadioLibHal* hal = new ArduinoHal();
+
+//   return(LoRaWANPackageManager(hal, node, secondsCb));
+// }
+// #endif
```
jgromes, line 45: Commented out code is ugly, please remove ;)

jgromes (commented):
I did a light review and left some comments, but truth be told I can't really review most of the changes in pacman since that's just way over my head at this point.

src/protocols/LoRaWAN/LoRaWANPacMan.cpp
```diff
@@ -57,8 +54,27 @@ LoRaWANPackageManager::LoRaWANPackageManager(LoRaWANNode* node, GetSecondsCb_t s
...
 
   // store node and callback reference
+  this->hal = hal;
   this->lorawanNode = node;
   this->getSecondsCb = secondsCb;
+
+  // if the user does not provide their own AES-128, use the software one
+  #if !RADIOLIB_CUSTOM_AES128
+  static RadioLibSoftwareAES128 RadioLibAES128Instance;
+  this->hal->aes128 = &RadioLibAES128Instance;
```
jgromes, line 64: > this requires a manual instance of ArduinoHal on the Arduino platforms, where it instead is typically automatically constructed within the Module constructor

Why does this need to set up the AES for the HAL? I don't see it being used anywhere in pacman and for LoRaWAN the Module-provided HAL is used anyway.

StevenCellist:
@jgromes I addressed the feedback, but there is one float in here. I can prematurely convert it into the new integer format since the code is rarely used anyway (not even by me), or shall we catch it after merging the different PRs?

jgromes:
@StevenCellist if it is not blocked by the other stuff being merged first then let's do it right now so that we don't forget about it.

StevenCellist:
Done and ready.

jgromes:
Looking good - merged, thanks!

PR #1826, opened 2026-07-01
Fix building STM32 in PIO

### LINKED ISSUE #1825: STM32WL regression: SubGhz.h: No such file or directory on PlatformIO (RAK3172)
Building RadioLib HEAD (847b01c) for STM32WL under PlatformIO fails again with fatal error: SubGhz.h: No such file or directory when compiling hal/Stm32duino/Stm32wlHal.cpp as in #718 

The Fix was moving #include <SubGhz.h> into Module.h so LDF would pick it up. The recent refactor and separate HAL moved it back out into leaf files behind #if defined(ARDUINO_ARCH_STM32) && defined(STM32WLxx)
Restoring the SubGhz.h include in Module.h fixes it.

caveman99:
fix #1825

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -9,6 +9,11 @@
   #include <SPI.h>
 #endif
 
+// Keep this SubGhz include here to fix Platformio LDO (#718).
+#if defined(STM32WLxx)
+  #include <SubGhz.h>
+#endif
+
 /*!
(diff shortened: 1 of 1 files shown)


jgromes:
Looks OK, thanks for the fix. It might be worth it to figure out Platformio builds in the CI t ocatch these kinds of bugs.

PR #1828, opened 2026-07-03
[ESP-IDF] Don't build EspHal for Arduino targets

caveman99:
EspHal only compiles clean with ESP-IDF above 5.2 - the older arduino versons in platformio lack the neccessary headers. Sadly Arduino also defines ESP_PLATFORM since it is built on ESP-IDF, so those builds fail to compile even though EspHal is never used there.

Gate the HAL on `!defined(ARDUINO)` so it only builds for pure ESP-IDF, mirroring the RADIOLIB_BUILD_ARDUINO guard in ArduinoHal.cpp. The header guard runs before RadioLib.h is included, so it tests the predefined ARDUINO macro directly rather than RADIOLIB_BUILD_ARDUINO.

This is a quick shot fix that may or may not require additional changes. 

Example: https://github.com/meshtastic/firmware/actions/runs/28668405797/job/85033145387?pr=10854

Arduino builds should not even try to compile a different HAL.

### DIFF
--- a/src/hal/ESP-IDF/EspHal.cpp
+++ b/src/hal/ESP-IDF/EspHal.cpp
@@ -3,7 +3,12 @@
 
 #include "EspHal.h"
 
-#if defined(ESP_PLATFORM)
+// Only build the native ESP-IDF HAL for pure ESP-IDF targets. Under Arduino
+// (including Arduino-ESP32, which also defines ESP_PLATFORM) RadioLib uses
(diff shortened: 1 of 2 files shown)


jgromes:
Thanks for the fix! Since we only build Arduino Mega during non-release CI builds, this one slipped by in the recent ESP-IDF reworks.

caveman99:
No sweat, the ESP-IDF bindings work like a charme by the way, doing some testing with them.

Issue #1830, opened 2026-07-07
Scanchannel / CAD hang


thebentern:
**Describe the bug**

On a Semtech LR2021, `scanChannel()` never returns on an idle channel - it hangs forever. `startChannelScan()` succeeds and the CAD runs, but the blocking wait `while(!this->mod->hal->digitalRead(this->mod->getIrq()))` never sees the IRQ pin assert.

Root cause (checked on hardware): the **CadDone interrupt latches in the IRQ status register (~19 ms after the scan starts) but is never asserted on the DIO pin** when the CAD exits back to standby (the `CAD_ONLY` exit mode `scanChannel()` uses). The status register clearly shows CadDone (bit 20), the DIO IRQ mask is correct (`0x00100080` -> DIO5), and DIO5 is configured as IRQ - yet the pin never toggles, not even a 1 us pulse (microsecond-resolution poll). Rx/Tx work only because they attach a hardware edge interrupt to a pin that does assert; CAD's software poll of the pin can never complete.

For context, this looks like a residual of #1727 (merged, present in 7.7.1). That PR fixed the CAD *command* path (SetLoraCadParams / SetLoraCad), so CadDone is now generated correctly - but `scanChannel()` still hangs, because its wait loop polls the DIO pin rather than the status register, and CAD completion never reaches the pin.

Fix that works: poll the IRQ **status register** in `scanChannel()` (which `getChannelScanResult()` already reads) instead of the DIO pin, bounded by a timeout derived from the CAD duration. `scanChannel()` then returns `RADIOLIB_CHANNEL_FREE` in ~19 ms.

Also, unrelated but spotted while here: the CAD exit-mode constants don't match datasheet Table 6-18 - `CAD_RX` should be `0x01` and `CAD_LBT` `0x10`, but RadioLib has `CAD_EXIT_MODE_TX = 0x01` / `CAD_EXIT_MODE_RX = 0x02`; `0x02` returns a command error on the chip. The default `0x00` (CAD_ONLY) path is unaffected.

A two-commit fix is ready and hardware-verified if a PR is welcome.

<details><summary>Debug mode output</summary>
<p>

Not standard RadioLib debug output - this is my instrumented probe reading the raw IRQ status register versus the DIO pin during CAD:

```
[startScan] irqFlags=0x00000180 mapped=0x00100080 dio=5   ; CAD_DONE|CAD_DETECTED correctly mapped to DIO5
[...]

PR #1831, opened 2026-07-07
Lr2021 cad scanchannel fix

thebentern:
Fixes an LR2021 `scanChannel()` that hangs forever on an idle channel, plus a related CAD exit-mode constant mismatch. Full root-cause writeup and hardware evidence are in #1830 .

**Goal**

Two commits:

1. **`scanChannel()` polls the IRQ status register instead of the DIO pin.** On the LR2021 the CadDone interrupt latches in the IRQ status register but is never asserted on the DIO pin when the CAD exits to standby (the CAD_ONLY mode `scanChannel()` uses). The wait loop polled the pin, so it never returned. It now polls the status register (which `getChannelScanResult()` already reads), bounded by a timeout derived from the CAD duration as a safety net. This is the remaining half of the freeze #1727 addressed: that PR fixed the CAD command so CadDone fires; this fixes the wait so `scanChannel()` actually returns.

2. **Corrects the CAD exit-mode constants** to match datasheet Table 6-18 (`CAD_RX = 0x01`, `CAD_LBT/TX = 0x10`). RadioLib had `TX = 0x01` / `RX = 0x02`; `0x02` is not a valid exit mode and the chip returns a command error for it (observed on hardware).

**Why upstream**

LR2021 CAD / `scanChannel()` is unusable without this - any idle-channel scan hangs the caller indefinitely. The constant fix stops anyone using the (currently wrong) TX/RX exit modes from writing an invalid value.

**Foreseen impacts**

- `LR2021::scanChannel()` behavior changes from "block forever" to "return in ~19 ms" (or `RADIOLIB_ERR_RX_TIMEOUT` if a scan never completes). No API/signature change; no other module is touched.
- The exit-mode constant values change, but RadioLib only uses `FALLBACK` (0x00) internally, so the default CAD path is unaffected. This only matters for code that references `CAD_EXIT_MODE_TX` / `CAD_EXIT_MODE_RX` directly.

**Testing**

Hardware-tested on a Semtech LR2021 (ESP32-S3): `scanChannel()` previously hung; it now returns `RADIOLIB_CHANNEL_FREE` in ~19 ms, repeatably. The idle / CHANNEL_FREE path is verified end-to-end; the DETECTED path uses the same status-register read but I could not stage live traffic on my bench to exercise it. I have not run the Raspberry Pi runtime CI locally - the change is LR2021-specific and should not affect other platforms.

Written with AI assistance.

### DIFF
--- a/src/modules/LR2021/LR2021.cpp
+++ b/src/modules/LR2021/LR2021.cpp
@@ -421,11 +421,21 @@ int16_t LR2021::scanChannel() {
 
 int16_t LR2021::scanChannel(const ChannelScanConfig_t &cfg) {
   // set mode to CAD
-  int state = startChannelScan(cfg);
+  int16_t state = startChannelScan(cfg);
   RADIOLIB_ASSERT(state);
 
-  // wait for channel activity detected or timeout
-  while(!this->mod->hal->digitalRead(this->mod->getIrq())) {
+  // Wait for the CAD to finish. The LR2021 latches the CadDone (and CadDetected) interrupt in the IRQ status
+  // register, but does not assert it on the DIO pin when the CAD exits back to standby (the CAD_ONLY exit
+  // mode used here). Polling the DIO pin (as Rx/Tx do) therefore never completes and hangs forever, so poll
+  // the IRQ status register instead, bounded by a timeout derived from the CAD duration as a safety net.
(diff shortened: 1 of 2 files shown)


thebentern:
Closing. Issue was determined to be a pin mistake.

PR #1833, opened 2026-07-13
Fix AX25Frame::operator= writing through unallocated buffers

94xhn:
## What

`AX25Frame::operator=` copies into `this->info`, `this->repeaterCallsigns` and `this->repeaterSSIDs` without ever (re)allocating them first. In the default dynamic-memory build (`RADIOLIB_STATIC_ONLY == 0`), these are raw pointer members that constructors only allocate when `infoLen`/`numRepeaters` is non-zero. A frame built via the info-less/repeater-less constructor overload therefore has these pointers indeterminate, and assigning a frame that *does* have an info field or repeaters onto it writes through that indeterminate pointer:

```cpp
AX25Frame dst("DEST", 0, "SRC", 0, 0x03);           // no info, no repeaters
AX25Frame src("DEST", 0, "SRC", 0, 0x03, 0xF0, payload, 8);
dst = src;                                          // segfault (dst.info never allocated)
```

Reproduced on a clean host build (real repo sources, no mocks): the above sequence reliably segfaults before this change.

## Fix

Mirror the allocation pattern already used by the copy constructor, and the freeing pattern already used by the destructor: free any buffer `this` currently owns, then allocate fresh buffers sized to match the source frame before copying. Added a self-assignment guard, which becomes required once the operator frees its own buffers before copying (otherwise self-assignment would free the very buffer it's about to read from).

Verified locally (built the actual library target via CMake, MinGW g++): the crash repro above no longer crashes and copies correctly; also checked the repeaters variant, self-assignment, shrink/grow reassignment, and a 200k-iteration reassignment loop for leaks (working-set growth ~0.1 bytes/iteration, not the ~512 bytes/iteration a real leak would show).

## Disclosure

Per this repo's CONTRIBUTING.md AI-use policy: this fix was produced with AI assistance (Claude), with the bug independently verified by compiling and running the reproduction above against current `master`, and the fix independently compiled and tested before submission.

### DIFF
--- a/src/protocols/AX25/AX25.cpp
+++ b/src/protocols/AX25/AX25.cpp
@@ -108,6 +108,12 @@ AX25Frame::~AX25Frame() {
 }
 
 AX25Frame& AX25Frame::operator=(const AX25Frame& frame) {
+  // guard against self-assignment - required since the buffers owned by
+  // "this" are freed below before being (re)populated from "frame"
+  if(this == &frame) {
+    return(*this);
+  }
+
   // destination callsign/SSID
   memcpy(this->destCallsign, frame.destCallsign, strlen(frame.destCallsign));
   this->destCallsign[strlen(frame.destCallsign)] = '\0';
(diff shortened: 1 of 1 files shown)


jgromes:
Thank you for the contribution, this is indeed an oversight on my part. However, to make it fully safe, I think we cannot rely on `this->numRepeaters` and `this->infoLen` to decide whether the dynamic arrays can be deleted or not. Since they are all public member variables, the library user may have modified them in some way so that `this->numRepeaters > 0`, however, `this->repeaterCallsigns == NULL` or `this->repeaterCallsigns[i] == NULL`, at which point we would be calling delete on a null pointer.

So we should probably always check whether the opinter actually points to something before deleting.

94xhn:
Thanks, updated in cb1c89d. Cleanup now checks the actual info, repeaterCallsigns, repeaterSSIDs, and individual callsign pointers before releasing them, rather than relying only on the public length/count fields. I also initialize the dynamic pointers to NULL in both construction paths and use the same pointer-based cleanup in the destructor.\n\nVerified with the full MinGW CMake library build, a focused assignment harness covering self-assignment, repeater copying, and nonzero public counts/lengths with NULL pointers, plus a WSL GCC 13 ASan/UBSan leak-enabled run. The static-only AX25 source also compiles cleanly with -Werror.

jgromes:
Thank you for the addition - no additional comments, so merged!

94xhn:
Thank you for reviewing and merging this fix. I appreciate it.

PR #1835, opened 2026-07-16
[MOD] Fix setRfSwitchTable Doxygen to match 5-pin array size

mayl:
The `setRfSwitchTable` doc comment in `Module.h` still says the `pins` argument "should always be an array of 3 elements" and its inline example uses a 3-element initializer — leftovers from before f78b3ccc9 raised `RFSWITCH_MAX_PINS` to 5. Since the parameter is `const uint32_t (&)[RFSWITCH_MAX_PINS]`, the documented example doesn't compile.

This updates the text and the example to the 5-element form used by the STM32WLx examples. Documentation only, no code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Module.h
+++ b/src/Module.h
@@ -435,7 +435,7 @@ class Module {
 
 
       \param pins A reference to an array of pins to control. This
-      should always be an array of 3 elements. If you need less pins,
+      should always be an array of 5 elements. If you need less pins,
       use RADIOLIB_NC for the unused elements.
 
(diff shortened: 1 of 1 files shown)


jgromes:
A minor oversight during an update (not sure how is that worth burning tokens on), but thanks for fixing it.

mayl:
Ya it was not the focus of my session, I was trying to understand something else and I got confused by this so just opened a PR while I was in there

PR #1850, opened 2026-08-11
[LR2021] fix CR constants from the datasheet, add missing header mode check

caveman99:
Per LR2021 datasheet Rev 1.1 Table 9-4, cr 0x07 is long interleaver
coding rate 4/8, not 4/7. Long interleaved 4/7 does not exist, which
setCodingRate() already reflects by mapping 4/8 long interleaved onto
0x07, and LR11x0 names the same value RADIOLIB_LR11X0_LORA_CR_4_8_LONG.

The same table defines 0x08 and 0x09 as long interleaver convolutional
4/6 and 4/8. These are LR2021 only, incompatible with SX126x, SX127x,
SX128x and LR11x0, and were missing.

Also add the explicit header mode check that LR11x0, SX126x, SX127x and
SX128x already perform before reporting received header information.

### DIFF
--- a/src/modules/LR2021/LR2021.cpp
+++ b/src/modules/LR2021/LR2021.cpp
@@ -1190,6 +1190,11 @@ uint8_t LR2021::randomByte() {
 }
 
 int16_t LR2021::getLoRaRxHeaderInfo(uint8_t* cr, bool* hasCRC){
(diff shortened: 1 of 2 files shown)


jgromes:
Looking good - merged, thanks! I probably just misread the datasheet at some point, and those long interleaver CRs are probably not used very often anyway, hence why this escaped.

PR #1851, opened 2026-08-11
[LR11xx] fix actual coding rate return

caveman99:
Per LR1110 User Manual Rev 2.3 Table 7-18, the GetLoRaRxHeaderInfos response byte is RFU(7:5), CRC(4), RFU(3), coding_rate(2:0). Reading (buff[0] & 0x70) >> 4 skipped the coding rate entirely and returned the CRC flag in its place, so getLoRaRxHeaderInfo() reported coding rate 1 for every packet received with CRC enabled, and 0 for every packet without.


jgromes:
Unsure how this happened, but you're right, that was indeed incorrect as shown also in the LR1110 datasheet. Merged, thanks for the fix!

caveman99:
It went unnoticed in Meshtastic since it only affects airtime calculation of Rx packets and by chance the returned value for 'CRC on' is the same we expected for the default LongFast CR of 4/8. It only surfaces while hunting for a genuine problem in our calculations for LR2021 and comparing the returns to LR1110

PR #1852, opened 2026-08-12
[CC1101] Sync word repeat and TXOFF_RX modes

Mick605:
Hello,

As discussed [here](https://github.com/jgromes/RadioLib/discussions/1846), this pull requests add 2 features for CC1101:
 - Enable the use of repeated sync word of 32 bits (SYNC_MODE_30_32)
 - Enable the use of TXOFF_RX mode
 
 I have tested it in my use case, it works well. I also tried sending message of 200 bytes, to test the case where the data > 64 bytes, and it seems to work properly.

Regarding the concerns about the `transmit()` function that was broken after manually calling `radio.SPIsetRegValue(RADIOLIB_CC1101_REG_MCSM1, RADIOLIB_CC1101_TXOFF_RX, 1, 0);`, :

The function was returning a timeout because of the finishTransmit function which was expecting MARCSTATE to go back to IDLE after transmission. I changed it to either expect IDLE or RX based on the state of MCSM1. 

As discussed, I implemented `enableRxAfterTx` as a separate function to minimize API impacts. Another possible implementation would have been to add an argument `bool switchToRx = false` to `transmit`, `startTransmit` and `finishTransmit`, and enable/disable the TXOFF mode inside these functions...

Please let me know if everything is correct, or if you want to change something.
Thank you

### DIFF
--- a/src/modules/CC1101/CC1101.cpp
+++ b/src/modules/CC1101/CC1101.cpp
@@ -354,18 +354,25 @@ int16_t CC1101::startTransmit(const uint8_t* data, size_t len, uint8_t addr) {
 int16_t CC1101::finishTransmit() {
   // set mode to standby to disable transmitter/RF switch
   
-  // Check MARCSTATE for Idle to let anything in the FIFO empty
+  // expected MARCSTATE should be 0x01 (IDLE) except if TXOFF_RX is enabled
+  byte expectedState = (SPIgetRegValue(RADIOLIB_CC1101_REG_MCSM1, 1, 0) == RADIOLIB_CC1101_TXOFF_RX ? 0x0D : 0x01);
+
+  // Check MARCSTATE is in expected state to let anything in the FIFO empty
   // Timeout is 2x FIFO transmit time
   RadioLibTime_t timeout = (1.0f/(this->bitRate))*(RADIOLIB_CC1101_FIFO_SIZE*2.0f);
   RadioLibTime_t start = this->mod->hal->millis();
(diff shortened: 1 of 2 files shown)


src/modules/CC1101/CC1101.cpp
```diff
@@ -354,18 +354,25 @@ int16_t CC1101::startTransmit(const uint8_t* data, size_t len, uint8_t addr) {
 int16_t CC1101::finishTransmit() {
   // set mode to standby to disable transmitter/RF switch
   
-  // Check MARCSTATE for Idle to let anything in the FIFO empty
+  // expected MARCSTATE should be 0x01 (IDLE) except if TXOFF_RX is enabled
+  byte expectedState = (SPIgetRegValue(RADIOLIB_CC1101_REG_MCSM1, 1, 0) == RADIOLIB_CC1101_TXOFF_RX ? 0x0D : 0x01);
```
jgromes: `byte` is a type defined in Arduino, so it cannot be used on non-Arduino platforms, that is why the CI is failing. Please use fixed-width `stdint.h` types, `uint8_t` in this case.

jgromes (changes requested):
Thank you for the contribution! I left one comment that needs to be addressed to fix the CI and complication on non-Arduino platforms in general. After that, this is good to go.

Mick605:
Thanks for the feedback and the explanation. It should be fixed now

jgromes:
All good now - merged, thanks!

PR #1859, opened 2026-08-22
[PHY] Output power offset

### LINKED ISSUE #1840: Add gain / PA offset for transmission power
Heltec is rolling out v4 boards with an external PA. These boost output by something like 12dB or so. And then of course there are antennas with a certain gain, positive or negative.
For LoRa(WAN), it is essential to follow the requested (and mandated) power output values. I would therefore like to request some setter that adds (/subtracts) a flat Tx power offset at the radio level. Then when for instance a US915 device wants to use 24dB, it calls `setOutputPower()` with 26dB and in the case of the 12dB example above the radio would be configured for just 14dB instead of clipping or setting a value that is then also boosted even more. `checkOutputPower()` should also account for this offset of course.

jgromes:
This PR adds support for configuring output power offsets - mostly useful for devices with external power amplifiers. It supports both linear configuration (with a single offset value), as well as configuration via LUT.

Draft because it is currently completely untested, and so far only implements SX126x and LR2021.

Closes #1840


src/protocols/PhysicalLayer/PhysicalLayer.cpp
```diff
@@ -638,3 +638,27 @@ int16_t PhysicalLayer::calculateRxDutyCycle(size_t txPreLen, size_t rxPreLen, ui
...
+  return(RADIOLIB_ERR_NONE);
+}
+
+int16_t PhysicalLayer::applyOutputPowerOffset(int8_t base, const int8_t* pwrIn, int8_t* pwrOut) {
+  RADIOLIB_ASSERT_PTR(pwrIn);
+  RADIOLIB_ASSERT_PTR(pwrOut);
+  if(base - *pwrIn >= sizeof(this->paOffsetLut)) {
+    return(RADIOLIB_ERR_INVALID_OUTPUT_POWER);
+  }
+  *pwrOut = this->paOffsetLut[base - *pwrIn] + *pwrIn;
```
StevenCellist: This index is the wrong way round: it should subtract `base` instead.

Moreover, `setOutpowerPowerOffset(12)` should mean that there is a 12dB amplification in the line (as per the doxygen comment), so the `paOffsetLut` should be subtracted instead.

This lands me at the following (variable not necessary):
```cpp
  const int idx = (int)*pwrIn - (int)base;
  if((idx < 0) || ((size_t)idx >= sizeof(this->paOffsetLut))) {
    return(RADIOLIB_ERR_INVALID_OUTPUT_POWER);
  }
  *pwrOut = *pwrIn - this->paOffsetLut[idx];
```

StevenCellist (commented):
Thanks for drafting this, there is a logic bug though. 
Also, the `checkOutpowerPower` function will also need to account for this PA offset. Should we just add it like in `setOutputPower()` or should we try and be a bit smarter about this double code? I'd like to leave the radio alone to do the required power check for LoRaWAN, but it's becoming a bit of a hassle with these additions.

src/protocols/PhysicalLayer/PhysicalLayer.cpp
jgromes: Oops, right you are - fixed and checked the correct offset gets aplpied using GDB,

StevenCellist:
`readio` typo twice ;)

jgromes:
> readio typo twice ;)

Ugh, I should learn to not push when too tired :D

> Also, the checkOutpowerPower function will also need to account for this PA offset. Should we just add it like in setOutputPower()

We could just move the call to `applyOutputPowerOffset` beacuse it is being called from `setOutputPower` anyway, and use its `clipped` argument to pass the adjusted value. Admittedly it will no longer be only "clipping" but might also be increasing the value.

StevenCellist:
OK so probably an idea would be to remove this whole `checkOutputPower` stuff and expose the radio's min/max Tx power through PhysicalLayer. Then the LoRaWAN stack can just MIN/MAX the requested value with the radio's capabilities without requiring a whole bunch of modifications. And then we can add the PA offset to the returned value for min/max Tx power. Should work for me I think, how about you?

jgromes:
I kinda like the simplification because `checkOutputPower` is the odd one out, we don't have dedicated validator methods for other parameters. But:

1. Breaking change so this would have to be merged into `8.0.0-staging` (could be maybe postponed by leaving the current `checkOutputPower` unchagned but ... eww).
2. I will still need to keep the logic to do the check in `setOutputPower` since the inputs from user is still untrusted.

Most likely neither is a real problem but pointing it out.

StevenCellist:
Yeah the range checking should just move back to `setOutputPower` where it was before I mangled your work. And I guess we should push for 8.0.0 anyway. I can probably crunch my two PRs by next week, how far is your ~sinking~ defloating project?

jgromes:
OK, we can do it like that. I tried a couple rebases to get these comits onto `8.0.0-staging` butr probably messed up at some point, so it will be easier to start with a fresh branch and cherry-pick the needed commits.

PR #1864, opened 2026-09-06
Feature: Add DCDC mode for LR2021

carlhodder:
This PR adds methods swtich between use of LDO/DCDC (AKA SIMO) mode for the LR2021, it ports the workaround code mentioned in the datasheet:
SIMO-datasheet

Retrieved from: [Lora-net](https://github.com/Lora-net/usp/blob/main/smtc_rac_lib/radio_drivers/lr20xx_driver/src/lr20xx_workarounds.c)

**Changes**

The two new functions mostly map:

- `setDCDCworkaround `is `lr20xx_workarounds_dcdc_configure`
- `resetDCDCworkaround `is `lr20xx_workarounds_dcdc_reset`

I replaced the two register reads for frequency and is-high-freq with local vars and effectively inlined the 'lr20xx_workaround_dcdc_set_frequency' function.

I have this running for on 3 LR2021's (LF), and the current consumption drops ~4ma for me. Back to back programming and checking the noise floor suggests you get a bit less than a dB lower on the noise floor with this workaround than without.

I will do some more testing around this to make sure, I figured it's probably a good time for a review though.

The fix is applied for these functions (mentioned in datasheet, and code):

- `setGfskModulationParams`
- `setFlrcModulationParams`
- `setLoRaModulationParams`
- `setOokModulationParams`
- `setRxPath`

And also the undocumented reset called in that [code base](https://github.com/Lora-net/usp/blob/main/smtc_rac_lib/radio_drivers/lr20xx_driver/src/lr20xx_radio_common.c) on method:

- `setPacketType`

One semi-related change to highlight:

**setRegMode**
I could not find info in the datasheet for a variant that takes ramp times as a parameter, so I've replaced the existing (unused) implementation. I can revert if you have more knowledge about this?

Quite happy to make changes or answer questions :)


StevenCellist:
It's up to the boss to comment on the PR, I'll just add a comment on the code/style: as we're trying to de-float the library, can you precalculate the float multiplications please? Probably add the original calculation as a comment to make the resulting magic number a little less magic as it can then be cross-referenced with Semtech's code. Thanks!

carlhodder:
Oof, I obviously missed something trying to rush before bed. Sorry guys.

I'll be over this with fresh eyes tomorrow.

@StevenCellist thank you! I'll tweak that tomorrow. Although I'm not sure what light I can shed on those numbers - the numbers are inline without comments in the source too. But I can add links, and I can try emailing semtech.

StevenCellist:
Oh we're familiar with Semtech's quirks, don't worry about that. Just something like this (not the actual number):

```cpp
... = 450581950; // 430000000 * 1.047865, magic number from .....
```

src/modules/LR2021/LR2021.h
```diff
@@ -1057,7 +1070,10 @@ class LR2021: public LRxxxx {
 
     // test commands
     int16_t setTxTestMode(uint8_t mode);
-};
```
jgromes, line 1060: @carlhodder if you're wondering about the CI failures, this is the reason ;)

carlhodder: Thank you, made this easy to spot - I discovered I checked the wrong clone, I still had unstaged changes. Doh.

carlhodder:
If you want me to remove the comment to the repo, or switch it with this PR #, etc. more than happy to. 

I'll also test the HF mode in the next day or two, just to be sure.

src/modules/LR2021/LR2021_cmds_chip_control.cpp
```diff
@@ -341,4 +337,65 @@ int16_t LR2021::getPramVersion(uint16_t* version) {
...
+// workaround: port of semtech's code altered to use local freqMHz/highFreq - magic numbers are theirs
+// https://github.com/Lora-net/usp/blob/351b20153506/smtc_rac_lib/radio_drivers/lr20xx_driver/src/lr20xx_workarounds.c
+int16_t LR2021::setDCDCworkaround() {
+  uint32_t adcCtrlRaw = 0;
+  int16_t state = this->readRegMem32(RADIOLIB_LR2021_REG_DCDC_ADC_CTRL, &adcCtrlRaw, sizeof(adcCtrlRaw));
+  RADIOLIB_ASSERT(state);
+  const uint32_t anaDec = ( adcCtrlRaw >> 8 ) & 0x7;
+
+  if (!this->highFreq && (anaDec == 1 || anaDec == 2)) {
+    state = this->writeRegMemMask32(RADIOLIB_LR2021_REG_DCDC_SWITCHER, 0xF << 20, 11 << 20);
```
jgromes, line 357: The constants here need to be forced to a sufficiently large type (e.g., `0xFUL`) to avoid this getting truncated on platforms with 16-bit `int`

src/modules/LR2021/LR2021_cmds_chip_control.cpp
```diff
@@ -341,4 +337,65 @@ int16_t LR2021::getPramVersion(uint16_t* version) {
...
+  } 
+  state = this->writeRegMem32(RADIOLIB_LR2021_REG_DCDC_FREQ_LF, &freq_lf, sizeof(freq_lf));
+  RADIOLIB_ASSERT(state);
+  
+  state = this->setFrequency(this->freqMHz, true);
+  return(state);
+}
+
+int16_t LR2021::resetDCDCworkaround() {
+  if(this->freqMHz) {
```
jgromes, line 382: I'm not completely sure this is correct - the referenced [Semtech driver](https://github.com/Lora-net/usp/blob/351b2015350670eb4dfa3aec35eb04433e062654/smtc_rac_lib/radio_drivers/lr20xx_driver/src/lr20xx_workarounds.c#L510) does not decide based on the frequency, it always writes these.

jgromes (commented):
Thank you for the contribution @carlhodder, I reviewed it and left some comments. Overall nothing major.

jgromes:
Also one small note:

> setRegMode
I could not find info in the datasheet for a variant that takes ramp times as a parameter, so I've replaced the existing (unused) implementation. I can revert if you have more knowledge about this?

This was the case in a previous (preview) version of the datasheet. It has been removed from subsequent ones so probably this is no longer support by the commercially available devices.

src/modules/LR2021/LR2021_cmds_chip_control.cpp
carlhodder: Good call. I've pulled the check so it only guards setFrequency - the first call to modSetup calls setPacketType (via config) before setFrequency.

I feel this check/bypass is acceptable rather than refactoring anything - setFrequency is the following method called, and we still need a subsequent call to setRxPath and/or setXxxPacketParams before transitioning to RX or TX mode. Small side effect - a second call to begin() will have the workaround setFrequency to the current operating freq before it's called again with new freq.

src/modules/LR2021/LR2021_cmds_chip_control.cpp
carlhodder: I've updated those calls to specify UL type for these parameters +  a personal tweak - altered 0xF to 0x0F as I find it a little clearer when speed-reading to have the full byte.

carlhodder:
A bit late, whew has it been a week! I had suspected the ramp times were provisional, good to have that confirmed thank you. I've also set my test pair up on 2.4GHz and confirmed they talk to each other - only found one 2.4GHz antenna so had to make a sleeve dipole first haha.

Misc note:
I couldn't find anything indicating that setRegMode needs/should be called before the two calibrations for the LR2021 or other variants. I did a couple of back-to-back tests adding a manual calibration after and it didn't seem to have an effect.

jgromes:
@carlhodder thank you, everything loks good now - thanks for the contribution!

PR #1866, opened 2026-09-10
[PHY] Output power offset

### LINKED ISSUE #1840: Add gain / PA offset for transmission power

jgromes:
Continuation from https://github.com/jgromes/RadioLib/pull/1859

Closes #1840


jgromes:
@StevenCellist updated to use the min/max power vvia PHY approach, let me know if this works for you. If so, I will do the rest of the modules.

Just as a reminder, LR2021 has different power range based on the frequency band. Since presumably a different PA will be used for the two RF ports, I think the usage there will be a bit more complicated:

```c++
radio.setOutputPowerOffset(10); // +10 dB PA on the high-frequency RF port
radio.begin(); // start at sub-GHz band
(...)
radio.setFrequency(2400); // go to 2.4 GHz
radio.setOutputPowerOffset(5); // +5 dB PA on the high-frequency RF port
radio.setOutputPower(10);
```

StevenCellist:
@jgromes now that is just straight up an improvement, I now pushed the matching LoRaWAN update.

PR #1869, opened 2026-09-12
fix fabs() parentheses in setRxbandwidth

reidprichard:
As written, it was setting RX bandwidth (register `MDMCFG4`) to the first `point` value >= `rxBw * 1000 + 1000`:
* `fabs((rxBw * 1000.0f - point) <= 1000.0f)`

 It seems the intent is to set it to the first `point` value *within 1kHz of* the desired bandwidth (otherwise why bother with `fabs()`?):
* `fabs(rxBw * 1000.0f - point) <= 1000.0f`


jgromes:
Thank you for the contribution! And indeed, this is a mistake. It can be merged but to be honest it will be pretty much a hotfix - we are currently reworking large parts of the library in preparation for the 8.0.0 release, and as part of that we are dropping most floats from the public API (see #1868), including Rx bandwidth configuration via float.

So merged, but unfortunately it won't stay in the library for long.

